PR #16, opened 2022-01-24
Refactor NWBInspector

CodyCBakerPhD:
An overhaul of the existing NWBInspector that introduces a testing workflow as well as a systematic and simplified method of adding check functions with infrastructure friendly to cross-repo plugins (mainly for extensions).

A number of items left on the to-do to avoid losing currently offered features. Various files/functions also ought to be renamed appropriately to replace the original files.

Other To-Do's:

- [x] Consider how class-based approach might save code or make things easier; this is mainly for the main inspector code, for settings, etc.; the decorator approach to the checks themselves seems pretty established from `pycodestyle`
- [x] Go through the original `nwbinspector.py` and make sure we have `check_functions` for everything supported there.
- [x] Command line functionality
- [x] Error code structure

Features for future PRs:
- [ ] flags for ability to run on builders, possible including iterator objects for large data objects
- [ ] Easily loadable extensions (framework in place should be analogous to plugins from `pycodestyle`/`flake8`


CodyCBakerPhD:
@bendichter The only thing about having this PR pointing to the main repo is, without the workflows activated in the main branch here they won't trigger on either CI. I could open a dummy PR back on our fork just for running those, however.

CodyCBakerPhD:
We are considering how an error code structure might be beneficial for classifying types and severity in both a more diverse manner than our `severity=1,2,3`, and less subjective.

Just a first proposition for iteration:

`code = {two-letter Code that distinguishes type}{severity from 0 to 9}`

MM = missing metadata (mostly things DANDI needs for upload)
IM = improperly formatted metadata; things like ISO format
DV = data value based improvements; things like compression, chunking, bool vs int, rate vs timestamp, etc

PI = possibly invalid type or format; things that pass pynwb validation, can't be checked via direct validation, but are likely to be invalid nonetheless. Things like time as longer dimension of a series not being the first axis, or TimeInterval values clearly not being with respect to the session start time, etc.

and the bigger one to discuss with regards to what this repo is intended for (enforcement of non-required best practices)

IV = invalid type or format; naturally more severe than others. Things like timestamps not matching zero-axis length, missing rows in DynamicTable, None values for column in DynamicTable, etc. Things that probably shouldn't actually even pass pynwb validation, and ought to be testable in that validation, so we may want to even just remove this type of thing from the scope of this repo and move them over to pynwb.


Even just those basic codes represent all the checks we have right now, and ought to carry enough freedom for future expansion. An example of how this might go in practice is

```python
@nwbinspector_check(code="DV5", neurodata_type=pynwb.TimeSeries)
def check_regular_timestamps(time_series: pynwb.TimeSeries, time_tol_decimals=9):

...

@nwbinspector_check(code="PI9", neurodata_type=pynwb.TimeSeries)
def check_data_orientation(time_series: pynwb.TimeSeries):
```

For skipping checks in the main inspector functions, one could either specify the exact names of checks not to run, generic code types to skip (e.g., all `MM` types, regardless of severity), or severity threshold similar to what we have now but applied to `0-9`

CodyCBakerPhD:
From discussion, I think the Critical, Best Practice Violations, DANDI requirements, and Best Practice Suggestions are a good four levels of effect 'severity' to have; but I should definitely change all reference of this to strings so that users (and developers) don't have to interface with an arbitrary numerical system.

CodyCBakerPhD:
Changelog on this PR (https://github.com/catalystneuro/nwbinspector/pull/8) shows what a more complex return structure and importance/severity distinction might look like.

PR #19, opened 2022-01-28
Refactor NWBInspector 

CodyCBakerPhD:
An overhaul of the existing NWBInspector that introduces a testing workflow as well as a systematic and simplified method of adding check functions with infrastructure friendly to cross-repo plugins (mainly for extensions).

A number of items left on the to-do to avoid losing currently offered features. Various files/functions also ought to be renamed appropriately to replace the original files.

Final To-Do's:
- [x] Format output of check according to format below.
- [x] Print formatted output to console with nice text wrapping based on window size.
- [x] Write full output to log file for permanent reference.
- [x] NWBFile location references.
- [x] Directory tracking of NWBFiles.
- [x] File output; currently the default, and current overwrite default is `True` to ensure the log output always matches the NWBFile. (If the person ran the logs on a previous version of the file, then fixed the issues and re-created the file, then re-ran the logs). Saves to the same location as the NWBFile to reduce chance of misplacing.
- [x] Console display of file output.
- [x] Complete final review of anything missing.

Features for future PRs:
- [ ] Uncomment each old code block one at a time to update to new standard.
- [ ] Flags for ability to run on builders, possibly including compatibility with iterator objects for large datasets.
- [ ] Easily loadable custom checks to attach with NWB extensions (framework in place should be analogous to plugins from `pycodestyle`/`flake8`.
- [ ] Swap to using `click` instead of `argparse` for command line usage; a bit more robust.
- [ ] More specific usage of `skip` argument to include exclude errors on individual results specific to an NWBFile; basically, just the `1.2.3` number from a previous inspector report (will require all checks on a given file to be numbered even their absence; i.e., the counter still increments but the check does not run or return in the report).
- [ ] Super-nice text-wrapping updates to changing console size. `textwrap` is definitely a component of this, but it still does not resolve a changing terminal size on it's own. Packages like `curses`, `rich`, or any of the other more customizable packages commonly supported. But the default behavior is good enough for now.
- [ ] More robust location parsing, possible form builders or from ExternalResources code?
- [ ] Deal with schema version questions and possibly pynwb version issues (not pinning pynwb at all right now; do really old versions have all data types and behaviors?)
- [ ] Enhanced API for rendering output, possibly a class that allows user specifiable options


CodyCBakerPhD:
Format I'm working to match:
```
1. NWBFile: some_nwbfile_name.nwb
===============================

Critical
--------
 1.1.1. TimeSeries 'MyElectricalSeries' located in 'nwbfile/acquisition'
        check_timestamps_match: The first dimension of `data` does not match the length of `timestamps`. Your data is likely transposed. Time should always be the first dimension.

Best Practice Violation
-----------------------
 1.2.1. SpatialSeries 'SubjectPosition' located in 'nwbfile/processing/behavior/some_task_name/position'
        check_dataset_compression: Your SpatialSeries data (0.5 GB) is not compressed!

 1.2.2. Metadata located in 'nwbfile/general/subject'
        check_subject_species_format: You set your subject_species, but it's not the Latin form.

 1.2.3. TimeSeries 'MyElectricalSeries' located in 'nwbfile/acquisition'
        check_dataset_compression: Your raw data (250 GB) is not compressed!

Best Practice Suggestion
------------------------
 1.3.1. Metadata located in 'nwbfile/general'
        check_institution_metadata: You added some metadata about the experimenter but didn't add their institution...

 1.3.2. TableColumn 'random_1s_and_0s' from DynamicTable 'trials' located in 'nwbfile/intervals/trials/'
        check_boolean_column: Could be boolean instead...

 1.3.3. Metadata located in 'nwbfile/processing/improper_module_name}'
        check_processing_module_modality: The name of 'improper_module_name' processing module is not a recommended modality...
```

CodyCBakerPhD:
@oruebel @bendichter Most of formatting is done, and I was thinking it would be good to keep a 'ground truth' example file in the testing folder so we can have a quick permanent reference to the basic output format: https://github.com/catalystneuro/nwbinspector/blob/write_and_print_logs/tests/true_nwbinspector_log_file_testing.txt

I still need to get the object location reference within the NWBFile working, as well as NWBFile name tracking with directory usage, but I'll probably get those done by EOD.

CodyCBakerPhD:
Basic colors now available in console output
image

bendichter:
@CodyCBakerPhD looks great! Are you handling terminal width?

bendichter:
@CodyCBakerPhD looking good! Can't wait to see the final result!

CodyCBakerPhD:
@bendichter Not in a super nice way just yet; default wrapping produces this kind of result, which I suppose is still better than clipping.
image

bendichter:
minor edit: can you take the underscores out of the titles e.g. "CRITICAL_IMPORTANCE -> "CRITICAL IMPORTANCE"

CodyCBakerPhD:
@bendichter @oruebel I'm focusing on getting the core aspect of 'location' within the NWBFile of the object associated with the violation; since the run logic is based on the `nwbfile.objects` flattening. I do notice that the keys within that have something that might help me determine a human readable form, but I'm having trouble figuring out the best way to do that. Any suggestions?

CodyCBakerPhD:
> minor edit: can you take the underscores out of the titles e.g. "CRITICAL_IMPORTANCE -> "CRITICAL IMPORTANCE"

Ah, yes. Very easy fix

CodyCBakerPhD:
> @bendichter @oruebel I'm focusing on getting the core aspect of 'location' within the NWBFile of the object associated with the violation; since the run logic is based on the nwbfile.objects flattening, I do notice that the keys (example: d4f03889-6bea-44da-b138-c0b46c1ce0d9) within that have something that might help me determine a human readable form, but I'm having trouble figuring out the best way to do that. Any suggestions?

I've experimented with some methods acting on the object, like `this_nwbfile_object.parent.name` from which I can potentially come up with a kind of recursive location inference method (if there aren't any more canonical methods in `pynwb` or `hdmf` dev tools), but for `TimeSeries` objects in `acquisition` I'm still having trouble there inferring that they are in `acquisition` since their parent/ancestor is the `NWBFile` itself, and they are actually within a `field` of that.

bendichter:
> recursive location inference method

what do you mean by this?

For any Group with a Dataset, you can do this by accessing the name of the parent of the h5py.Dataset object like this:

```python
nwb.acquisition["ci_frames"].data.parent.name
```

`'/acquisition/ci_frames'`

CodyCBakerPhD:
> what do you mean by this?

I'm looking for something that provides an explicit mapping of `object_id` for any object within an NWBFile to a path-like description of where that object can be found within the file. A basic example is exactly as you say, for some object  added to `acquisition`, I'd want to take the `object_id` of that and spit out `/acquisition/{name_of_thing_in_acquisition}`.

More explicit and complex example: `SpatialSeries` object  named `my_spatial_series` is in a `Position` container named `example_task` contained in processing module `behavior`. This object was found during iteration of `nwbfile.objects.items()`, for which it has `object_id=d4f03889-6bea-44da-b138-c0b46c1ce0d9`.

I want to report this 'location' as: `/processing/behavior/example_task/my_spatial_series`.

I'm guessing I would just start with the original `SpatialSeries` object and pull the `parent.name`, then follow the chain up until the parent is the outermost NWBFile container (`root`, as reported by this process). Only trouble is sometimes, as with `acquisition`, the `parent` only reports the container that holds the object, not which specific `field` name of the `parent` contains the object, but I think I can figure it out from here.

Just wanted to check if I was missing any default tools that already do this kind of thing.

bendichter:
I think the problem is that the builder layer makes it difficult to map Containers to h5py groups. In that particular case, you could do `spatial_series.data.parent.name`

oruebel:
> I'm looking for something that provides an explicit mapping of `object_id` for any object within an NWBFile to a path-like description

@rly if I remember correctly we had to do similar logic to determine paths for the ExternalResources. Is there logic accessible for this that Cody could use?

oruebel:
> Oh, no worries now - I figured out

Great. For Containers I think the logic of iterating through the names of parents should work since ObjectMappers usually do not modify the names of Containers (although this is theoretically possible, I don't think we do this in practice). For attributes of Containers this logic would not work since attributes are often remapped.  In general, I think one would need to look at the object mapping as part of the build process to get ensure that the path is correct, but I think for now this should work.

rly:
@oruebel we have somewhat similar code, where given a container and the name of an attribute on the container, get the container that is lowest in the type hierarchy on the path to the attribute and get the relative path to the attribute from the HDF5 group/dataset mapped to that container: 
https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/common/resources.py#L359-L389

For example:
- given a `pynwb.TimeSeries` object and the "description" attribute on the object, return the TimeSeries and "description"
- given a `pynwb.TimeSeries` object and the "unit" attribute on the object, return the TimeSeries and "data/unit"
- given a `pynwb.misc.Units` object and the "waveform_rate" attribute on the object, return the VectorData for waveform_mean and "sampling_rate".

rly:
For Containers that are read from disk, I think you can use the BuildManager to translate from Container to path.
```
builder = manager.get_builder(container)
print(builder.path)
```

@CodyCBakerPhD your parse location script works only if there has not been any remapping. For example, there is no "acquisition" container. A TimeSeries in the /acquisition/ group has parent NWBFile, not Acquisition.

That said, it's not clear to me which path is better to show: a) the steps to get to the object in PyNWB or b) the HDF5 path to the object. PyNWB users will mostly not know how to translate one to the other. But non-PyNWB users don't care about the PyNWB steps.

oruebel:
> That said, it's not clear to me which path is better to show ...

For a error-reporting tool like this one, I think it may be useful to simply provide both``path=/acquistion/raw , pynwb=nwbfile.acquisition``. Otherwise, one could also make it a configuration option for the inspector to set which path's to use in error reporting. If this is just for reporting, then it may also be useful to leave this decision up the renderer for the report.

CodyCBakerPhD:
> That said, it's not clear to me which path is better to show: a) the steps to get to the object in PyNWB or b) the HDF5 path to the object. PyNWB users will mostly not know how to translate one to the other. But non-PyNWB users don't care about the PyNWB steps.

Trying to go for the structure that mimics other user-facing access patterns, such as HDFView and NWBWidgets (which yes, is technically python based, but the notebook scan be opened with a bare minimum Python knowledge, and on open platforms like DANDIHub without even needing the user to install Python/conda). Those tend to mimic the HDF5 path, which as Ben points out is best retrieved whenever the object itself is or contains a HDF5 Dataset.

As such I've altered the logic to attempt to pull the information in one-shot from any situation a `h5py.Dataset` is recognized for the object the check runs on, with a the back-up recursion logic (without type map renaming from builders) as the final option before returning default of 'unknown'.

The other options are getting a touch complicated, but should eventually be investigated and incorporated if they turn out to be more reliable. Easiest way to debug these will eventually be to try to make a check that runs on that particular object type, and systematically debug the parsing one we have an explicit failing case; I'll leave this as a long-term to do for now, since this isn't a huge deal either way - I just wanted to point the user in the right direction in case they had a lot of different modules.

CodyCBakerPhD:
@bendichter Directory functionality, which was the last big thing to polish, has now been added. Example file now appears as: https://github.com/catalystneuro/nwbinspector/blob/write_and_print_logs/tests/true_nwbinspector_log_file.txt

When running it on actual files, the actual file path is of course passed (the ones in the example file above are renamed for both my personal anonymity and the testing infrastructure which runs on temporary directories), see console output here: 
image


I found it very handy to enforce a platform-independent way of ordering the NWBFiles delivered in the report so that two people running the `nwbinspector` on the same directory don't end up with different check numbers by chance. `natsort` is by far the best way to do this, and also makes it look more well-organized in my opinion. But that does mean I had to add it to the requirements. It's a very light dependency so hopefully that's not an issue for anyone.

Tomorrow I'll finish off the remaining unit testing that really ought to be included in this first refactor and raise any final questions I might have about specific details.

But otherwise it's ready for final review prior to merge. The original comment to this PR has been updated with the long-term to-do list for future PR's, which I'll probably transfer to specific issues and deal with one at a time in subsequent modular PR's.

bendichter:
@CodyCBakerPhD looks great!

nwbinspector/register_checks.py
```diff
@@ -0,0 +1,97 @@
...
+global available_checks
+CRITICAL_IMPORTANCE = 2
+BEST_PRACTICE_VIOLATION = 1
+BEST_PRACTICE_SUGGESTION = 0
+
+importance_levels = OrderedDict(
+    CRITICAL_IMPORTANCE=CRITICAL_IMPORTANCE,
+    BEST_PRACTICE_VIOLATION=BEST_PRACTICE_VIOLATION,
+    BEST_PRACTICE_SUGGESTION=BEST_PRACTICE_SUGGESTION,
+)
```
oruebel, line 16: I think it would be simpler and cleaner to use a single Enum class to define the importance levels. In this way the levels are logically grouped together in class, rather than each level being a separate global variable. 

```
from enum import Enum

class Importance(Enum):
   """Importance levels for inspector messages"""
   BEST_PRACTICE_SUGGESTION = 0
   BEST_PRACTICE_VIOLATION = 1
   CRITICAL_IMPORTANCE = 2
   
class Severity(Enum):
  """Severity levels for inspector messages for internal use"""
  LOW_SEVERITY = 4
  HIGH_SEVERITY = 5
```

nwbinspector/register_checks.py
```diff
@@ -0,0 +1,97 @@
...
+            auto_parsed_result = check_function(*args, **kwargs)
+            if auto_parsed_result is not None:
+                auto_parsed_result.update(
+                    importance=check_function.importance,
+                    severity=levels_to_severity[auto_parsed_result.get("severity", severity_levels[None])],
+                    check_function_name=check_function.__name__,
+                    object_type=type(obj).__name__,
+                    object_name=obj.name,
+                    location=parse_location(nwbfile_object=obj),
+                )
```
oruebel, line 55: I would suggest to create a data class to represent inspector messages, rather than hand-written dicts. Here an example of how this class could look like. 

```
from dataclasses import dataclass

@dataclass
class InspectorMessage:
   # User-defined parameters
   severity: Severity
   message: str

   # Parameters that are usually set py the register_class decorator
   importance: Importance = Importance.BEST_PRACTICE_SUGGESTION
   check_function_name: str = ""
   object_type: str = ""
   object_name: str = ""
   location: str = None
   
   def update(self,
              importance: Importance,
              check_function_name: str,
              object_type: str, 
              object_name: str,
              location: str):
      """Set parameter values used internally by the NWBInspector"""
      self.importance = importance
      self.check_function_name = check_function_name
      self.object_type = object_type
      self.object_name = object_name
      self.location = location
```

By defining a data class we can define the API more rigorously because we can explicitly define which parameter we expect for the return messages and we can also use type annotations to explictly define what kind of data we expect for each parameter. E.g., by defining the importance levels as an Enum type (see my above suggestion), we can define that the importance parameter should use the values defined in the ``Importance`` enumeration. Another issue this avoids, is that when using hand-written dicts a user may write a new check and accidentally omit or misspell a required key. With hand-written dicts, this will result in an error at runtime somewhere in the internals of the nwbinspector because we are trying to access a key that doesn't exists. In contrast, when we define this as a class, the code will fail already during parsing of the code because Python will raise an error that a parameter is missing or unknown. When using the ``@dataclass`` approach we also get a nice default representation when we want to print ``InspectorMessage`` objects.

Normally I would use ``NamedTuple`` from the ``typing`` module as base class for the ``InsepctorMessage`` class, but  ``NamedTuple`` is immutable and since we here want to be able to update the values of the ``object_name`` etc. in the decorator,  a mutable ``dataclass`` seems simpler and should give us all the functionality we need. 

I added an ``update`` method to the class here, mainly so that the register_check function doesn't need to change and to encapsulate the internal updates in a single function call, but I don't think this is required.

nwbinspector/register_checks.py
CodyCBakerPhD: Great find, that class looks perfect for what we're doing!

nwbinspector/register_checks.py
CodyCBakerPhD: I was a bit hesitant at first since this seemed like it was doing almost the same thing, but see https://github.com/catalystneuro/nwbinspector/pull/10 where I do conclude this can help make this safer in a number of ways.

CodyCBakerPhD:
@oruebel Thanks for the comments, I think both of those details improved the structure quite a bit.

@oruebel @bendichter @rly PR now ready for final review: you can view testing status on https://github.com/catalystneuro/nwbinspector/pull/9

Approximate summary of changelog:
+50 lines of github action workflows and `__init__` for loading the checks
+/- 200 lines of unchanged code for logic that performs check functions in the old way; these are currently commented out but will be modularly updated  (one at a time) to the new API after this PR
+100 for parsing/rendering tools to produce the final log file and console printout; this may eventually become a Rendering class with specifiable options, but let's keep it nice and simple for now. It being the most downstream item, it will be easy to change later.
+124 / -253 improved logic of main inspector code, including command line arguments and new skip/threshold functionality
+139 check registration decorator, Importance/Severity definition, and InspectorMessage data class.
+20 Common tools and utils used across the repo
+20 toml file for black configuration (kind of like the PEP8 config file)
+730 Tests of primary inspector code, registration logic, and unit tests for the few check functions initially included.
+ 80 Placeholder (skipped) tests for the future check functions to be added from the old logic

nwbinspector/checks/time_series.py
```diff
@@ -0,0 +1,85 @@
...
+
+import pynwb
+
+# from ..tools import all_of_type
+from ..register_checks import register_check, Importance, Severity, InspectorMessage
+from ..utils import check_regular_series
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=pynwb.TimeSeries)
+def check_regular_timestamps(time_series: pynwb.TimeSeries, time_tol_decimals=9, gb_severity_threshold=1.0):
```
oruebel, line 12:
```suggestion
def check_regular_timestamps(time_series: pynwb.TimeSeries, time_tol_decimals: int = 9, gb_severity_threshold: float = 1.0):
```
Just adding type hints for the additional parameters. Just a curious question, how are additional parameters on checks handled? Are additional parameters from checks exposed to the user so that they can be set via the command line?

nwbinspector/checks/time_series.py
```diff
@@ -0,0 +1,85 @@
...
+    """If the TimeSeries uses timestamps, check if they are regular (i.e., they have a constant rate)."""
+    if time_series.timestamps is not None and check_regular_series(
+        series=time_series.timestamps, tolerance_decimals=time_tol_decimals
```
oruebel, line 15: Just for consistency, I would suggest renaming ``time_tol_decimals`` to ``time_tolerance_decimals``

nwbinspector/nwbinspector.py
```diff
@@ -1,290 +1,161 @@
+"""Cody Baker, Ben Dichter, and Ryan Ly."""
```
oruebel, line 1: Instead of listing authors in each Python module, I would suggest maybe adding a Contributors file to the repo. You may also want to update the list of authors in setup.py. I think docstrings of modules should ideally just briefly describe what the module does.

nwbinspector/register_checks.py
```diff
@@ -0,0 +1,139 @@
...
+    message : str
+        A message that informs the user of the violation.
+    severity : Severity, optional
+        If a check of non-CRITICAL importance has some basis of comparison, such as magitude of affected data, then
+        the developer of the check may set the severity as Severity.HIGH or Severity.LOW by calling
+        `from nwbinspector.register_checks import Severity`. A good example is comparing if h5py.Dataset compression
+        has been enabled on smaller vs. larger objects (see nwbinspect/checks/nwb_containers.py for details).
+
+        The user will never directly see this severity, but it will prioritize the order in which check results are
+        presented by the NWBInspector.
```
oruebel, line 50: It would be useful to add documentation for the other parameters as well.

nwbinspector/register_checks.py
```diff
@@ -0,0 +1,139 @@
...
+    message: str
+    severity: Severity = None
+    importance: Importance = Importance.BEST_PRACTICE_SUGGESTION
+    check_function_name: str = ""
+    object_type: str = ""
+    object_name: str = ""
+    location: str = ""
+
+
+def register_check(importance, neurodata_type):
```
oruebel, line 62:
```suggestion
def register_check(importance: Importance, neurodata_type):
```

nwbinspector/register_checks.py
```diff
@@ -0,0 +1,139 @@
...
+            return auto_parsed_result
+
+        available_checks[check_function.importance][check_function.neurodata_type].append(auto_parse_some_output)
+
+        return auto_parse_some_output
+
+    return register_check_and_auto_parse
+
+
+def parse_location(nwbfile_object):
```
oruebel, line 107:
```suggestion
def parse_location(nwbfile: : NWBFile):
```

Just for consistency, I would suggest renaming the parameter to ``nwbfile`` and adding the type hint.

setup.py
```diff
@@ -1,26 +1,20 @@
...
-    description='tool to inspect NWB files for best practices compliance',
+    name="nwbinspector",
+    version="0.2.2",
+    description="tool to inspect NWB files for best practices compliance",
     long_description=long_description,
-    long_description_content_type='text/markdown',
-    author='Ryan Ly, Ben Dichter',
-    author_email='rly@lbl.gov, ben.dichter@gmail.com',
+    long_description_content_type="text/markdown",
+    author="Ryan Ly, Ben Dichter",
```
oruebel, line 13:
```suggestion
    author="Ryan Ly, Ben Dichter, Cody Baker",
```

oruebel (approved):
Overall this looks good to me. Thanks for the effort!

I added a couple of nitpicky comments, but nothing critical, mostly minor style edits. I noticed that several files contain lots of functions that have been commented. If that is code intended for future PR's then feel free to leave it in, otherwise, if that code is not needed, I would suggest to remove it.

nwbinspector/checks/time_series.py
CodyCBakerPhD: Sure, I'll go in and triple check all annotations where possible.

As for how those parameters get set by the user, right now they are not nor have we discussed how they should be; I'm treating them purely as internal developer-set values, with `check_function`-specific defaults relative to the severity thresholds for that specific check. So the only situation someone would actually use it in as an optional argument is manually applying one of the checks to a specific NWBFile object (outside command line use).

From my perspective, I am also trying to make such options as general as possible so that the same reserved option name can apply in the same manner across multiple checks, i.e., `gb_severity_threshold` could be a common threshold for any data size issues that effect severity. This means we could theoretically allow it to be specified at command line as a global option to apply to every check that takes that extra argument, but I believe we should think long and hard about that approach as over time it would undoubtedly lead to a large-scale accumulation of additional arguments, most of which a random user would likely not want to take the time to learn about or concern themselves with.

nwbinspector/nwbinspector.py
CodyCBakerPhD: Sure; I usually just do the authors by default to avoid `pydocstyle D100` warnings. Will swap them to brief descriptions, and add a contributors file.

nwbinspector/register_checks.py
CodyCBakerPhD: I didn't do that originally as I did not want the user to think they had to or even should set those themselves at `__init__` of the object.

I could break up the docstring into `Parameters` and `Returns` and add the documentation details for the other fields under `Returns`, thus implying that all those others values are things that should be 'returned' or otherwise filled in automatically by the class after the check completes.

PR #28, opened 2022-02-14
Adding black checker

CodyCBakerPhD:
Adding black checker to workflow to ensure outside contributions meet expected format.


CodyCBakerPhD:
Debugging on Catalyst fork: https://github.com/catalystneuro/nwbinspector/pull/11

.github/workflows/testing.yml
```diff
@@ -1,5 +1,5 @@
 name: Testing
-on: [pull_request, push]
+on: push
```
CodyCBakerPhD: @rly Also, if you prefer activating on push that essentially precedes and accompanies all pull requests; adding the PR option in addition seems to add an artificial display that always shows up as 'skipped'

.github/workflows/testing.yml
```diff
@@ -26,7 +26,8 @@ jobs:
         run: pip install -e .
       - name: Run pytest with coverage
         run: pytest --cov=./ --cov-report xml:./nwbinspector/nwbinspector/coverage.xml
-      - name: Upload coverage to Codecov
+      - if: ${{ matrix.python-version == '3.9' && matrix.os == 'ubuntu-latest' }}
```
CodyCBakerPhD, line 29: Also added this line to constrain unnecessary Codecov uploads, at least until we make Python version-specific or platform-specific checks (not sure we ever will, though...)

CodyCBakerPhD:
@rly @oruebel Also, something we like to use at CatalystNeuro is a codecov bot poster that will display the diff of coverage on a per-push basis. Example display: 
image

This can make it really easy to avoid merging code that would reduce coverage, and is an immediate indicator of if more testing is necessary.

Would you guys have an interest in having this setup for the NWBInspector?

.github/workflows/testing.yml
CodyCBakerPhD: E.g., 
image

.github/workflows/flake8.yml
```diff
@@ -0,0 +1,21 @@
...
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: flake8
+        uses: py-actions/flake8@v2
+        with:
+          max-line-length: 120
+          exclude: "nwbinspector/__init__.py"
```
rly: I think it would be better to set this up in `setup.cfg` and have this Action use `setup.cfg` if it does not by default. Then, devs can run `flake8` locally using the same config.

.github/workflows/testing.yml
rly: I agree the current setup adds an ugly skipped check, so I would like to avoid that if possible. The "pull_request" trigger was to ensure that the job runs in this repo on pushes to a PR from a fork. I *think* with only the "push" trigger, the job will run in the fork repo and not in the PR in this repo. That's ok, but for things like codecov, it is nice to go to https://codecov.io/github/NeurodataWithoutBorders/nwbinspector/ for a report instead of https://codecov.io/github/catalystneuro/nwbinspector/ .

.github/workflows/flake8.yml
CodyCBakerPhD: Agreed, adding that now

.github/workflows/testing.yml
rly: If you think this is not a big deal, feel free to remove the "pull_request" trigger or maybe there is a cleaner solution (see also line 8 for the relevant "if")

CodyCBakerPhD:
@rly @oruebel Actually, Ben and I are wondering now if you'd be open to using `black` instead of `flake8` for this project? It's a stronger enforcement of PEP8, and all the code currently in the repo follows it since my system auto-adjusts to it.

Also the parameters of the checker are already established in the `.toml` for anyone that wants to use it locally, so for the CI here it would just be a simple switch from `flake8` linting to `black` linting.

.github/workflows/testing.yml
CodyCBakerPhD: That's an interesting point on the outside fork, I'll play around with that a bit to determine if it's necessary.

rly:
I support using `black`. Would code be autoformatted on push to GitHub? We should provide instructions for local setup if that is recommended.

CodyCBakerPhD:
> I support using black. Would code be autoformatted on push to GitHub? We should provide instructions for local setup if that is recommended.

Great.

An auto-formatter (either a workflow like we use on NWBCT, or pre-commit hook like DANDI uses) could be something for us to discuss at the Hackathon; there are benefits and drawbacks for each approach. For now it's just linting, identical to what flake8 would do but stronger.

Documentation is definitely a big TODO for this project (and probably shouldn't wait until the Docuthon to get started), but the instructions for getting the black lint to pass are pretty simple (just run `black .`, lol). I'll add something to the PR checklist about that, which is the minimal amount of documentation to begin with there.

setup.cfg
```diff
@@ -1,7 +1,10 @@
...
 exclude =
   .git,
-  .tox,
-  __pycache__
+  __pycache__,
+  build/,
+  dist/,
+  docs
+per-file-ignores =
+  nwbinspector/__init__.py:F401,F403
```
CodyCBakerPhD, line 10: The CI now uses `black`, which is configured via the `pyproject.toml` (more modern and more broadly supported PEP518 config file) file instead of `setup.cfg` (see this delightfully long conversation about why `flake8` won't support it: https://github.com/PyCQA/flake8/issues/234), but I've updated the `setup.cfg` anyway if someone wants to generate a flake8 report separately.

.github/workflows/testing.yml
CodyCBakerPhD: Confirmed; with `push` alone, from an outside fork, the tests are never triggered on `NeurodataWithoutBorders` and hence no codecov upload.

With `pull_request`, from an outside fork, no weird skipping is displayed. When opened within the repo however, it skips one of the two trigger conditions resulting in the weird display, but that's fine given how minor it is versus how important it is for us to support outside contributions.

rly: I agree. That said, if there is a cleaner way to do this, I am all for it. It feels a bit hacky.

rly:
The codecov bot should auto-post when it works (sometimes it just won't run because of issues on their end). If it is not auto-posting, I am all for enabling that.

CodyCBakerPhD:
Working on the pre-commit usage now

CodyCBakerPhD:
Looking good and functional!

Big thanks for the help @satra @yarikoptic @rly!

bendichter (approved):

PR #42, opened 2022-02-15
add checks for ElectricalSeries.data dimension

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/nwbinspector/checks/ecephys.py
+++ b/nwbinspector/checks/ecephys.py
@@ -3,6 +3,9 @@ import numpy as np
 from numbers import Real
 
 from pynwb.misc import Units
+from pynwb.ecephys import ElectricalSeries
+
+from hdmf.utils import get_data_shape
 
 from ..register_checks import register_check, Importance, InspectorMessage
 
@@ -21,4 +24,22 @@ def check_negative_spike_times(units_table: Units):
                 "This Units table contains negative spike times. Time should generally be aligned to the earliest "
                 "time reference in the NWBFile."
             )
+        )
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=ElectricalSeries)
+def check_electrical_series_dims(electrical_series: ElectricalSeries):
+    data = electrical_series.data
+    electrodes = electrical_series.electrodes
+
+    data_shape = get_data_shape(data, strict_no_data_load=True)
+    if data_shape and len(data_shape) == 2 and data_shape[1] != len(electrodes.data):
+        if data_shape[0] == len(electrodes.data):
+            return InspectorMessage(
+                message="The second dimension of data does not match the length of electrodes, "
+                "but instead the first does. Data is oriented incorrectly and should be transposed."
+            )
+        return InspectorMessage(
+            message="The second dimension of data does not match the length of electrodes. Your "
+            "data may be transposed."
         )
--- a/tests/unit_tests/test_ecephys.py
+++ b/tests/unit_tests/test_ecephys.py
@@ -1,6 +1,14 @@
-from pynwb.misc import Units
+from unittest import TestCase
 
-from nwbinspector import check_negative_spike_times
+from uuid import uuid4
+from datetime import datetime
+
+import numpy as np
+from pynwb import NWBFile
+from pynwb.misc import Units
+from pynwb.ecephys import ElectricalSeries
+
+from nwbinspector.checks.ecephys import check_negative_spike_times, check_electrical_series_dims
 from nwbinspector.register_checks import InspectorMessage, Importance, Severity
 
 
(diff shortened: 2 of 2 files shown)


nwbinspector/checks/ecephys.py
```diff
@@ -22,3 +25,21 @@ def check_negative_spike_times(units_table: Units):
...
             )
         )
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=ElectricalSeries)
+def check_electrical_series_dims(electrical_series: ElectricalSeries):
+    data = electrical_series.data
+    electrodes = electrical_series.electrodes
+
+    data_shape = get_data_shape(data, strict_no_data_load=True)
```
CodyCBakerPhD, line 35: What does this helper function do differently than `data.shape`?

tests/unit_tests/test_ecephys.py
```diff
@@ -1,6 +1,14 @@
+from unittest import TestCase
+
+from uuid import uuid4
+from datetime import datetime
+
+import numpy as np
```
CodyCBakerPhD:
```suggestion
import numpy as np
from unittest import TestCase
from uuid import uuid4
from datetime import datetime
```

tests/unit_tests/test_ecephys.py
```diff
@@ -1,6 +1,14 @@
...
+from pynwb import NWBFile
 from pynwb.misc import Units
+from pynwb.ecephys import ElectricalSeries
 
-from nwbinspector import check_negative_spike_times
+from nwbinspector.checks.ecephys import check_negative_spike_times, check_electrical_series_dims
```
CodyCBakerPhD, line 10: All checks can be imported from outermost level; the name of the check ought usually be sufficient for specifying its modality of usage.

tests/unit_tests/test_ecephys.py
```diff
@@ -32,3 +40,77 @@ def test_check_negative_spike_times_some_negative():
         object_name="Units",
         location="/",
     )
+
+
+class TestCheckElectricalSeries(TestCase):
+    def setUp(self):
```
CodyCBakerPhD, line 45:
```suggestion
    @classmethod
    def setUpClass(cls):
```
Suggest using `setUpClass` any time you make multiple usage of the attributes set at setup; individual test function usage is exactly the same but then the setup only needs to be run once in total (makes CI much faster).

tests/unit_tests/test_ecephys.py
```diff
@@ -32,3 +40,77 @@ def test_check_negative_spike_times_some_negative():
...
+                x=3.0,
+                y=3.0,
+                z=3.0,
+                imp=-1.0,
+                location="unknown",
+                filtering="unknown",
+                group=group,
+            )
+
+        self.nwbfile = nwbfile
```
CodyCBakerPhD, line 67:
```suggestion
        cls.nwbfile = nwbfile
```
Only modification needed for `setUpClass` usage.

nwbinspector/checks/ecephys.py
bendichter: `get_data_shape` works on `h5py.Dataset`, `DataChunkIterator`, lists or lists, and numpy arrays. The `.shape` method will break if the inspector is run on an in-memory nwbfile where the user has used a list of lists. It is common, for instance, for `electrodes.data` to be a `list`, not a numpy array.

tests/unit_tests/test_ecephys.py
bendichter: > Imports should be grouped in the following order:
Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.

https://www.python.org/dev/peps/pep-0008/#imports

tests/unit_tests/test_ecephys.py
bendichter: I'd like to have the option to have checks in different modules share the same name or to have modules that have heavy inits that don't get called automatically

tests/unit_tests/test_ecephys.py
bendichter: @CodyCBakerPhD this actually won't work here. I need a fresh NWBFile every time otherwise I will be trying to add data of the same name to a single NWBFile.

nwbinspector/checks/ecephys.py
CodyCBakerPhD: Gotcha

tests/unit_tests/test_ecephys.py
CodyCBakerPhD: Yup, that's what I'm suggesting. All these are standard library imports for Python 3.9, no?

CodyCBakerPhD: Ah, numpy still isn't.

CodyCBakerPhD:
The reason we need to use `nwbfiles` in the tests, instead of just individual `pynwb` objects, is to use the `create_electrode_table_region` function, right?

CodyCBakerPhD (approved):

PR #50, opened 2022-02-16
add checks for time intervals

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -23,18 +23,16 @@ def check_time_interval_time_columns(time_intervals: TimeIntervals):
     if unsorted_cols:
         return InspectorMessage(
             message=f"{unsorted_cols} are time columns but the values are not in ascending order."
-                    ". All times should be in seconds with respect to the session start time."
+            ". All times should be in seconds with respect to the session start time."
         )
 
 
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeIntervals)
 def check_time_intervals_stop_after_start(time_intervals: TimeIntervals):
-    if np.any(
-            np.asarray(time_intervals["stop_time"][:]) - np.asarray(time_intervals["start_time"][:]) < 0
-    ):
+    if np.any(np.asarray(time_intervals["stop_time"][:]) - np.asarray(time_intervals["start_time"][:]) < 0):
         return InspectorMessage(
             message="stop_times should be greater than start_times. Make sure the stop times are with respect to the "
-                    "session start time."
+            "session start time."
         )
 
 
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -1,5 +1,7 @@
 """Check functions that can apply to any descendant of DynamicTable."""
 from hdmf.common import DynamicTable
+import numpy as np
+from pynwb.file import TimeIntervals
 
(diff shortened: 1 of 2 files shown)


nwbinspector/checks/tables.py
```diff
@@ -11,6 +13,29 @@ def check_empty_table(table: DynamicTable):
...
+def check_time_interval_time_columns(time_intervals: TimeIntervals):
+    unsorted_cols = []
+    for column in time_intervals.columns:
+        if column.name[-5:] == "_time":
+            if np.any(np.diff(column[:]) < 0):
+                unsorted_cols.append(column.name)
+    if unsorted_cols:
+        return InspectorMessage(
+            message=f"{unsorted_cols} are time columns but the values are not in ascending order."
+            ". All times should be in seconds with respect to the session start time."
```
CodyCBakerPhD:
```suggestion
            "All times should be in seconds with respect to the session start time."
```

tests/unit_tests/test_tables.py
```diff
@@ -24,6 +25,54 @@ def test_check_empty_table_without_data():
...
 
+def test_check_time_interval_time_columns():
+    time_intervals = TimeIntervals(name="test_table", description="desc")
+    time_intervals.add_row(start_time=2.0, stop_time=3.0)
+    time_intervals.add_row(start_time=1.0, stop_time=2.0)
+
+    assert check_time_interval_time_columns(time_intervals) == InspectorMessage(
+        severity=Severity.NO_SEVERITY,
+        message="['start_time', 'stop_time'] are time columns but the values are not in ascending order."
+        ". All times should be in seconds with respect to the session start time.",
```
CodyCBakerPhD:
```suggestion
        "All times should be in seconds with respect to the session start time.",
```

CodyCBakerPhD:
So rather than iterating over check applications on each column, you'd rather scan the columns within the check function and return a single `InspectorMessage` summarizing which columns fail the check?

Do we want to support iterable check functions at all then?

bendichter:
@CodyCBakerPhD I'm not sure, I guess I could go either way. I wasn't sure how hard it would be to support yeilding and I wanted to move forward with the check. I guess I have a slight preference for yielding. What do you prefer?

CodyCBakerPhD:
> @CodyCBakerPhD I'm not sure, I guess I could go either way. I wasn't sure how hard it would be to support yeilding and I wanted to move forward with the check. I guess I have a slight preference for yielding. What do you prefer?

I don't think it would be hard to support at all, and it does allow for more sophisticated usage in certain cases where that might be important.

There are two points:

a) using the iteration to return a single `InspectorMessage` per column does increase spam in the sense that it might amplify the number of total items in the report by more than other checks would.

b) however, it is possible that a given column that raised a violation is actually correct while other columns on that same table did need to be fixed in accordance with the violation; having separate messages per column more appropriately matches that case

So for accuracy of the reporter (and also thinking of future more sophisticated `ignore` usage that considers individual objects not just general function names) I'd support yielding.

For simplicity of condensed output, then doing the internal check compilation would be better.

CodyCBakerPhD:
So maybe the approach should be to support yielding for more severe violations that ought to be reported a per-column basis, but for the ones you have here single return compilation in the message is fine.

Basically, "it depends on the check" as well as the check developer's preference for such things.

CodyCBakerPhD (approved):

PR #53, opened 2022-02-17
changes to dataset compression check

### LINKED ISSUE #52: problem with compression test
This is showing up for many datasets:

```
BEST PRACTICE VIOLATION
-----------------------
1.1.1   NWBFile 'root' located in '/'
        check_dataset_compression: Consider enabling compression when writing a large dataset.
```
It's not clear what dataset is causing the violation.

bendichter:
fix #52 

* only activate warning when the dataset is > 1GB
* change warning from violation to suggestion
* include field name in inspector message

## Motivation

## How to test the behavior?
```
```

## Checklist


CodyCBakerPhD:
> only activate warning when the dataset is > 1GB

I'll take a second to investigate what is triggering the issue; depending on what it is, 1GB might be a bit larger than necessary to make this a non-issue, but I'm not against setting a hard lower bound on triggering this.

> change warning from violation to suggestion

I'm against this, suggestions are the lowest priority items but it's possible for a dataset that triggered this to be extremely large. Also, if we're already setting a lower bound to yield more accurate (and therefore fewer) trigger scenarios, then the situations where it does flag this would thus be more important.

Also, I get that when we add the DANDI `intention` feature it will escalate something like this to `CRITICAL`, most likely, but I still think as a default practice it should be given higher due consideration than missing metadata.

> include field name in inspector message

Great change.

CodyCBakerPhD:
Some suggestions and tests fixes: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/54

bendichter:
@CodyCBakerPhD can you fix this path issue?

bendichter:
I may have messed things up when I resolved a merge conflict

CodyCBakerPhD:
@bendichter fixed

CodyCBakerPhD (approved):

PR #54, opened 2022-02-17
Fix compression check suggestions

CodyCBakerPhD:
## Motivation

Suggestions for fixing the compression checks; I tried it out on a few of the same DANDISet files you posted to the hub.

Some of these files are also failing `pynwb validation`, so watch for changes later today when I enabled writing of those to the `.txt` report instead of just console.

The things that keep triggering this are `/general/` tab metadata such as `keywords` and the like. A simple 50MB lower bound for activation ought to be sufficient to ignore any metadata fields accidentally triggering this, not to mention very, very small data objects for which compression means nothing. I'd say once you start getting into the 100's of MB though, compression does become beneficial.


nwbinspector/nwbinspector.py
```diff
@@ -111,7 +111,7 @@ def inspect_all(
         print_to_console(log_file_path=log_file_name)
         print(f"{os.linesep*2}Log file saved at {str(log_file_name)}!")
     if num_invalid_files:
-        print(f"{num_exceptions}/{num_nwbfiles} files are invalid.")
+        print(f"{num_invalid_files}/{num_nwbfiles} files are invalid.")
```
CodyCBakerPhD, line 114: Caught this issue earlier

nwbinspector/checks/nwb_containers.py
```diff
@@ -19,11 +19,18 @@ def check_dataset_compression(nwb_container: NWBContainer, gb_severity_threshold
         if (
             isinstance(field, h5py.Dataset)
             and field.compression is None
-            and field.nbytes > (gb_severity_threshold * 1e9)
+            and field.size * field.dtype.itemsize > 50 * 1e6  # 50 MB lower bound
```
CodyCBakerPhD: Note that `field.nbytes` only works with rather recent versions of `h5py`. This package inherits it's `h5py` version support range from `pynwb`, which includes versions that do not support that. But they all support `size` * `itemsize` operation.

CodyCBakerPhD: Specifically, `pynwb` still supports `<v3.0`, feature was added in exactly `v3.0`.

tests/nwbinspector_log_file.txt
```diff
@@ -0,0 +1,26 @@
...
+
+
+
+NWBFile: C:\Users\Raven\AppData\Local\Temp\tmpek5s1w3v\testing1.nwb
+===================================================================
+
+BEST PRACTICE VIOLATION
+-----------------------
+2.1.1   TimeSeries 'test_time_series_2' located in '/acquisition/'
+        check_regular_timestamps: TimeSeries appears to have a constant sampling rate. Consider specifying starting_time=1.2 and rate=2.0 instead of timestamps.
```
bendichter: please remove this

tests/unit_tests/test_containers.py
```diff
@@ -52,16 +60,27 @@ def test_check_dataset_compression_above_default_threshold(self):
     def test_check_dataset_compression_below_manual_threshold(self):
         with h5py.File(name=self.file_path, mode="w") as file:
             nwb_container = self.add_dataset_to_nwb_container(file=file, gb_size=0.1)
-            assert check_dataset_compression(nwb_container=nwb_container, gb_severity_threshold=0.15) is None
+            true_output = InspectorMessage(
+                severity=Severity.LOW,
+                message="test_dataset is not compressed. Consider enabling compression when writing a dataset.",
+                importance=Importance.BEST_PRACTICE_VIOLATION,
```
bendichter: I'd rather make this a suggestion instead of using severity

nwbinspector/checks/nwb_containers.py
bendichter: ok

tests/unit_tests/test_containers.py
CodyCBakerPhD: > change warning from violation to suggestion

I'm against this, suggestions are the lowest priority items but it's possible for a dataset that triggered this to be extremely large. Also, if we're already setting a lower bound to yield more accurate (and therefore fewer) trigger scenarios, then the situations where it does flag this would thus be more important.

CodyCBakerPhD:
To-Do:
- [x] Split into two checks: `check_large_dataset_compression` (VIOLATION, probably like 20GB minimum threshold) and `check_small_dataset_compression` (SUGGESTION, does not apply to anything above 20GB. High severity above 10GB, does not even apply to anything below 50 MB).

nwbinspector/checks/nwb_containers.py
```diff
@@ -5,25 +5,53 @@
...
+    If the data in the Container object is a 'large' h5py.Dataset, check if it has compression enabled.
+
+    Will only return an inspector warning if the size of the h5py.Dataset is larger than 20 GB.
+    """
+    for field in getattr(nwb_container, "fields", dict()).values():
+        if (
+            isinstance(field, h5py.Dataset)
+            and field.compression is None
+            and field.size * field.dtype.itemsize > 20 * 1e9  # 20 GB lower bound
+        ):
```
bendichter:
```suggestion
def check_large_dataset_compression(nwb_container: NWBContainer, threshold_gb: float = 20.0):
    """
    If the data in the Container object is a 'large' h5py.Dataset, check if it has compression enabled.

    Will only return an inspector warning if the size of the h5py.Dataset is larger than (default) 20 GB.
    """
    for field in getattr(nwb_container, "fields", dict()).values():
        if (
            isinstance(field, h5py.Dataset)
            and field.compression is None
            and field.size * field.dtype.itemsize >  threshold_gb * 1e9
        ):
```

nwbinspector/checks/nwb_containers.py
```diff
@@ -5,25 +5,53 @@
...
+            return InspectorMessage(
+                severity=Severity.HIGH,
+                message=f"{os.path.split(field.name)[1]} is a large uncompressed dataset! Please enable compression.",
+            )
 
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBContainer)
-def check_dataset_compression(nwb_container: NWBContainer, gb_severity_threshold: float = 1.0):
+def check_small_dataset_compression(nwb_container: NWBContainer, gb_severity_threshold: float = 10.0):
```
bendichter:
```suggestion
def check_small_dataset_compression(nwb_container: NWBContainer, mb_lower_threshold: float = 50.0, gb_severity_threshold: float = 20.0):
```

nwbinspector/checks/nwb_containers.py
```diff
@@ -5,25 +5,53 @@
...
 
     Will only return an inspector warning if the size of the h5py.Dataset is larger than bytes_threshold.
     """
     for field in getattr(nwb_container, "fields", dict()).values():
+        nbytes = field.size * field.dtype.itemsize
         if (
             isinstance(field, h5py.Dataset)
             and field.compression is None
-            and field.nbytes > (gb_severity_threshold * 1e9)
+            and nbytes > 50 * 1e6  # 50 MB lower bound
```
bendichter:
```suggestion
            and nbytes > mb_lower_threshold * 1e6  # 50 MB lower bound
```

nwbinspector/checks/nwb_containers.py
```diff
@@ -5,25 +5,53 @@
...
+            and nbytes < 20 * 1e9  # 20 GB upper bound to prevent double-raise with check_large_dataset_compression
```
bendichter:
```suggestion
            and nbytes < gb_severity_threshold * 1e9  # upper bound to prevent double-raise with check_large_dataset_compression
```

bendichter (approved):

PR #55, opened 2022-02-17
fix save path for log file

CodyCBakerPhD:
## Motivation

Previously, the `nwbinspector` would save the log file to the location the `nwbinspector` was called via CLI. When the CLI points to an absolute path of an NWBFile in a different directory, the two no longer correspond.


CodyCBakerPhD:
To-Do:
- [x] check if full path can be specified via CLI
- [x] adjust tests call to save path to not save within repo (most annoying thing atm)

bendichter (approved):

PR #56, opened 2022-02-17
added optional disabling of color from CLI

CodyCBakerPhD:
## Motivation

On certain consoles, coloration of output is not supported; when it is not supported it adds a bunch of garbage characters to the left of every line of the output.

We still want to print colors by default, so I've added an optional flag for it to be disabled, as well as a safety fall-back function from Django that attempts (but may not always succeed) in determining a console environments ability to render the color.

## How to test the behavior?
No clue how to test color output in a CI console.


CodyCBakerPhD:
Updates: I tested it out locally on Windows with and without `colorama`, and there are some issues with the display even when it works (some of our colors aren't really supported with it). So I removed support for it here.

Also checked the codecov reports and added `# pragma: no cover` on the things we can't test.

bendichter (approved):

PR #62, opened 2022-02-17
Propogate validiation and errors

### LINKED ISSUE #36: Allow validator error messages to propagate
It only currently tells which files and how many files failed due to validations and other errors. Basically just copy the Exception traceback outside those try/excepts

CodyCBakerPhD:
## Motivation

@bendichter First details for incorporating errors/validation stuff into the reports. I still need to polish it a bit and add tests for things.

Outputs the information, though - for DANDISet 000004 (plus an manual error-inducing file above that):
[nwbinspector_log_file.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8092339/nwbinspector_log_file.txt)


CodyCBakerPhD:
@bendichter Got this working; adds two extra fields to the top of the report (only if they are present - that they were triggered by the inspector), "ERROR" with exact traceback, "PYNWB_VALIDATION" with the pynwb-parsed location of the object along with the validator message.

The biggest change this required to current API relevant to your usage, is the `inspect_new` function now takes `nwbfile_path` as an argument instead of the open `nwbfile` object. This is largely because of the way the `pynwb.validator` works, which only takes in an open `io` so that it can run on builders rather than the NWBFile itself.

PR #71, opened 2022-02-20
Improved templates for issues

CodyCBakerPhD:
## Motivation

GitHub is improving their support for advanced issue templates; these new ones appear and behave more like abbreviated forms, which makes it much easier for users and developers to communicate the minimal amount of information directed to their current needs. It also allows us to include tidbits of extra information (like nicer external links) to direct them in their search for answers.

It's hard to prototype these on the central repo here, so feel free to examine the working prototype on the CatalystNeuro fork here: https://github.com/catalystneuro/nwbinspector/issues/new/choose

with some example outputs here: https://github.com/catalystneuro/nwbinspector/issues

That said, we talked about also doing this for pull requests, but unfortunately that is not quite as well supported yet. So we still need to decide on a good way to provide change-specific checklists for PRs, which will differ based on if the person is simply adding a new check function, simple debug, or new feature/refactor. So I'm keeping that template general for now.


CodyCBakerPhD:
```matlab
a = 1
b = a + 1
```

bendichter (approved):

PR #72, opened 2022-02-21
[Add Check]: table column could be boolean

CodyCBakerPhD:
## Motivation

If a column on a `DynamicTable` has data values that could be simplified to `bool`'s to save on storage space, we warn the user as such. Pretty low importance, but it is best practice.

This also represents the first check function that intentionally iterates (via `yield`) over columns so it can report messages (and stats such as saved # bytes) for each individual column separately. This did require one little change to the decorator to properly return `None` when no such column yielded a violation.

Also, intentionally submitted this via an outside fork to test our CI capabilities in that respect.

## How to test the behavior?
Added tests for the skipping behavior, which bypasses several cases such as arrays (content not inspected), jagged arrays, or cases where the storage data type is not more expensive than bools.

## Checklist


CodyCBakerPhD:
Note to merger (including self): accidentally inherited the template forms present on the CatalystNeuro fork, inflating the commit history. May want to squash this one.

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,39 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+            column_data = column.data if hasattr(column.data, "dtype") else np.array(column.data)
+            unique_values = np.unique(column_data[:nelems])
+            column_dtype = column_data.dtype
+            saved_bytes = (column_dtype.itemsize - 1) * column_data.size
+            if (
+                column_data.dtype != np.dtype("object")  # jagged shaped lists
+                and len(column_data.shape) == 1
+                and saved_bytes != 0
+                and len(unique_values) == 2
+            ):
```
bendichter, line 102: a better way to handle jagged arrays would be to access the unindexed data directly. Something like 

```python
if isinstance(column, VectorIndex):
    data = column.target.data
```

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,39 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+        Only check the first {nelems} elements. This is useful in case there columns are
+        very long so you don't need to load the entire array into memory. Use None to
+        load the entire arrays.
+    """
+    for column in table.columns:
+        if hasattr(column, "data"):
+            column_data = column.data if hasattr(column.data, "dtype") else np.array(column.data)
+            unique_values = np.unique(column_data[:nelems])
+            column_dtype = column_data.dtype
+            saved_bytes = (column_dtype.itemsize - 1) * column_data.size
```
bendichter: what is this doing? Are you checking if the data type is boolean? If so, why not just do that directly?

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,39 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+            if (
+                column_data.dtype != np.dtype("object")  # jagged shaped lists
+                and len(column_data.shape) == 1
```
bendichter: Why does it matter if the data is a 1D vector? This test seems applicable to columns of any shape.

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,39 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+        load the entire arrays.
+    """
+    for column in table.columns:
+        if hasattr(column, "data"):
+            column_data = column.data if hasattr(column.data, "dtype") else np.array(column.data)
+            unique_values = np.unique(column_data[:nelems])
+            column_dtype = column_data.dtype
+            saved_bytes = (column_dtype.itemsize - 1) * column_data.size
+            if (
+                column_data.dtype != np.dtype("object")  # jagged shaped lists
```
bendichter: a better way to handle jagged arrays would be to access the unindexed data directly. Something like 

```python
if isinstance(column, VectorIndex):
    data = column.target.data
```

nwbinspector/checks/tables.py
CodyCBakerPhD: The question is, should it be the responsibility of this check to examine the set of values from all nested array contents?

I was operating under the assumption that it would only operate on simple 1D vectors.

nwbinspector/checks/tables.py
CodyCBakerPhD: Reporting the # of bytes that could be saved if the user fixes this check

nwbinspector/checks/tables.py
bendichter: > The question is, should it be the responsibility of this check to examine the set of values from all nested array contents?

what do you mean?

CodyCBakerPhD: Right now it only inspects the contents of columns that contain a single value, such as https://github.com/NeurodataWithoutBorders/nwbinspector/pull/72/files#diff-c5a67d4b9c1f4fdf1c7808b5f06bdea82886bd08a3700d3073a79008253b86b2R164-R182

It then bypasses the more complicated settings where the value at each row is itself a vector: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/72/files#diff-c5a67d4b9c1f4fdf1c7808b5f06bdea82886bd08a3700d3073a79008253b86b2R132-R153

nwbinspector/checks/tables.py
bendichter: Could we add a bytes formatter that naturally converts to KB, MB, GB, etc?

nwbinspector/checks/tables.py
bendichter: I think it would be better to check the target data values in the case of a VectorIndex like I am suggesting above

nwbinspector/checks/tables.py
CodyCBakerPhD: Sure

bendichter: https://stackoverflow.com/a/1094933/2559070

CodyCBakerPhD:
@bendichter I refined the logic a bit and extended full support for array contents (including jagged). Also added the byte size formatter.

There are some things to be aware of as we iterate over columns of a table:

(1a) If a `VectorData` column as set `index=True`, it seems that calling `table[column.name].data` returns the data of the corresponding `VectorIndex`, indicating that in such a situation calling `table[column.name].target.data` is the appropriate response.

(1b) However, if we iterate over `table.columns`, we do actually eventually hit the `VectorData` column itself that is the exact output of  `table[column.name].target.data` of the corresponding `VectorIndex`.

(1c) During iteration of `table.columns` in this manner, we will eventually hit both the `VectorData` and its corresponding `VectorIndex`, leading to a double-raise of this violation if I were to implement something of the form

```python
if isinstance(column, VectorIndex):
    data_to_dest = column.target.data
```

As such, what I've opted for instead is to simply skip `VectorIndex` types since their associated data targets will be iterated over eventually.

CodyCBakerPhD:
Other things to be aware of (you probably already know these, though):

(2)

Before an `io.write` to an NWBFile, you can set the column data of a `DynamicTable` to any kind of nested list without setting `index=True`. If the list forms a jagged array, you do indeed get an error on `io.write`. If it is regular, though, the data still gets saved to the NWBFile as a `h5py.Dataset`, and no associated `VectorIndex` is included.

This is still handled by the newly refined logic (`np.unique()` will act on the shaped data but seemingly only applies to the interior contents). The only minor difference in this case is that `column.data[:nelems]` will extract the arrays for the first `nelem` columns, so if those arrays are themselves quite long, it could lead to more data being loaded than one might have expected.

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,34 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+        very long so you don't need to load the entire array into memory. Use None to
+        load the entire arrays.
+    """
+    for column in table.columns:
+        if hasattr(column, "data") and not isinstance(column, VectorIndex):
+            unique_values = np.unique(column.data[:nelems])
+            saved_bytes = (unique_values.dtype.itemsize - 1) * np.product(
+                get_data_shape(data=column.data, strict_no_data_load=True)
+            )
+            if saved_bytes != 0 and len(unique_values) == 2:
```
bendichter: * How would you have 0 saved bytes with 2 unique values? Don't you then need at least 2 values?
* What about data this is of type bool?

tests/unit_tests/test_tables.py
```diff
@@ -116,6 +121,59 @@ def test_pass_check_time_intervals_stop_after_start():
...
+        table.add_row(test_col=x)
+    assert check_column_binary_capability(table=table) is None
+
+
+def test_pass_array_check_column_binary_capability():
+    table = DynamicTable(name="test_table", description="")
+    table.add_column(name="test_col", description="")
+    for x in [[1.0, 2.0], [2.0, 3.0], [1.0, 2.0]]:
+        table.add_row(test_col=x)
+    assert check_column_binary_capability(table=table) is None
```
bendichter: I think it would be better to test if all of the individual elements of the matrix are 2 values, e.g. all 0s and 1s. This comes up with image masks in `PlaneSegmentation` objects, which are often binary ndarrays and sometimes they mistakenly use floats.

nwbinspector/checks/tables.py
CodyCBakerPhD: bools have an itemsize of 1, same as a `uint8`/`int8`. Kind of an edge case there. Could still be best practice to reduce the `uint8` values to a `bool`, just because it makes it more obvious that there are only two possible values there.

tests/unit_tests/test_tables.py
CodyCBakerPhD: You mean add a failing check (where the data is indeed boolean but stored as float, thus returning an `InspectorMessage`), but in array? This particular test is passing because there are 3 unique values among all the elements of the 3x2 array.

bendichter:
So there's a complication to be aware of here. Technically, HDF5 does not have a binary type. h5py defines its own enumerated type which under-the-hood is actually using uint8 mapped to TRUE and FALSE [source](https://docs.h5py.org/en/stable/faq.html#what-datatypes-are-supported). This is tricky for MatNWB users to write, since it is an h5py-specific hack, and using straight uint8 with no enum layer is just as memory-efficient. so I think we should allow uint8 in addition to logical type.

tests/unit_tests/test_tables.py
bendichter: oh, ok, I think I misunderstood and that you are doing this right. I didn't realize this was passing the check

CodyCBakerPhD:
> So there's a complication to be aware of here. Technically, HDF5 does not have a binary type. h5py defines its own enumerated type which under-the-hood is actually using uint8 mapped to TRUE and FALSE [source](https://docs.h5py.org/en/stable/faq.html#what-datatypes-are-supported). This is tricky for MatNWB users to write, since it is an h5py-specific hack, and using straight uint8 with no enum layer is just as memory-efficient. so I think we should allow uint8 in addition to logical type.

Ah, that would explain it. I knew otherwise that in Python, all objects have to be addressable down only to the level of a byte (hence why a bool, which ought to be representable as a single bit, takes 8 bits per item), but I didn't know the mapping onto `uint8` was that strong.

The good news is this check currently won't trigger on `uint8` data since it doesn't save any size - this is tested explicitly here: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/72/files#diff-c5a67d4b9c1f4fdf1c7808b5f06bdea82886bd08a3700d3073a79008253b86b2R148-R153

nwbinspector/checks/tables.py
bendichter: ok I see that this technically works correctly but for the sake of my sanity can you please change it to something like 

```python
if data.dtype.itemsize == 1:
    continue  # is already boolean, int8, or uint8
```

nwbinspector/checks/tables.py
```diff
@@ -74,6 +75,36 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
...
+    time_intervals: DynamicTable
+    nelems: int
+        Only check the first {nelems} elements. This is useful in case there columns are
+        very long so you don't need to load the entire array into memory. Use None to
+        load the entire arrays.
+    """
+    for column in table.columns:
+        if hasattr(column, "data") and not isinstance(column, VectorIndex):
+            if np.asarray(column.data[0]).itemsize == 1:
+                return  # already boolean, int8, or uint8
```
CodyCBakerPhD:
```suggestion
                continue # already boolean, int8, or uint8
```

I think if this is `return` it won't proceed to examine any other columns on that table.

bendichter: ah

bendichter:
So I am thinking maybe we want to only raise this if the values are 0.0 and 1.0. The reason is that if you have values like 'left', and 'right', maybe using compression will be better because it will probably end up as about the same size and will be much easier for a re-user of the data to interpret.

CodyCBakerPhD:
> So I am thinking maybe we want to only raise this if the values are 0.0 and 1.0. The reason is that if you have values like 'left', and 'right', maybe using compression will be better because it will probably end up as about the same size and will be much easier for a re-user of the data to interpret.

Yeah that is definitely a design consideration.

A counter-example though would be a column of string values "yes" and "no" which has the same informative effect as `True` and `False`.

The "left" vs. "right" case is definitely more informative, but that's why this is only a `BEST_PRACTICE_SUGGESTION`. The user should ignore this check for the column in this case.

Also, I don't think columns are compressed by default, and I don't think I've ever invoked that myself. Should we start recommending that?

Side-topic: It's also kind of been unofficial best practice to name true boolean columns `is_{condition}` to additionally indicate the nature of the results. Should we include that in the official Best Practices and mention it in the message here?

bendichter:
The thing is in order to use compression you can't build the tables row-by-row, you have to build them column-by-column. Other than that I think yes it is probably a good idea to compress them. That will give you the added bonus of being able to append to tables by row post-write.

We could try to come up with a list of values: 
* 1.0, 0.0
* .lower() == 'Y', 'N'
* .lower() == 'yes', 'no'
* 'lower() == 'T', 'F'
* .lower() == 'true', 'false'

I like the `is_` convention. I use it for this and I also use it for functions that output T/F. I would support adding it to best practices

CodyCBakerPhD:
Just remembered a case from one of our past projects: "hit" and "miss". We ended up changing the column name to "is_hit" and mapped "hit" -> True, "miss" -> False.

CodyCBakerPhD:
> 1.0, 0.0

It's a good point that if (i) a non-zero float value is present, the value itself could be informative of something else, and (ii) if the two unique float values are both non-zero, could indicate even more secondary information.

CodyCBakerPhD:
@bendichter Ready for review now

bendichter (approved):

PR #86, opened 2022-03-02
add config file option

### LINKED ISSUE #37: Add "config file" support
not yet sure about all "features" but such config file should support but it must allow to specify 
- which checks for which paths to ignore

This is largely inspired by config file in bids-validator, see https://github.com/bids-standard/bids-validator#configuration 

It is needed since some of such ignores are "per dataset" and thus it should be possible to specify what to ignore in a specific dataset (or dandiset) in DANDI land, instead of relying on user to know what command line options to provide.

bendichter:
fix #37 

restructure available_checks
add config json schema
add example config file
add configure_checks function and corresponding test

TODO: expose to command line


bendichter:
@CodyCBakerPhD I started to build out the feature for custom configuration file that changes the severity of checks. In doing so, I simplified the available_checks dictionary from:

{Importance: {nd_type: [checks]}}

to

{Importance: [checks]}

the nd_type is pulled from the attribute of the check. This is a simplification without loss of function for the library, and it makes the config feature easier to implement.

CodyCBakerPhD:
@bendichter  👍 Yup, looking over it now.

The main reason for pre-sorting by neurodata type was to save time complexity from having to manually inspect it during each step of the outer iteration, but I can re-analyze it again and see if there's an even better way to structure that.

nwbinspector/config.schema.json
```diff
@@ -0,0 +1,23 @@
...
+        "type": "string"
+      }
+    },
+    "BEST_PRACTICE_VIOLATIONS": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "BEST_PRACTICE_SUGGESTIONS": {
```
CodyCBakerPhD:
```suggestion
    "BEST_PRACTICE_VIOLATION": {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    "BEST_PRACTICE_SUGGESTION": {
```

These should exactly match the name of the corresponding `Importance` levels, right?

nwbinspector/nwbinspector.py
```diff
@@ -117,6 +124,24 @@ def inspect_all(
...
+        for func_name in func_names:
+            for importance, functions in output_checks.items():
+                if importance.name == importance_name:
+                    continue
+                i = 0
+                while i < len(functions):
+                    if functions[i].__name__ == func_name:
+
+                        output_checks[Importance._member_map_[importance_name]].append(functions.pop(i))
+                    else:
```
CodyCBakerPhD: While I think more about the main iteration restructuring, a couple implementation questions: 

a) note that this does not alter the internal `importance` attribute set within each check function; i.e., `functions[i].importance` - is this intended to remain at its default value both in the new copy and in the global available_checks?
b) For `Importance._member_map_[importance_name]` I'm pretty sure you can just do `Importance[importance_name]`, right?

nwbinspector/nwbinspector.py
```diff
@@ -172,22 +197,21 @@ def inspect_nwb(
                     unorganized_results["PYNWB_VALIDATION"].append(message)
             nwbfile = io.read()
             check_results = list()
-            for importance, checks_per_object_type in checks.items():
+            for importance, check_functions in checks.items():
                 if importance.value >= importance_threshold.value:
-                    for check_object_type, check_functions in checks_per_object_type.items():
-                        for nwbfile_object in nwbfile.objects.values():
-                            if issubclass(type(nwbfile_object), check_object_type):
-                                for check_function in check_functions:
```
CodyCBakerPhD, line 180: I remember now; the original hope was that we would be able to do away with a level of iteration by directly subsetting the registry based on the NWBFile object's type; that is, `for check_function in check_functions[type(nwbfile_object)]`. That would have saved a certain amount of time from unnecessary `if` checks, however, when we started defining checks over general object types (`NWBContainer`, `TimeSeries`, etc.) this became impossible which is why we test `if issubclass`.

So yeah, I suppose we don't actually gain anything anymore by sub-indexing `Importance` per `neurodata_type`, so flattening the registry should be fine then.

nwbinspector/nwbinspector.py
```diff
@@ -172,22 +197,21 @@ def inspect_nwb(
...
-                                            check_results.extend(output)
-                                        else:
-                                            check_results.append(output)
+                    for nwbfile_object in nwbfile.objects.values():
+                        for check_function in check_functions:
+                            if issubclass(type(nwbfile_object), check_function.neurodata_type):
+                                if ignore is not None and check_function.__name__ in ignore:
+                                    continue
+                                if select is not None and check_function.__name__ not in select:
+                                    continue
```
CodyCBakerPhD, line 208:
```suggestion
                    for nwbfile_object in nwbfile.objects.values():
                        for check_function in check_functions:
                            if ignore is not None and check_function.__name__ in ignore:
                                 continue
                            if select is not None and check_function.__name__ not in select:
                                 continue
                            if issubclass(type(nwbfile_object), check_function.neurodata_type):
```

Minor suggestion for more speed; can ignore `if issubclass` call entirely if this check name isn't selected or is ignored.

CodyCBakerPhD:
As I look at the primary iteration usage of `ignore` and `select`, would these make sense to include somehow as config options used to shorten the full `available_checks`? Or even possibly `importance_threshold`?

That way, we could move those nested `if` checks outside of the iteration and instead pass a shortened collection of `checks` into `inspect_nwb`, reducing the number of operations that get multiplied. And also making the input arguments to `inspect_nwb` much simpler by only expecting some pre-configured specification of `checks` to run on the NWBFile.

Extreme example: we only want to `select` a couple checks. Currently the code still iterates over the entire registry of check functions **for each object** in the NWBFile.

```python
                    for nwbfile_object in nwbfile.objects.values():    <---- could have a large number of objects
                        for check_function in check_functions:    <---- could have a large number of functions
                            if issubclass(type(nwbfile_object), check_function.neurodata_type):
                                if ignore is not None and check_function.__name__ in ignore: <-- line gets hit every time
                                    continue
                                if select is not None and check_function.__name__ not in select:
                                    continue
                                output = check_function(nwbfile_object)    <---- only reaches here a couple of times
```

CodyCBakerPhD:
Oh, I also think it would help the design of the configuration approach (and probably other things too...) if the initial registry `availabe_checks` was a dictionary whose keys are the check function names and whose values are the check functions themselves. An advantage of this would be the ability to take those string names from the JSON and directly index like `available_checks[check_function_name]`.

With this approach, the default configuration would simply be the `available_checks` structured into what they are now (an `OrderedDict` of `Importance` with lists of check functions as values at each level), but any other config would follow the same structure +/- manipulation of importance levels, inclusion or exclusion of particular check names, etc.

What do you think?

nwbinspector/nwbinspector.py
```diff
@@ -117,6 +124,24 @@ def inspect_all(
...
+    for importance_name, func_names in config.items():
+        for func_name in func_names:
+            for importance, functions in output_checks.items():
+                if importance.name == importance_name:
+                    continue
+                i = 0
+                while i < len(functions):
+                    if functions[i].__name__ == func_name:
+
+                        output_checks[Importance._member_map_[importance_name]].append(functions.pop(i))
```
```suggestion
                        output_checks[importance_name].append(functions.pop(i))
```

nwbinspector/nwbinspector.py
bendichter: a) good point
b) yes, also good point. That's much better syntax.

nwbinspector/nwbinspector.py
```diff
@@ -117,6 +124,24 @@ def inspect_all(
...
+    for importance_name, func_names in config.items():
+        for func_name in func_names:
+            for importance, functions in output_checks.items():
+                if importance.name == importance_name:
+                    continue
+                i = 0
+                while i < len(functions):
+                    if functions[i].__name__ == func_name:
+
+                        output_checks[importance_name].append(functions.pop(i))
```
```suggestion
                        output_checks[Importance[importance_name]].append(functions.pop(i))
```

bendichter:
@CodyCBakerPhD yes, I think that could make things simpler

bendichter:
So the config logic would simply be something like:

```python
checks = copy.copy(available_checks)
for importance, func_names in config.items():
    func = checks[func]
    func.importance = importance
    check[func] = func
````

**Comment by CodyCBakerPhD** [comment]
> So the config logic would simply be something like:

Yeah, pretty much. Looks much cleaner.

**Comment by bendichter** [comment]
@CodyCBakerPhD I went even further and just made `available_checks` a list of checks

File: nwbinspector/dandi.inspector_config.yaml [hunk 7]
```diff
@@ -0,0 +1,7 @@
+CRITICAL:
+  - check_subject_exists
+  - check_subject_id_exists
+BEST_PRACTICE_VIOLATIONS:
```
CodyCBakerPhD:
```suggestion
BEST_PRACTICE_VIOLATION:
```

Same thing from before, I assume this has to match exactly.

Using the JSON validation against the schema should catch misspellings in instances of configurations like this, right?

requirements.txt
```diff
@@ -1 +1,2 @@
 pynwb
+PyYAML
```
CodyCBakerPhD, line 2: Oh, didn't even see that we had a requirements file but weren't using it. Should we even keep this file, then? Or should we just update the setup to read from it?

CodyCBakerPhD:
> I went even further and just made available_checks a list of checks

@bendichter Yup, this all looks really good now.

CodyCBakerPhD:
Also, I just remembered Dorota pointed out that it may be for the best to only allow elevation of Importance levels via the config to prevent people from being able to forcibly downgrade important things; do you agree with that?

Also also, Yarik mentioned having the config be able to specify SKIP type behavior on a per-NWBFile level when this config is meant to be run on a directory. Any ideas for how that could go into this schema?

bendichter:
> Also also, Yarik mentioned having the config be able to specify SKIP type behavior on a per-NWBFile level when this config is meant to be run on a directory. Any ideas for how that could go into this schema?

I don't know what you mean

CodyCBakerPhD:
From #37 

> It is needed since some of such ignores are "per dataset" and thus it should be possible to specify what to ignore in a specific dataset (or dandiset) in DANDI land, instead of relying on user to know what command line options to provide.

Unless he means to apply the SKIP rule over all files in that directory and not NWBFile-specific.

Example: `/000003/file1.nwb` has a `check_data_orientation` violation that is later confirmed to be OK so `check_data_orientation` is added to the SKIP field of the config used when running on the entire dataset `000003`. But `/000003/file2.nwb` also had a `check_data_orientation` violation that was confirmed to be a user mistake.

bendichter:
@CodyCBakerPhD OK, I understand. Skipping checks for specific files. I'd say if you want to do that you can run your own script and do something like 

```python
for file in glob.glob(os.path.join(path, '*.nwb'):
    if file == 'bad_file.nwb':
        ignore = ["check_data_orientation"]
    else:
        ignore = None
    inspect_nwb(file, ignore=ignore)

```

requirements.txt
bendichter: it's not a problem for now. I personally like having both so we can hard set this one to specific versions and keep the `install_requires` arg more flexible.

CodyCBakerPhD: 👍

nwbinspector/config.schema.json
```diff
@@ -0,0 +1,29 @@
+{
+  "type": "object",
```
CodyCBakerPhD: @bendichter Should we version this schema?

bendichter: good call!

nwbinspector/nwbinspector.py
```diff
@@ -63,12 +65,20 @@ def inspect_all_cli(
     ignore: Optional[str] = None,
     select: Optional[str] = None,
     threshold: str = "BEST_PRACTICE_SUGGESTION",
+    config_path: Optional[str] = None,
     json_file_path: str = None,
 ):
     """Primary CLI usage."""
+    if config_path is not None:
+        with open(file=config_path, mode="r") as stream:
+            config = yaml.load(stream, yaml.Loader)
```
CodyCBakerPhD, line 74: Should this perform

```python
with open(file=Path(__file__).parent / "config.schema.json", mode="r") as fp:
    schema = json.load(fp=fp)
jsonschema.validate(config, schema)
```

do you think?

CodyCBakerPhD:
@bendichter Two minor suggestions for schema-pertaining items.

Otherwise LGTM

CodyCBakerPhD (approved):

nwbinspector/nwbinspector.py
bendichter:
```python
file=Path(__file__).parent / "config.schema.json"
```

I don't like hard-coding this. We should allow the user to provide their own config file from anywhere they want to store it.

Validating seems like a good idea though.

bendichter: I see. Sure.

PR #89, opened 2022-03-07
Flatten output

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -74,8 +74,8 @@ def display_messages_by_importance(messages, indent_sz=2):
         disp.append("-" * (len(importance.name) + 4))
         for ii, (check_name, check_data) in enumerate(imp_data.items()):
             disp.append(f"{i}.{ii}.  {check_name}")
+            counter = 0
             for file, file_messages in check_data.items():
-                counter = 0
                 for message in file_messages:
                     disp.append(
                         f"{indent}{i}.{ii}.{counter}.  {file}:{message.location}{message.object_name} -"
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -41,10 +41,7 @@ def organize_messages_by_file(messages):
     for file in files:
         out[file] = {}
         file_messages = list(filter(lambda x: x.file == file, messages))
-        for importance in sorted(
-                set(message.importance for message in file_messages),
-                key=lambda x: -x.value
-        ):
+        for importance in sorted(set(message.importance for message in file_messages), key=lambda x: -x.value):
             out[file][importance] = sorted(
                 filter(lambda x: x.importance == importance, file_messages),
                 key=lambda x: -x.severity.value,
@@ -91,7 +88,10 @@ def display_messages_by_importance(messages, indent_sz=2):
 @click.option("-m", "--modules", help="Modules to import prior to reading the file(s).")
 @click.option("--no-color", help="Disable coloration for console display of output.", is_flag=True)
 @click.option(
-    "--report-file-path", default=None, help="Save path for the report file.", type=click.Path(writable=True),
+    "--report-file-path",
+    default=None,
+    help="Save path for the report file.",
+    type=click.Path(writable=True),
 )
 @click.option("-o", "--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("-i", "--ignore", help="Comma-separated names of checks to skip.")
@@ -274,11 +274,7 @@ def inspect_nwb(
         try:
             nwbfile = io.read()
         except Exception as ex:
-            yield InspectorMessage(
-                message=traceback.format_exc(),
-                importance=Importance.ERROR,
-                check_function_name=ex
-            )
+            yield InspectorMessage(message=traceback.format_exc(), importance=Importance.ERROR, check_function_name=ex)
         if select:
             checks = [x for x in checks if x.__name__ in select]
         elif ignore:
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -52,15 +52,46 @@ def organize_messages_by_file(messages):
     return out
 
 
+def organize_messages_by_importance(messages):
+    out = {}
+    for message in sorted(messages, key=lambda x: (-x.importance.value, -x.severity.value, x.file)):
+        if message.importance not in out:
+            out[message.importance] = dict()
+        if message.check_function_name not in out[message.importance]:
+            out[message.importance][message.check_function_name] = dict()
+        if message.file not in out[message.importance][message.check_function_name]:
+            out[message.importance][message.check_function_name][message.file] = []
+        out[message.importance][message.check_function_name][message.file].append(message)
+    return out
+
+
+def display_messages_by_importance(messages, indent_sz=2):
+    indent = " " * indent_sz
+    disp = []
+    data = organize_messages_by_importance(messages)
+    for i, (importance, imp_data) in enumerate(data.items()):
+        disp.append(f"{i}.  {importance.name}")
+        disp.append("-" * (len(importance.name) + 4))
+        for ii, (check_name, check_data) in enumerate(imp_data.items()):
+            disp.append(f"{i}.{ii}.  {check_name}")
+            for file, file_messages in check_data.items():
+                counter = 0
+                for message in file_messages:
+                    disp.append(
+                        f"{indent}{i}.{ii}.{counter}.  {file}:{message.location}{message.object_name} -"
+                        f" {message.message}"
+                    )
+                    counter += 1
+        disp.append("")
+    return disp
+
+
 @click.command()
 @click.argument("path")
 @click.option("-m", "--modules", help="Modules to import prior to reading the file(s).")
 @click.option("--no-color", help="Disable coloration for console display of output.", is_flag=True)
 @click.option(
-    "--report-file-path",
-    default=None,
-    help="Save path for the report file.",
-    type=click.Path(writable=True),
+    "--report-file-path", default=None, help="Save path for the report file.", type=click.Path(writable=True),
 )
 @click.option("-o", "--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("-i", "--ignore", help="Comma-separated names of checks to skip.")
@@ -96,14 +127,16 @@ def inspect_all_cli(
     else:
         config = None
 
-    messages = list(inspect_all(
-        path,
-        modules=modules,
-        config=config,
-        ignore=ignore if ignore is None else ignore.split(","),
-        select=select if select is None else select.split(","),
-        importance_threshold=Importance[threshold],
-    ))
+    messages = list(
+        inspect_all(
+            path,
+            modules=modules,
+            config=config,
+            ignore=ignore if ignore is None else ignore.split(","),
+            select=select if select is None else select.split(","),
+            importance_threshold=Importance[threshold],
+        )
+    )
     if json_file_path is not None:
         with open(json_file_path, "w") as fp:
             json.dump(messages, fp, cls=InspectorOutputJSONEncoder)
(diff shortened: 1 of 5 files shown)


bendichter:
@CodyCBakerPhD 

There is a usability issue with the current way to generate reports. What if the user has 100+ files? Maybe on the first 50 or so there are no critical errors, but there is a critical error on file 51? This would be difficult for the user to spot, because they'd probably scroll through the first few files and assume that they are all the same. To address this, we could display the results in a different order, e.g.:

```
0.  CRITICAL
------------
0.0.  check_data_orientation
  0.0.0.  test:/processing/behavior/Position/my_spatial_series - Data may be in the wrong orientation. Time should be in the first dimension, and is usually the longest dimension. Here, another dimension is longer.
0.1.  check_timestamps_match_first_dimension
  0.1.0.  test:/acquisition/test_time_series_3 - The length of the first dimension of data does not match the length of timestamps.
  0.1.1.  test2.nwb:/acquisition/test_time_series_3 - The length of the first dimension of data does not match the length of timestamps.

1.  BEST_PRACTICE_VIOLATION
---------------------------
1.0.  check_regular_timestamps
  1.0.0.  test:/acquisition/test_time_series_2 - TimeSeries appears to have a constant sampling rate. Consider specifying starting_time=1.2 and rate=2.0 instead of timestamps.

2.  BEST_PRACTICE_SUGGESTION
----------------------------
2.0.  check_small_dataset_compression
  2.0.0.  test:/acquisition/test_time_series_1 - data is not compressed. Consider enabling compression when writing a dataset.
```

This would gather critical errors across all files to the top, and would aggregate them by check.

In order to accomplish this, I reorganized the code to return a flat list of messages that includes "file", and delay the organization part, since these two different printing methods require different organization. Once that was done, it made sense to `yield` all messages, since Satra asked for that and it seemed like it might be a useful feature. Let's discuss in our meeting today.

CodyCBakerPhD:
@bendichter Yeah, looking over this now. Mostly looks good; was there a reason the severity levels had to change?

bendichter:
I wanted to be able to sort messages by the value of severity: `sorted(messages, key=lambda x: x.severity.value)`

CodyCBakerPhD:
Right, but the actual values for the severity comparison changed:

https://github.com/NeurodataWithoutBorders/nwbinspector/pull/89/files#diff-cfa07db9e698a249418609afea3dcd87686a8f03b6217ca57e75e6f50579d6c6L28-R29

This means a dataset compression below 1 GB would now get ranked below missing metadata, for instance.

bendichter:
@CodyCBakerPhD should "low" be above "no severity"? I find that pretty counter-intuitive.

CodyCBakerPhD:
The way we've treated it thus far - Most of the time, severity is `NO_SEVERITY` because there is no basis for comparison, as is the case with missing metadata.

Basically there are two intentions (a) the act of specifying severity at all means you care about the ordering of this check relative to the average check that doesn't have ranked severity, (b) for two of the same checks raised on the same exact file but with clearly different magnitude and order of priority in fixing them (e.g., a 15GB dataset not compressed vs. a 0.8GB dataset not compressed), the higher severity one is ranked higher.

Point (b) makes the logical use of severity comparisons specific to an individual check irrespective of the severities specified by other checks of the same importance level.

Whereas by setting the baseline to be between HIGH and LOW, when writing a new check we would have to think about where it would fit into the global priority order of checks within that importance, which is a more difficult task as the number of checks grows.

bendichter:
@CodyCBakerPhD all severity values have numerical values, which means they are being compared to each other and there really isn't a "no severity." Maybe the best would be to have severity = low, medium (default) or high

CodyCBakerPhD:
> all severity values have numerical values, which means they are being compared to each other and there really isn't a "no severity." Maybe the best would be to have severity = low, medium (default) or high

That is indeed effectively how it is; the point to `NO_SEVERITY` as the default, however, is that the majority of checks do not specify any `severity`, and not specifying at all (the default) should still be the lowest case.

It's extra work during check development and review to have to think about whether the output of a check should be manually ranked _lower_ than the average of all other checks of that importance, which makes it harder to add new checks. Especially for really basic stuff like missing metadata. The only time `severity` is intended to be specified and thought about is when there's some quantitative basis for comparison, which is mostly for data-related checks.

If it really bothers you the simplest thing we can just remove `NO_SEVERITY` altogether and set `LOW` to be the default for everything.

bendichter:
> If it really bothers you the simplest thing we can just remove NO_SEVERITY altogether and set LOW to be the default for everything.

I'm fine with that

CodyCBakerPhD:
Putting this in draft while I work on splitting it into separate PRs for easier review and integration.

nwbinspector/nwbinspector.py
```diff
@@ -34,6 +35,54 @@ def default(self, o):
...
+    out = {}
+    for file in files:
+        out[file] = {}
+        file_messages = list(filter(lambda x: x.file == file, messages))
+        for importance in sorted(set(message.importance for message in file_messages), key=lambda x: -x.value):
+            out[file][importance] = sorted(
+                filter(lambda x: x.importance == importance, file_messages),
+                key=lambda x: -x.severity.value,
+            )
+    return out
```
CodyCBakerPhD, line 49: After testing this, one thing to note is the time complexity involved in applying a `filter` that iterates a list checking for a potentially small number of boolean matches is sub-optimal. Here's a self-contained comparison that can be applied to any directory with a decent number of files in it

```python
"""Simple script for testing performance of two iteration methods for NWBInspector."""
from time import time
from pathlib import Path

import numpy as np

files = [x.name for x in Path(".").iterdir()]
nfiles = len(files)
messages = [
    dict(msg="foo", file=files[np.random.randint(low=0, high=nfiles)])
    for _ in range(nfiles * 50)  # average of 50 messages per file; feel free to change
]

start = time()
out1 = {file: list(filter(lambda x: x["file"] == file, messages)) for file in files}
end = time()
time1 = end - start

start = time()
out2 = {file: list() for file in files}
for message in messages:
    out2[message["file"]] = message
end = time()
time2 = end - start
```

For my system at least, the binning method (proposed option 2) is 10x faster.

CodyCBakerPhD: As such, I was going to refactor this type of functionality like

```python
def organize_messages_by_file(messages: List[InspectorMessage]):
    """Order InspectorMessages by file name then importance."""
    files = natsorted(set(message.file for message in messages))
    importance_levels = [Importance.CRITICAL, Importance.BEST_PRACTICE_VIOLATION, Importance.BEST_PRACTICE_SUGGESTION]
    out = {file: {importance: list() for importance in importance_levels} for file in files}  # always add the small number of importance bins
    for message in messages:  # linear time, single pass through message list
        out[message.file][message.importance].append(message)
    for file in files:
        for importance in importance_levels:  # Worst case complexity: (number of files) x 3
            if out[file][importance]:
                out[file][importance] = sorted(out[file][importance], key=lambda x: -x.severity.value) 
            else:
                out[file].pop(importance)
    return out
```

Basically, the more pre-sorting you can apply to the output structure ahead of actual iteration, the better.

@bendichter What do you think? Any other suggestions for improvement?

bendichter: @CodyCBakerPhD how about this as a general solution

```python
def hier_org(objects, levels):
    unique_attrs = sorted(set(getattr(object, levels[0]) for object in objects))
    if len(levels) > 1:
        return {attr: hier_org([obj for obj in objects if getattr(obj, levels[0]) == attr], levels[1:])
            for attr in unique_attrs}
    else:
        return {attr: [obj for obj in objects if getattr(obj, levels[0]) == attr] for attr in unique_attrs}
    

# test
from collections import namedtuple

Obj = namedtuple("Obj", "a b c")

objects = []
for a in range(3):
    for b in range(3):
        for c in range(3):
            objects.append(Obj(a=a,b=b,c=c))
            
hier_org(objects, ["a", "b", "c"])
```

```
{0: {0: {0: [Obj(a=0, b=0, c=0)],
   1: [Obj(a=0, b=0, c=1)],
   2: [Obj(a=0, b=0, c=2)]},
  1: {0: [Obj(a=0, b=1, c=0)],
   1: [Obj(a=0, b=1, c=1)],
   2: [Obj(a=0, b=1, c=2)]},
  2: {0: [Obj(a=0, b=2, c=0)],
   1: [Obj(a=0, b=2, c=1)],
   2: [Obj(a=0, b=2, c=2)]}},
 1: {0: {0: [Obj(a=1, b=0, c=0)],
   1: [Obj(a=1, b=0, c=1)],
   2: [Obj(a=1, b=0, c=2)]},
  1: {0: [Obj(a=1, b=1, c=0)],
   1: [Obj(a=1, b=1, c=1)],
   2: [Obj(a=1, b=1, c=2)]},
  2: {0: [Obj(a=1, b=2, c=0)],
   1: [Obj(a=1, b=2, c=1)],
   2: [Obj(a=1, b=2, c=2)]}},
 2: {0: {0: [Obj(a=2, b=0, c=0)],
   1: [Obj(a=2, b=0, c=1)],
   2: [Obj(a=2, b=0, c=2)]},
  1: {0: [Obj(a=2, b=1, c=0)],
   1: [Obj(a=2, b=1, c=1)],
   2: [Obj(a=2, b=1, c=2)]},
  2: {0: [Obj(a=2, b=2, c=0)],
   1: [Obj(a=2, b=2, c=1)],
   2: [Obj(a=2, b=2, c=2)]}}}
```

CodyCBakerPhD: That looks like it should cover my next question if implemented correctly, which was to have a single `organize_messages` function that had an argument that could dynamically specify the order in which we want things organized (`["file", "importance"]`, or `["importance", "neurodata_type", "importance"]`, etc.)

I'll play around with this a bit and get back to you!

CodyCBakerPhD: @bendichter That code had some great notions of recurrence logic in it, but it still suffered from the same nested iteration as before due to the line

```python
        return {attr: hier_org([obj for obj in objects if getattr(obj, levels[0]) == attr], levels[1:])
            for attr in unique_attrs}
```

which still has iterations over **every** `obj` for **each** `unique_aatr` to extract the subset that correspond to a particular `attr`. A time complexity analysis of this shows a worst-case time of 

image

which will tend to be the two largest values of 

image

image

I took this logic and made two versions seen on this commit on another branch: https://github.com/NeurodataWithoutBorders/nwbinspector/commit/21a30aadaf1e0f51cc5bfd183b79fa181315a233

First version 'fancy' is pretty much that same exact code but built on our objects +/- some Enum logic for sorting.

Second version 'efficient' is my binned approach that separates the dictionary construction from the data filtering, though it needs some helper functions for indexing arbitrarily nested dictionaries. This decoupling approach improves the time complexity to

image

To test performance, I downloaded DANDISet `000004` and set up the following

```python
messages = list()
for folder in Path(".../000004").iterdir():
    messages.extend(inspect_all(path=folder))
```

which gives us a list of 611 InspectorMessages. Now

```python
levels = ["file", "importance"]
 # just for testing, technically any order and combination of attributes for an InspectorMessage can be used

start = time()
res1 = fancy_organize_messages(messages=messages, levels=levels)
end = time()
time1 = end - start

start = time()
res2 = efficient_organize_messages(messages=messages, levels=levels)
end = time()
time2 = end - start
```

As before, the 'efficient' method is clocking much faster (about 6x on the case of real data), so I'll proceed with adding tests and integrating that method.

Overall very cool advancement that allows arbitrary user choice of output structure, if they want any.

bendichter: how about this. I create a `objects_by_attr` dict so that the objects are only looped over once per recursion

```python
from collections import defaultdict

def hier_org(objects, levels, reverse=None):
    if reverse is None:
        reverse = [False] * len(levels)
    unique_attrs = sorted(set(getattr(object, levels[0]) for object in objects))
    if reverse[0]:
        unique_attrs = unique_attrs[::-1]
    objects_by_attr = defaultdict(list)
    [objects_by_attr[getattr(obj, levels[0])].append(obj) for obj in objects]
    if len(levels) > 1:
        return {attr: hier_org(objects_by_attr[attr], levels[1:], reverse[1:]) for attr in unique_attrs}
    else:
        return {attr: objects_by_attr[attr][0] for attr in unique_attrs}
    
    

# test
from collections import namedtuple

Obj = namedtuple("Obj", "a b c")

objects = []
for a in range(3):
    for b in range(3):
        for c in range(3):
            objects.append(Obj(a=a, b=b, c=c))
            
hier_org(objects, ["a", "b", "c"], [True, False, True])
```

CodyCBakerPhD:
All relevant code changes proposed by the PR have been merged along separate channels. Closing this now.

PR #91, opened 2022-03-08
removed NO_SEVERITY; adjusted tests

CodyCBakerPhD:
## Motivation

`Severity.NO_SEVERITY` as default was confusing; the new default is `LOW`.

### DIFF
--- a/nwbinspector/register_checks.py
+++ b/nwbinspector/register_checks.py
@@ -120,7 +120,7 @@ def auto_parse(check_function, obj, result: Optional[InspectorMessage] = None):
     """Automatically fill values in the InspectorMessage from the check function."""
     if result is not None:
         auto_parsed_result = result
-        if auto_parsed_result.severity not in [Severity.LOW, Severity.HIGH]:
+        if not isinstance(auto_parsed_result.severity, Severity):
             raise ValueError(
                 f"Indicated severity ({auto_parsed_result.severity}) of custom check "
                 f"({check_function.__name__}) is not a valid severity level! Please choose one of "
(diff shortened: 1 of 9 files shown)


nwbinspector/register_checks.py
```diff
@@ -121,9 +120,7 @@ def auto_parse(check_function, obj, result: Optional[InspectorMessage] = None):
     """Automatically fill values in the InspectorMessage from the check function."""
     if result is not None:
         auto_parsed_result = result
-        if auto_parsed_result.severity is None:  # For perfect consistency with not specifying
-            auto_parsed_result.severity = Severity.NO_SEVERITY
-        if auto_parsed_result.severity not in Severity:
+        if auto_parsed_result.severity not in [Severity.LOW, Severity.HIGH]:
```
bendichter: Would keeping `if auto_parsed_result.severity not in Severity` be OK?

bendichter: or `if isinstance(auto_parsed_result.severity, Severity)`

bendichter (approved):

nwbinspector/register_checks.py
CodyCBakerPhD: So this is a silly little thing that I noticed when I changed this syntax similarly for checking `Importance` a while back. Technically, if you're just talking about Python 3.7 they do the same exact thing.

But for 3.8 and beyond you get a much less informative error message if you make a mistake like `severity="LOW"` instead of `severity=SomeOtherEnumClass.LOW`.

So this minor swap in syntax lets me get rid of an entire test from the suite, and removed the platform dependence from testing as well, and will always give the more informative error message.

CodyCBakerPhD: Looks like `isinstance(auto_parsed_result.severity, Severity)` works well across platforms, though, I'll just swap to that, then. Good catch!

PR #92, opened 2022-03-08
Modular PR: flatten output and generalize organization

CodyCBakerPhD:
## Motivation

Separates the flattening of PR #89 from the yielding also included within that PR.

Also incorporates the more general organization structure along with tests for it and some minor debugs to allow those tests to pass.

## Checklist

### DIFF
--- a/nwbinspector/inspector_tools.py
+++ b/nwbinspector/inspector_tools.py
@@ -5,7 +5,6 @@ from collections import OrderedDict
 from typing import Dict, List
 from pathlib import Path
 from natsort import natsorted
-from copy import copy
 
 import numpy as np
 
@@ -23,16 +22,16 @@ def sort_by_descending_severity(check_results: list):
 def organize_messages_by_file(messages: List[InspectorMessage]):
     """Order InspectorMessages by file name then importance."""
     files = natsorted(set(message.file for message in messages))
-    messages_by_file = {file: list for file in files}
-    out = copy(messages_by_file)
+    importance_levels = [Importance.CRITICAL, Importance.BEST_PRACTICE_VIOLATION, Importance.BEST_PRACTICE_SUGGESTION]
+    out = {file: {importance: list() for importance in importance_levels} for file in files}
     for message in messages:
-        messages_by_file[message.file] = message
-    for file, messages_for_file in messages_by_file.items():
-        realized_importance = sorted(set([message.importance for message in messages_for_file]), key=lambda x: -x.value)
-        messages_per_importance = {importance: [] for importance in realized_importance}
-        for message in messages_for_file:
-            messages_per_importance[message.importance]
-        out[message.file] = sorted(messages_per_importance, key=lambda x: -x.severity.value)
+        out[message.file][message.importance].append(message)
+    for file in files:
+        for importance in importance_levels:
+            if out[file][importance]:
+                out[file][importance] = sorted(out[file][importance], key=lambda x: -x.severity.value)
+            else:
+                out[file].pop(importance)
     return out
 
 
--- a/nwbinspector/inspector_tools.py
+++ b/nwbinspector/inspector_tools.py
@@ -11,7 +11,23 @@ from .utils import FilePathType
 
 
 def organize_messages(messages: List[InspectorMessage], levels: List[str]):
-    """General function for organizing list of InspectorMessages into a nested dictionary structure."""
+    """
+    General function for organizing list of InspectorMessages.
+
+    Returns a nested dictionary organized according to the order of the 'levels' argument.
+
+    Parameters
+    ----------
+    messages : list of InspectorMessages
+    levels: list of strings
+        Each string in this list must correspond onto an attribute of the InspectorMessage class, excluding the
+        'message' text.
+    """
+    assert "message" not in levels, (
+        "You must specify levels to organize by that correspond to attributes of the InspectorMessage class, not "
+        "including the text message."
+    )
+
     unique_values = list(set(getattr(message, levels[0]) for message in messages))
     sorted_values = sort_unique_values(unique_values)
     if len(levels) > 1:
--- a/nwbinspector/inspector_tools.py
+++ b/nwbinspector/inspector_tools.py
@@ -23,7 +23,11 @@ def organize_messages(messages: List[InspectorMessage], levels: List[str]):
         }
     else:
         return {
-            value: [message for message in messages if getattr(message, levels[0]) == value] for value in sorted_values
+            value: sorted(
+                [message for message in messages if getattr(message, levels[0]) == value],
+                key=lambda x: -x.severity.value,
+            )
+            for value in sorted_values
         }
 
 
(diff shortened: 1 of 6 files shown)


CodyCBakerPhD:
Tomorrow I'll 

- [x] Add a battery of tests for the new organization functionality
- [x] Fix the inspector test on 'true' log output

In a future PR I will...
- [ ] Expose `levels` to CLI
- [ ] This requires a bit more generalizing of those types of functions, this time for displaying the arbitrarily organized dictionary
- [ ] add more examples of log files using different methods of structuring for demonstration purposes to give ourselves and users a better feel for alternative organization methods and some might be more beneficial.

nwbinspector/register_checks.py
```diff
@@ -62,6 +60,8 @@ class InspectorMessage:
         The name of the instantiated object being inspected.
     location : str
         The location relative to the root of the NWBFile where the inspected object may be found.
+    file : str
+        The path to the NWBFile this message pertains to.
```
CodyCBakerPhD:
```suggestion
    file : str
        The name of the NWBFile this message pertains to.
```

Upon using this code for the past day, I actually think there is sufficient identifiability just from the file name alone. 

Two questions then,

(a) Should this attribute be renamed to (i) `filename`, (ii) `file_name`, (iii) or `file_path` to make it clear it's not an open `file` object (one of the other common uses of that keyword)?

(b) Should we include or clip the `.nwb`; i.e., use the `.stem` or the `.name` of the `Path` object?

nwbinspector/inspector_tools.py
```diff
@@ -1,34 +1,80 @@
...
+
+    Returns a nested dictionary organized according to the order of the 'levels' argument.
+
+    Parameters
+    ----------
+    messages : list of InspectorMessages
+    levels: list of strings
+        Each string in this list must correspond onto an attribute of the InspectorMessage class, excluding the
+        'message' text.
+    """
```
CodyCBakerPhD: @bendichter Amongst your suggestions for this new function was a `reverse` option for flipping the direction of ordering along a particular nesting level.

(a) Can you think of situations where this might be wanted by a user or ourselves?

(b) Did you want that exposed to the CLI as well (TODO in a future PR)? If so what tag, `-r`/`--reverse`? Encoded as a comma-separated string equal in length to `--levels`?

bendichter: Yes this is needed, some we want the results to show up from high to low. Another option is to do the reversing in the report creation function

bendichter: I don't think this should be in the CLI

nwbinspector/register_checks.py
CodyCBakerPhD: @bendichter Thoughts on this or good as is?

nwbinspector/inspector_tools.py
CodyCBakerPhD: I'll add reverse to the organization, then, but not to the CLI

CodyCBakerPhD:
@bendichter Added `reverse` with tests. Just need confirmation on `file` stuff above to finish this PR to move onto the others

nwbinspector/register_checks.py
bendichter: "filename" or "file_name" would be fine. Let's not call it a path unless it is truly a path. I think we should keep the .nwb

CodyCBakerPhD: A quick glance around other Python references, seems `filename` is the most common, so going with that.

CodyCBakerPhD:
@bendichter Swapped to `InspectorMessage.filename`.

Ready for final review.

bendichter:
Do we really want to assume that filenames are unique? It's possible for them to be non-unique, if they are nested in folders.

CodyCBakerPhD:
@bendichter That really doesn't seem like it would be best practice.

(a) If you had two NWBFiles with the same name, but different locations, what would happen if you simply moved them into the same folder? Or if the full path is that important to identifying  context for the file, what if it was simply misplaced?

(b) DANDI folders will always be unique by design.

(c) Have you ever seen a case of this in practice? Labs I've seen are generally very good (sometimes too good...) at making important file names unique based on combinations of session ID, datetime of session start, or subect ID.

(d) Current globbing protocol in `inspect_all` only looks at a single-level folder depth, so I would certainly hope they are unique with that.

For (d), a user could *technically* go beyond our standard library usage to assemble their own list of messages through repeated calls to `inspect_*`, and if some of the filenames (if not using best practices of data management) are indeed the same, that would be a counter-example of uniqueness when organizing. Note however they could also change their relative path at any point, so if we wanted the `InspectorMessage` to track a full `file_path`, it would have to be absolute to avoid the same issue.

bendichter:
> (d) Current globbing protocol in inspect_all only looks at a single-level folder depth, so I would certainly hope they are unique with that.

This was an oversight. I just submitted  PR to fix it.


Let's compromise. The inspector can store the relative path wrt to supplied directory. That way if the directory is indeed flat then the relative path == the filename, and if there is a nested structure that is displayed

CodyCBakerPhD:
> > (d) Current globbing protocol in inspect_all only looks at a single-level folder depth, so I would certainly hope they are unique with that.
> 
> This was an oversight. I just submitted PR to fix it.
> 
> Let's compromise. The inspector can store the relative path wrt to supplied directory. That way if the directory is indeed flat then the relative path == the filename, and if there is a nested structure that is displayed

OK - as long as you're aware they can do something like the following

```python
all_messages = []
all_message.extend(list(inspect_all(path=".", ...)))
# change working directory
all_message.extend(list(inspect_all(path=".", ...)))
```
and now the relative paths in `all_messages` are meaningless; would we simply call this user error or misuse of our library?

bendichter:
it's not based on the working directory, it's based on the `path` arg of `inspect_all`

bendichter:
oh, I see. You are right, if they use "." for the path, and change their working directory mid-run, and have non-unique filenames then the above workflow will have incorrect results. I'm ok with that, I don't think it's a problem.

CodyCBakerPhD:
> it's not based on the working directory, it's based on the `path` arg of `inspect_all`

Right, sorry for any confusion with how I was altering the path argument in that example. Here is a more explicit one.

```python
all_messages = []
all_message.extend(list(inspect_all(path=path1)))
all_message.extend(list(inspect_all(path=path2)))
```

And if we're worried about non-unique file names, we should be worried about non-unique folder names, too. So if we make it a `file_path` with a relative path according to `path` being globbed by each `insect_all` calls, two different `inspect_all` results collected in the same list of messages could potentially be non-unique.

Explicit example of such a structure:

`D:/test1/test3/test4.nwb` with `inspect_all` called from `path1=D:/test1`
`E:/test2/test3/test4.nwb` with `inspect_all` called from `E:/test2`

Edit:

> oh, I see. You are right, if they use "." for the path, and change their working directory mid-run, and have non-unique filenames then the above workflow will have incorrect results. I'm ok with that, I don't think it's a problem.

OK, making it a relative path then.

CodyCBakerPhD:
@bendichter relative `file_path` implemented and tested.

Summary of changelog for PR:

+110/-60: code changes for flattening raw output from inspection functions, adding file_path, and generalized organization of raw output

+340/-40: tests, mostly for new organization, but also for file_path and raw output structure

bendichter (approved):

PR #94, opened 2022-03-09
added parallelization

### LINKED ISSUE #83: [Feature]: parallelize over NWB files

the -j --njobs flag should control number of simultaneous jobs that run inspector over NWB files in parallel

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #83 

## Motivation

@bendichter Added simple parallelization via ~~`joblib`~~ `concurrent.futures`. Not the most sophisticated we could possibly do, but from past experience dealing with parallel access within and across NWBFiles, this gets the job done best with the least amount of hassle.

Already seeing a massive speedup when applying this to DANDISet `000004` (87 files) even just using 4 jobs. Run it on the Hub with 80+ jobs and I'm sure it'll feel practically instant.

## How to test the behavior?

Added new tests to assure the output matches when `n_jobs` passed to the library function.

The one weird thing to note, because I can't seem to get it to shape the way I wanted...

When calling `inspect_all` without parallel, what we've been doing to force the yields to collapse is calling 
```python
list(inspect_all(path=...))
```

This gives a simple, flat list of InspectorMessages same as `inspect_nwb`, but of course this one will have messages that span multiple files.

Now, when calling `inspect_all` with `n_jobs`, I have to do the following to reproduce this structure, otherwise I end up with a list-of-lists, the outer list being the number of NWBFiles in the iteration path, and the inner list being the list of actual messages.

```python
            test_results = []
            for test_result in inspect_all(
                path=Path(self.nwbfile_paths[0]).parent, select=[x.__name__ for x in self.checks], n_jobs=2
            ):
                test_results.extend(test_result)
```

This is likely due to the way I constructed the dispatcher

```python
        def dispatch_inspection(nwbfile_path: FilePathType, checks: list):
            # can't be a generator b/c it needs to be pickled
            return list(inspect_nwb(nwbfile_path=nwbfile_path, checks=checks))

        for message in Parallel(n_jobs=n_jobs)(
            delayed(dispatch_inspection)(nwbfile_path=nwbfile_path, checks=checks) for nwbfile_path in nwbfiles
        ):
            yield message
```

And it doesn't seem to matter if you try to `return` instead of `yield` the message in that last step; the function seems to always behave like a generator when you call it, although the `Parallel` doesn't dispatch until the first call to iterate that generator (including even a single `next`), and then it iterates over all the `nwbfiles` at once.

## Checklist

## Still Need To...

- [ ] add `-j, --n-jobs` flag to CLI
- [ ] add tests for CLI usage of the multiple jobs (kind of tough to actually make sure it's using multiple processes, but at least we can trust the argument passing and just ensure output matches expectations).

### DIFF
--- a/tests/test_inspector.py
+++ b/tests/test_inspector.py
@@ -107,6 +107,12 @@ class TestInspector(TestCase):
         for dictionary in true_list:
             self.assertIn(member=dictionary, container=test_list)
 
+    def assertListsEquivalent(self, test_list: list, true_list: list):
+        for member in test_list:
+            self.assertIn(member=member, container=true_list)
+        for member in true_list:
+            self.assertIn(member=member, container=test_list)
+
     def assertFileExists(self, path: FilePathType):
         path = Path(path)
         assert path.exists()
@@ -195,7 +201,7 @@ class TestInspector(TestCase):
                 file="testing1.nwb",
             ),
         ]
-        self.assertListEqual(list1=test_results, list2=true_results)
+        self.assertListsEquivalent(test_list=test_results, true_list=true_results)
 
         def test_inspect_all_parallel(self):
             test_results = []
@@ -264,7 +270,7 @@ class TestInspector(TestCase):
                     file="testing1.nwb",
                 ),
             ]
-            self.assertListEqual(list1=test_results, list2=true_results)
+            self.assertListsEquivalent(test_list=test_results, true_list=true_results)
 
     def test_inspect_nwb(self):
         test_results = list(inspect_nwb(nwbfile_path=self.nwbfile_paths[0], checks=self.checks))
@@ -314,7 +320,7 @@ class TestInspector(TestCase):
                 file="testing0.nwb",
             ),
         ]
-        self.assertListEqual(list1=test_results, list2=true_results)
+        self.assertListsEquivalent(test_list=test_results, true_list=true_results)
 
     def test_inspect_nwb_importance_threshold(self):
         test_results = list(
@@ -345,7 +351,7 @@ class TestInspector(TestCase):
                 file="testing0.nwb",
             ),
         ]
-        self.assertListEqual(list1=test_results, list2=true_results)
+        self.assertListsEquivalent(test_list=test_results, true_list=true_results)
 
     def test_command_line_runs_cli_only(self):
         console_output_file = self.tempdir / "test_console_output.txt"
@@ -410,7 +416,7 @@ class TestInspector(TestCase):
                 file="testing0.nwb",
             ),
         ]
-        self.assertListEqual(list1=test_results, list2=true_results)
+        self.assertListsEquivalent(test_list=test_results, true_list=true_results)
 
 
 def test_configure_checks():
--- a/tests/test_inspector.py
+++ b/tests/test_inspector.py
@@ -18,7 +18,7 @@ from nwbinspector import (
     check_data_orientation,
     check_timestamps_match_first_dimension,
 )
-from nwbinspector.nwbinspector import inspect_nwb, configure_checks
+from nwbinspector.nwbinspector import inspect_all, inspect_nwb, configure_checks
 from nwbinspector.register_checks import Severity, InspectorMessage, register_check
 from nwbinspector.utils import FilePathType
 from nwbinspector.tools import make_minimal_nwbfile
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
> add -j, --n-jobs flag to CLI

@bendichter Ohh, uhh... It seems that `-j` has been reserved now for the `json-file-path`. What else would you prefer for this CLI argument?

An alternative I might suggest would be to swap the `--json-file-path` to a `--format` (`-f`) option that can be either `.txt` or `.json` and uses the `--report-file-path` as the save location. This also opens the possibility for other future formats such as `.md`, `.rst`, etc.

bendichter:
@CodyCBakerPhD I'm fine with moving CLI args around, but I was thinking we'd use format to distinguish between file-based and importance-based reports.

bendichter:
I like sklearn's approach to setting number of jobs [here](https://scikit-learn.org/stable/glossary.html#term-n_jobs)

bendichter:
You might be able to get around the pickle issue with https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

CodyCBakerPhD:
> @CodyCBakerPhD I'm fine with moving CLI args around, but I was thinking we'd use format to distinguish between file-based and importance-based reports.

Wouldn't that type of organization would be controlled by `-l`,`--levels`?

file-based: `--levels "file, importance,..."
importance-based: `--levels "importance, file,..."`

CodyCBakerPhD:
> I like sklearn's approach to setting number of jobs [here](https://scikit-learn.org/stable/glossary.html#term-n_jobs)

You mean the `-1`/`-2` behavior? Pretty sure that comes along with standard passing into `joblib`. We already skip `joblib` usage if `n_jobs=1`.

CodyCBakerPhD:
> You might be able to get around the pickle issue with https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

Looks promising, trying it now.

bendichter:
Great! It's something to try but it's a nice-to-have, not worth holding up the project. Your current solution is already pretty good.

CodyCBakerPhD:
@bendichter Actually I just got it working; it's (a) easier to code (b) allows the exact same calling syntax to `inspect_all`, and (c) comes with standard Python. It's pretty nice actually.

Again using dandiset `000004` as a good test due to large file count, small file size, and rich output reports...

On my system, a single job scanning the entire directory took about 87 seconds to complete, so about 1 second per NWBFile.
With `n_jobs=4`, it went down to 15 seconds. That's better than 87/`n_jobs`, so I didn't even know speedups like that could happen 🤣

bendichter:
awesome!

CodyCBakerPhD:
For the CLI, I've simply set it to `--n-jobs` for now and we can discuss rearranging the single-letter values later on when we have the `--levels` implemented.

tests/test_inspector.py
```diff
@@ -101,6 +100,12 @@ def setUpClass(cls):
     def tearDownClass(cls):
         rmtree(cls.tempdir)
 
+    def assertListsEquivalent(self, test_list: list, true_list: list):
+        for member in test_list:
+            self.assertIn(member=member, container=true_list)
+        for member in true_list:
+            self.assertIn(member=member, container=test_list)
```
bendichter: Are you trying to test that the lists have the same elements, but may be in a different order? If so, this is probably better: `assert set(test_list) == set(true_list)` or perhaps you want [`assertCountEqual`](https://docs.python.org/3/library/unittest.html?highlight=assertcountequal#unittest.TestCase.assertCountEqual)

CodyCBakerPhD: `self.assertCountEqual` works perfectly, so I've swapped to that. It does print a lot of results in the output though, so I had to set the `maxDiff` flag to `None` just to view it on some simple cases.

I would prefer the `set` equal way, but unfortunately, casting a list as a set requires all the contained objects to be hashable, which our `InspectorMessages` are not. They do, however, have an equality method set by the `dataclass`, so you are able to test if two `InspectorMessages` are exactly equal (_i.e._, they have equal contents).

bendichter (approved):

bendichter:
Codecov is saying we are missing a test for organize_messages_by_importance, but that's a separate issue

bendichter:
Also, readthedocs is failing to build, but I have a separate PR that repairs that

PR #95, opened 2022-03-09
recursive file search

bendichter:


nwbinspector/nwbinspector.py
```diff
@@ -112,7 +112,7 @@ def inspect_all(
 
     in_path = Path(path)
     if in_path.is_dir():
-        nwbfiles = list(in_path.glob("*.nwb"))
+        nwbfiles = list(in_path.glob("**/*.nwb", recursive=True))
```
CodyCBakerPhD:
```suggestion
        nwbfiles = list(in_path.glob("**/*.nwb"))
```

Doesn't take this option - looks like it's recursive over sub-directories by default with the `**/`: https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob

```suggestion
        nwbfiles = list(in_path.rglob("*.nwb"))
```

CodyCBakerPhD (approved):

PR #96, opened 2022-03-09
[Bug]: Configuration fixes and tests

CodyCBakerPhD:
## Motivation

More robust testing suite for the check configuration behavior, which has revealed a handful of unintended issues.

Issues:

- SKIP wasn't actually skipping the check, it was still getting added to the output list of modified checks
- The schema validation was not complaining about incorrect/unused keys on the outer level; fixed by modifying the schema with `additionalProperties=False`
- altering the `.importance` level attribute of a `check` during configuration also permanently altered it within the base check registry, which not only messes up the tests in rather obscure ways but also seems unintended. My approach to fixing this is to fill the `checks_out` list returned by the configuration with copies of the check functions in the base registry, whose attribute can then be changed at will without altering the original check functions. Not 100% working just yet because it's setting the copy to the output decorator, so I'll have to think about how to get around that...

## Checklist


CodyCBakerPhD:
@bendichter Good to merge these fixes?

bendichter (approved):

PR #102, opened 2022-03-14
[Feature]: add internal configs

bendichter:
## Motivation

I'd like for the user to be able to just say `--config dandi` and have that yaml stored internally. They retain the ability of using a config yaml that is not stored internally. This feature still needs to be tested.


CodyCBakerPhD:
Looks fine to me in theory, but the addition of testing (which we should) will require some of the fixes from #96 so I'll see if I can get that done before our meeting

CodyCBakerPhD:
To-Do:

- [ ] add tests for dandi configuration; new testing suite only checks behavior of general config logic

CodyCBakerPhD:
Tested it locally on some actual NWBFiles; had to do a simple debug to point to the correct path but otherwise seems to be working as intended.

Writing explicit tests is proving a tad tricky, will take a bit.

Also observing a very bizarre behavior where if there are actual Errors (like a `ValueError`) triggered within one of the test functions (see PR #106 that fixes the ones I'm seeing), they aren't grabbed within the `try/except` of `run_checks` and included in the report, but instead are raised to console as true errors:

image

but if I don't include the `--config dandi` they behave as expected and go into that extra section of the final report.

While simply merging #106 will fix this particular issue, we should be aware of this behavior more generally. I tried looking through the logic of our `config` argument passing but I don't see anything obvious that would cause this.

CodyCBakerPhD:
Either way, it's not an issue specific to this PR but rather the CLI use of any config file. I'll keep looking into it but I tested these particular changes and they worked.

CodyCBakerPhD (approved):

PR #106, opened 2022-03-15
[Bug]: Fix binary capability

CodyCBakerPhD:
## Motivation

While making the slides for the presentation, I discovered some weird bugs for the columns of tables in DANDISet `000004`. These fixes correct the issues.

I also noticed there were still some references to the report output as a 'log', which I've fixed here as well.

### DIFF
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -96,9 +96,9 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
                 continue  # already boolean, int8, or uint8
             try:
                 unique_values = np.unique(column.data[:nelems])
-            except Exception:  # some columns can contain references to other objects in the NWBFile
+            except TypeError:  # some contained objects are unhashable or have no comparison defined
                 continue
-            if len(unique_values) > 2:
+            if unique_values.size != 2:
                 continue
             parsed_unique_values = np.array(unique_values)
             if isinstance(parsed_unique_values[0], Real):  # upcast to float for comparison
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -103,7 +103,7 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
             parsed_unique_values = np.array(unique_values)
             if isinstance(parsed_unique_values[0], Real):  # upcast to float for comparison
                 parsed_unique_values = parsed_unique_values.astype(float)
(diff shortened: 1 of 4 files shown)


nwbinspector/checks/tables.py
```diff
@@ -92,15 +94,18 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
         if hasattr(column, "data") and not isinstance(column, VectorIndex):
             if np.asarray(column.data[0]).itemsize == 1:
                 continue  # already boolean, int8, or uint8
-            unique_values = np.unique(column.data[:nelems])
+            try:
+                unique_values = np.unique(column.data[:nelems])
+            except Exception:  # some columns can contain references to other objects in the NWBFile
+                continue
```
bendichter: Is there a better way to catch this? I'd rather not use try/except, but if we do can we make the exception type more specific?

bendichter: from https://docs.python.org/3/tutorial/errors.html: 
> All exceptions inherit from [BaseException](https://docs.python.org/3/library/exceptions.html#BaseException), and so it can be used to serve as a wildcard. Use this with extreme caution, since it is easy to mask a real programming error in this way!

CodyCBakerPhD: Yep, I can catch the issue with a`TypeError`. It occurs whenever the contained object has no defined basis for comparison like a ">" operator (which `np.unique` uses). Likewise, the same object is unable to be hashed so something like `np.array(set(...))` also throws a `TypeError`, so either way I think we do need this.

The same underlying problem was triggered by two cases, one where the actual raw column data was a `np.array` with `dtype="O"`/`object`, and the other was when the column data was `ElectrodeGroup` references.

CodyCBakerPhD: While we could program explicit `if` skips for those specific cases instead of the `try/except`, I feel like that's a bandaid and we'd ultimately end  up having to add more and more `if` checks for other items that can't be passed through `np.uniques` or `set`.

bendichter: what about something like `if hasattr(data, '__gt__`)`

CodyCBakerPhD: Nope🤨 the offending data objects have that defined as an attribute, but the attribute is a method that does not support inter-object comparison, and further raises a `NotImplementedError` if you really try to force it to work by setting its comparison value.

bendichter: hmm ok

bendichter (approved):

PR #115, opened 2022-03-21
[Documentation]: Polishing TimeSeries Best Practice Section

CodyCBakerPhD:
Anything else we should polish here aside from external API links (which I'm working on separately)?

### DIFF
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -1,10 +1,7 @@
 Time Series
 ===========
 
-When using :py:class:`~pynwb.base.TimeSeries` or any of its descendants, please ensure the following practices are followed.
-
-Note that these extend to any general subtype, such as :py:class:`~pynwb.ecephys.ElectricalSeries`,
-:py:class:`~pynwb.behavior.SpatialSeries`, :py:class:`~pynwb.image.ImageSeries`, etc.
+When using :py:class:`~pynwb.base.TimeSeries` or any subtype (e.g. :py:class:`~pynwb.ecephys.ElectricalSeries`, :py:class:`~pynwb.behavior.SpatialSeries`, and :py:class:`~pynwb.image.ImageSeries`), please ensure the following practices are followed.
 
 
 
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -1,9 +1,9 @@
 Time Series
 ===========
 
-Many of the neurodata_types in NWB inherit from the TimeSeries neurodata_type.
+When using ``TimeSeries`` or any of its descendants, please ensure the following practices are followed.
 
-When using TimeSeries or any of its descendants, make sure the following are followed.
+Note that these extend to any general subtype, such as ``ElectricalSeries``, ``SpatialSeries``, ``ImageSeries``, etc.
 
 
 
@@ -30,8 +30,9 @@ Units of Measurement
 
 Time-related values should always in seconds. This include ``rate`` (if applicable), which should should be in Hz.
 
-Every TimeSeries instance has a unit as an attribute of the data Dataset, which is meant to indicate the unit of
-measurement of that data. We advise using SI units.
+Every TimeSeries instance has a ``unit`` as an attribute of the ``Dataset``, which is meant to indicate the unit of
+measurement for that data, using the appropriate type from the 
+`International System of Units (SI) <https://en.wikipedia.org/wiki/International_System_of_Units>`_
 
 
 
@@ -49,36 +50,31 @@ Global Time Reference
 Subtypes
 ~~~~~~~~
 
-ElectrialSeries are reserved for neural data. ElectrialSeries holds signal from electrodes positioned in or around the brain that are monitoring neural
-activity, and only those electrodes should be in the electrodes table.
+ElectrialSeries are reserved for neural data. ElectrialSeries holds signal from electrodes positioned in or around the 
+brain that are monitoring neural activity, and only those electrodes should be in the electrodes table.
 
-
-
-.. _best_practice_time_series_break_in_continuity
-
-Breaks in Continuity
-~~~~~~~~~~~~~~~~~~~~
-TimeSeries data should be stored as one continuous stream.
-
-Data should be stored in one continuous stream, as it is acquired, not by trial as is often reshaped for analysis.
-
-Data can be trial-aligned on-the-fly using the trials table.
-
-Storing measured data as a continuous stream ensures that other users have access to the inter-trial data, and that we can align the data with whatever window they need.
-
-If you only have data spanning specific segments of time, then only include those timepoints in the data.
-
-One could use timestamps for this, even if there is a constant sampling rate within each segment, and have the timestamps correctly reflect the gaps in the recording.
+For non-neural electrodes that still may store and report raw values in Volts, simply use a general ``TimeSeries`` 
+object with ``unit`` set to ``Volts``.
 
 
 
 .. _best_practice_timestamps_ascending
 
-Ascending Timestamps
+Breaks in Continuity
 ~~~~~~~~~~~~~~~~~~~~
+TimeSeries data should generally be stored as one continuous stream as it was acquired, not by trial as is often 
+reshaped for analysis.
 
-The timestamps field, as well as the corresponding ordering of those timestamps indices in the data array, should
-always be strictly increasing.
+Data can be trial-aligned on-the-fly using the ``trials`` table.
+
+Storing measured data as a continuous stream ensures that other users have access to the inter-trial data, and that we 
+can align the data within any specifiable window.
+
+If you only have data spanning specific segments of time, then only include those timepoints in the data, see 
+:ref:`best_practice_regular_timestamps` for more information.
+
+A primary implication is that the values in ``timestamps``, as well as the corresponding ordering of their indices 
+in the ``data`` array, should always be strictly increasing.
 
 Check function: :py:meth:`~nwbinspector.checks.time_series.check_timestamps_ascending`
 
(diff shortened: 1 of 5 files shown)


nwbinspector/checks/time_series.py
```diff
@@ -46,7 +46,11 @@ def check_data_orientation(time_series: TimeSeries):
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=TimeSeries)
 def check_timestamps_match_first_dimension(time_series: TimeSeries):
-    """If the TimeSeries has timestamps, check if their length is the same as the zero-axis of data."""
+    """
+    If the TimeSeries has timestamps, check if their length is the same as the zero-axis of data.
+
+    Best Practice: :ref:`best_practice_data_orientation`
+    """
```
CodyCBakerPhD, line 53: Note that this is what we have to do in those cases where a Best Practice has multiple ways of checking for violations.

CodyCBakerPhD:
Well, I suppose we'd want to include instructions on how to fix an issue once it's caught. Lemme add those details with external tutorial links.

docs/best_practices/time_series.rst
```diff
@@ -17,76 +18,125 @@ Keep in mind that the dimensions are reversed in MatNWB, so in memory in MatNWB
...
+
+Data writers can optimize the storage of large data arrays for particular uses by using compression applied to each 
+chunk individually.
+
+This is especially important when writing NWBFiles that are intended to be uploaded to the 
+:dandi-archive:`DANDI Archive <>` for storage, sharing, and publication.
+
+For more information about how to enable compression on your data, consult the 
+:pynwb-docs:`PyNWB tutorial <tutorials/advanced_io/h5dataio.html#compression-and-other-i-o-filters>` or the
+`MatNWB instructions <https://neurodatawithoutborders.github.io/matnwb/tutorials/html/dataPipe.html#2>`_
```
CodyCBakerPhD, line 144: This is the best I could find for the MatNWB side; is there something newer? I notice this is under the older `github.io`

docs/best_practices/time_series.rst
```diff
@@ -17,76 +19,125 @@ Keep in mind that the dimensions are reversed in MatNWB, so in memory in MatNWB
...
 
 
-.. _best_practice_unit_of_measurement
+.. _best_practice_unit_of_measurement:
 
 Units of Measurement
 ~~~~~~~~~~~~~~~~~~~~
 
-Time-related values should always in seconds. This include ```rate``` (if applicable), which should should be in Hz.
+Time-related values should always in seconds. This include ``rate`` (if applicable), which should should be in Hz.
```
bendichter:
```suggestion
Time-related values should always in seconds. This includes ``rate`` (if applicable), which should should be in Hz.
```

CodyCBakerPhD:
Figured out the external links, added a couple to the header of the section, going through the rest now.

docs/best_practices/time_series.rst
```diff
@@ -17,76 +19,125 @@ Keep in mind that the dimensions are reversed in MatNWB, so in memory in MatNWB
...
 
-Use timestamps, even if there is a constant sampling rate within each segment, and have the timestamps correctly
-reflect the gaps in the recording. Use the TimeSeries.description field to explain how the data was segmented.
-If the sampling rate is constant, use rate. TimeSeries allows you to specify time using either timestamps or rate and starting_time (which defaults to 0).
-For TimeSeries objects that have a constant sampling rate, rate should be used instead of timestamps. This will ensure that you can use analysis and
-visualization tools that rely on a constant sampling rate.
+``TimeSeries`` allows you to specify time using either ``timestamps`` or ``rate`` together with ``starting_time``
+(which defaults to 0).
+
+If the sampling rate is constant, then specify the ``rate`` and ``starting_time`` instead of writing the full ``timestamps`` vector.
```
bendichter:
```suggestion
(which defaults to 0). If the sampling rate is constant, then specify the ``rate`` and ``starting_time`` instead of writing the full ``timestamps`` vector.
```

docs/best_practices/time_series.rst
```diff
@@ -1,9 +1,11 @@
 Time Series
 ===========
 
-Many of the neurodata_types in NWB inherit from the TimeSeries neurodata_type.
+When using :py:class:`~pynwb.base.TimeSeries` or any of its descendants, please ensure the following practices are followed.
+
+Note that these extend to any general subtype, such as :py:class:`~pynwb.ecephys.ElectricalSeries`,
+:py:class:`~pynwb.behavior.SpatialSeries`, :py:class:`~pynwb.image.ImageSeries`, etc.
```
bendichter:
```suggestion
When using :py:class:`~pynwb.base.TimeSeries` or any subtype (e.g. :py:class:`~pynwb.ecephys.ElectricalSeries`, :py:class:`~pynwb.behavior.SpatialSeries`, and :py:class:`~pynwb.image.ImageSeries`), please ensure the following practices are followed.
```

docs/best_practices/time_series.rst
```diff
@@ -1,92 +1,145 @@
...
 Units of Measurement
 ~~~~~~~~~~~~~~~~~~~~
 
-Time-related values should always in seconds. This include ```rate``` (if applicable), which should should be in Hz.
+<<<<<<< HEAD
+Time-related values should always in seconds. This include :py:attr:`~pynwb.base.TimeSeries.rate` (if applicable), which
+should should be in Hz.
+=======
+Time-related values should always in seconds. This includes ``rate`` (if applicable), which should should be in Hz.
+>>>>>>> 09ce3dcf67d7be26d12a4710bc5ab252d9a1c791
```
bendichter: unresolved conflict

bendichter:
Would it be possible to link to types in the schema, e.g. https://nwb-schema.readthedocs.io/en/latest/format.html#nwbfile? I'd rather that than link to PyNWB specifically

CodyCBakerPhD:
> Would it be possible to link to types in the schema, e.g. https://nwb-schema.readthedocs.io/en/latest/format.html#nwbfile? I'd rather that than link to PyNWB specifically

I can try; they might have to be external links as opposed to inter-sphinx API, though.

CodyCBakerPhD:
OK so they can't be API links since it doesn't generate them through `automodule`, but they don't have to be pure external links either. Inter-sphinx can inherit all the internal section references, but this does have two differences:

(i) changes the appearance of the link to look more like a hyperlink as opposed to in-line code block
(ii) cannot seem to link to individual attributes, just the section definition for the object; *i.e.*, I can reference the nwb-schema for `TimeSeries`, but I can't link to the `TimeSeries.data` field directly, unlike
 ```
:py:attr:`~pynwb.file.TimeSeries.data`
```

Current example is in the Data Orientation section: https://nwbinspector--115.org.readthedocs.build/en/115/best_practices/time_series.html#data-orientation

**Comment by CodyCBakerPhD** [comment]
@bendichter One 'pro' to using the schema currently, if the format display is fine by you, is those attributes seem to have richer descriptions on the schema side even if we can't link to them directly.

**Comment by bendichter** [comment]
@CodyCBakerPhD hmm, I see what you mean about the trade-offs. It's there's also an inherent trade-off between being an approach that is general and language-agnostic rather than an approach that is directed and language-specific. I could lean towards using references to the schema. Ideally, I think the schema documentation should include links to the PyNWB and MatNWB API docs

**Comment by CodyCBakerPhD** [comment]
@bendichter One issue with using the schema references is not every section seems to have a clean internal reference. The most common types like

:nwb-schema:ref:`TimeSeries <sec-timeseries>` -> https://nwb-schema.readthedocs.io/en/latest/format.html#sec-timeseries
:nwb-schema:ref:`ElectricalSeries<sec-electricalseries>` -> https://nwb-schema.readthedocs.io/en/latest/format.html#sec-electricalseries

have these, but some others don't, such as

:nwb-schema:ref:`SpatialSeries<sec-spatialseries>` -/> https://nwb-schema.readthedocs.io/en/latest/format.html#sec-spatialseries
:nwb-schema:ref:`ImageSeries<sec-imageseries>` -/> https://nwb-schema.readthedocs.io/en/latest/format.html#sec-imageseries

**I could still use links to the NWB Schema for these, but they would be direct externally defined links to the webpage section**, if that's OK with you.

One 'pro' that would provide, is each and every schema reference would be then be checked by the CI (which does NOT currently check inter-sphinx or API references).

**Comment by bendichter** [comment]
I'm surprised that not every section has a label. That seems like a bug in the nwb schema docs

The docs are generated with [this script](https://github.com/hdmf-dev/hdmf-docutils/blob/main/hdmf_docutils/generate_format_docs.py).

It looks like the labels are generated [here](https://github.com/hdmf-dev/hdmf-docutils/blob/96d381e7e84f7c5c526c2e6d69ae110d8f2c52d6/hdmf_docutils/generate_format_docs.py#L165), which calls [this function](https://github.com/hdmf-dev/hdmf-docutils/blob/96d381e7e84f7c5c526c2e6d69ae110d8f2c52d6/hdmf_docutils/doctools/rst.py#L61-L67). I'm not sure why certain data types would be skipped. Is it possible the string is different from what you are expecting?

**Comment by CodyCBakerPhD** [comment]
>I'm surprised that not every section has a label. That seems like a bug in the nwb schema docs

>
>The docs are generated with [this script](https://github.com/hdmf-dev/hdmf-docutils/blob/main/hdmf_docutils/generate_format_docs.py).

>
>It looks like the labels are generated [here](https://github.com/hdmf-dev/hdmf-docutils/blob/96d381e7e84f7c5c526c2e6d69ae110d8f2c52d6/hdmf_docutils/generate_format_docs.py#L165), which calls [this function](https://github.com/hdmf-dev/hdmf-docutils/blob/96d381e7e84f7c5c526c2e6d69ae110d8f2c52d6/hdmf_docutils/doctools/rst.py#L61-L67). I'm not sure why certain data types would be skipped. Is it possible the string is different from what you are expecting?


I really don't know, then - looking at the code the labels ought to exist and ought to have pretty identical naming structures.

I guess it might come down the name that it draws from in the PyNWB namespaces that it seems to draw those sections and their `section.titles` from, but I'm unable to figure out how to load that into code for introspection.

I've exhausted a list of manual possibilities here: https://nwbinspector--115.org.readthedocs.build/en/115/best_practices/time_series.html#time-series

If you can figure out how to link to that one, I can probably figure out the pattern for any others.

**Comment by CodyCBakerPhD** [comment]
Turns out it was just a syntax error, my bad.

Updated all the rest to use the NWB Schema instead of PyNWB.

File: docs/best_practices/time_series.rst [hunk 7]
```diff
@@ -1,92 +1,152 @@
...
+Compress Data
+~~~~~~~~~~~~~
+
+Data writers can optimize the storage of large data arrays for particular uses by using compression applied to each
+chunk individually.
+
+This is especially important when writing NWBFiles that are intended to be uploaded to the
+:dandi-archive:`DANDI Archive <>` for storage, sharing, and publication.
+
+For more information about how to enable compression on your data, consult the
```
bendichter:
```suggestion
Data writers can optimize the storage of large data arrays for particular uses by using compression applied to each
chunk individually. This is especially important when writing NWBFiles that are intended to be uploaded to the
:dandi-archive:`DANDI Archive <>` for storage, sharing, and publication. For more information about how to enable compression on your data, consult the
```

PR #118, opened 2022-03-22
Declare & bundle license

jwodder:
This PR declares the project's license type in the metadata and also causes `license.txt` to be included in both sdists and wheels.  Such practice is strongly recommended so that there are no doubts by users about the project's licensing status.

**Note:** You may also want to change the name of the project at the start of `license.txt` from pynwb to nwbinspector, depending on whether you consider this project to be part of pynwb or not.  Also, unless NeurodataWithoutBorders is actually run by Berkeley, the license has fallen victim to blind copy & paste and needs to be corrected to list the actual copyright holders.

## Checklist

### DIFF
--- a/nwbinspector/checks/nwb_containers.py
+++ b/nwbinspector/checks/nwb_containers.py
@@ -41,9 +41,9 @@ def check_small_dataset_compression(
     """
     for field in getattr(nwb_container, "fields", dict()).values():
         if (
-                isinstance(field, h5py.Dataset)
-                and field.compression is None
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
@jwodder Thanks for pointing this out; looks like it was definitely just basic copy and paste during the early days.

@rly @oruebel Aside from changing the name from 'pynwb' to 'nwbinspector', are there other details that need to be changed or updated with this license?

bendichter (approved):

PR #121, opened 2022-03-22
[Feature]: PyNWB validation optional and update config for DANDI

CodyCBakerPhD:
## Motivation

A couple requests from the DANDI integration effort.

1. Make PyNWB validation optional in both CLI and library functions - this is in general useful when running on old files, but also links to https://github.com/NeurodataWithoutBorders/pynwb/pull/1432 and the surrounding issues.
2. Add small helper for loading internal configs; this was previously only a part of the CLI, this just makes the code importable from the library.
3. Small bug fix to config schema to get rid of deprecation warning that turns into an error on the DANDI side

To-Do really quick:

- [x] add test for `load_config` with `dandi` keyword at a minimum.

It's hard to add tests for the validation bypassing, unless you know of an easy way to generate an invalid NWBFile? Otherwise I just tested it locally on dandiset `000004`, which has always given` PYNWB_VALIDATION` levels.


nwbinspector/nwbinspector.py
```diff
@@ -66,6 +66,20 @@ def copy_function(function):
     return copied_function
 
 
+def load_config(filepath_or_keyword: PathType) -> dict:
```
bendichter, line 69: you are requiring filepath_or_keyword to be PathType?

bendichter: I don't see what this refactor has to do with skipping validation

CodyCBakerPhD: `PathType=Union[Path, str]`, which is technically what this argument takes, though the type name may be a bit misleading in that respect, since the `str` part could be either a string path or a string keyword.

From PR description

> Add small helper for loading internal configs; this was previously only a part of the CLI, this just makes the code importable from the library.

It's easier to setup both these changes so the DANDI PR can point to a single branch for the fixes

nwbinspector/nwbinspector.py
```diff
@@ -201,11 +214,12 @@ def inspect_all_cli(
 def inspect_all(
     path: PathType,
     modules: OptionalListOfStrings = None,
-    config: dict = None,
+    config: Optional[Union[dict, str]] = None,
```
bendichter: config can be a string here?

CodyCBakerPhD: Whoops, that was from an experimental approach that I abandoned.

bendichter (approved):

PR #123, opened 2022-03-23
[Add Check]: all ogen sites are referenced by ogen series

bendichter:


docs/best_practices/ogen.rst
```diff
@@ -0,0 +1,8 @@
+Optogenetics
+============
+
+OptogeneticSeries
+-----------------
+
+Each ``OptogeneticStimulusSite`` object present in an ``NWBFile`` should
+be referenced by at least on ``OptogeneticSeries`` in the same file.
```
CodyCBakerPhD, line 10:
```suggestion
be referenced by at least on ``OptogeneticSeries`` in the same file.

Check function: :py:meth:`~nwbinspector.checks.ogen.check_optogenetic_stimulus_site_has_optogenetic_series`
```

docs/best_practices/ogen.rst
```diff
@@ -0,0 +1,8 @@
+Optogenetics
+============
+
+OptogeneticSeries
```
CodyCBakerPhD, line 6:
```suggestion
.. _best_practice_optogenetic_stimulus_site_has_optogenetic_series:

OptogeneticSeries
```

CodyCBakerPhD (approved):

CodyCBakerPhD:
@bendichter Thanks for adding a BP section along with the check, I setup the inter-refs for it.

Check looks good. Very interesting logic, I think this is our first one that is a kind of 'second order' test in that it involves a second iteration over the NWBFile objects to find any objects related to the one of primary interest. Mentally noting that this is another place where the LRU caching could be beneficial, since the `neurodata_types` of the `nwbfile.objects` could have already been calculated in the core upstream functions.

PR #133, opened 2022-03-26
[Add Check]: Single-row tables

CodyCBakerPhD:
Uncommented check, tests, and added BP section for checking tables with only a single row.

### DIFF
--- a/docs/best_practices/tables.rst
+++ b/docs/best_practices/tables.rst
@@ -5,6 +5,18 @@ The DynamicTable data type that NWB uses allows you to define custom columns, wh
 
 
 
+.. _best_practice_single_row:
+
+Tables With Only a Single Row
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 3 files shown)


bendichter (approved):

docs/best_practices/tables.rst
```diff
@@ -5,6 +5,18 @@ The DynamicTable data type that NWB uses allows you to define custom columns, wh
 
 
 
+.. _best_practice_single_row:
+
+Tables With Only a Single Row
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is not common to save a table with only a single row entry. Consider if other ``neurodata_types`` might be better
+suited, such as a one-dimensional :nwb-schema:ref:`sec-TimeSeries` or any of its subtypes.
```
```suggestion
It is not common to save a table with only a single row entry. Consider other ``neurodata_types``, such as a one-dimensional :nwb-schema:ref:`sec-TimeSeries` or any of its subtypes.
```

PR #134, opened 2022-03-26
[Add Check]: table column data is not None

CodyCBakerPhD:
Uncommented check, added tests and BP section for when the data in a column is None.

One case that isn't currently being checked (and would actually fail to detect) is when a user adds data that is an `Iterable` (`list` or `array`, etc.) that itself contains some values set to `None`.

Is there a general way we could flatten any arbitrary nesting of such objects in a data field to examine the first `nelems` raw contents? I.e., such as when `data = [[[1, 2, None]]]`.

EDIT: switching to draft based on similarity to #135 and possibility of merging the two into a general check

### DIFF
--- a/docs/best_practices/tables.rst
+++ b/docs/best_practices/tables.rst
@@ -5,6 +5,19 @@ The DynamicTable data type that NWB uses allows you to define custom columns, wh
 
 
 
+.. _best_practice_column_data_is_not_none:
+
+Do Not Store Data as ``None``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specific to the :pynwb-docs:`PyNWB API <>`, when writing data to rows or columns of a :nwb-schema:`Dynamictable`, do
+not use the native Python type of ``None``, as it is not a :nwb-schema:ref:`supported value <common-attributes>` in the
+NWB schema.
+
+Check function :py:meth:`~nwbinspector.checks.table.check_column_data_is_not_none`
+
+
+
 .. _best_practice_dynamic_table_region_data_validity:
 
 Unassigned
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -32,6 +32,14 @@ def check_empty_table(table: DynamicTable):
         return InspectorMessage(message="This table has no data added to it.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=DynamicTable)
+def check_column_data_is_not_none(table: DynamicTable, nelems: int = 200):
+    """Check column values in a DynamicTable to ensure they are not None."""
+    for column in table.columns:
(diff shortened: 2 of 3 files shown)


docs/best_practices/tables.rst
```diff
@@ -5,6 +5,19 @@ The DynamicTable data type that NWB uses allows you to define custom columns, wh
...
 
 
+.. _best_practice_column_data_is_not_none:
+
+Do Not Store Data as ``None``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specific to the :pynwb-docs:`PyNWB API <>`, when writing data to rows or columns of a :nwb-schema:`Dynamictable`, do
+not use the native Python type of ``None``, as it is not a :nwb-schema:ref:`supported value <common-attributes>` in the
+NWB schema.
```
bendichter, line 15: We should include guidance for how to handle missing metadata for different data types

CodyCBakerPhD:
Actually, similar to what I'm finding in PR #135, it doesn't seem possible to actually complete an `io.write()` with any `None` values anywhere in the table. Do we even need this kind of check, then?

nwbinspector/checks/tables.py
```diff
@@ -32,6 +32,14 @@ def check_empty_table(table: DynamicTable):
         return InspectorMessage(message="This table has no data added to it.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=DynamicTable)
+def check_column_data_is_not_none(table: DynamicTable, nelems: int = 200):
+    """Check column values in a DynamicTable to ensure they are not None."""
+    for column in table.columns:
+        if any((x is None for x in column.data[:nelems])):
+            yield InspectorMessage(message=f"The column '{column.name}' has 'data' set to None.")
```
bendichter, line 40: How would this even happen? There is no `None` type in HDF5.

CodyCBakerPhD: See above: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/134#issuecomment-1080824595

> Actually, similar to what I'm finding in PR https://github.com/NeurodataWithoutBorders/nwbinspector/pull/135, it doesn't seem possible to actually complete an io.write() with any None values anywhere in the table. Do we even need this kind of check, then?

bendichter: it's more of an issue of how _to_ handle missing data in a VectorData object

CodyCBakerPhD: In that case, perhaps a separate check "check_missing_column_data" with logic "if len(column.data) == 0:"?

Note we already have one higher level case of this for entirely empty tables (no rows/cols): https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/checks/tables.py#L28-L32

PR #135, opened 2022-03-26
[Add Check]: TimeSeries data is not None

CodyCBakerPhD:
Shares a basic logical approach to #134, so indicates it could be possible to merge the functions into a general check somehow. Hence opening in draft until we can discuss.

Uncommented code, added tests, added BP section.

### DIFF
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -145,3 +145,20 @@ chunk individually. This is especially important when writing NWBFiles that are 
 
 Check functions: :py::meth:`~nwbinspector.checks.nwb_containers.check_large_dataset_compression`,
 :py::meth:`~nwbinspector.checks.nwb_containers.check_small_dataset_compression`
+
+
+
+.. _best_practice_time_series_data_is_not_none:
+
+Do Not Store Data as ``None``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specific to the :pynwb-docs:`PyNWB API <>`, when writing data to a :nwb-schema:ref:`sec-TimeSeries` object, or any
+descendant, do not use the native Python type of ``None``, as it is not a
+:nwb-schema:ref:`supported value <common-attributes>` in the NWB schema.
+
+There is one exception to this for the :nwb-schema:ref:`imageseries` type: when reading and writing such objects in the
+:pynwb-docs:`PyNWB API <>`, if the ``external_file`` field was specified instead, the ``data`` field will be parsed as
+``None``.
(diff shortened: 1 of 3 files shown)


docs/best_practices/time_series.rst
```diff
@@ -145,3 +145,20 @@ chunk individually. This is especially important when writing NWBFiles that are
...
+
+
+.. _best_practice_time_series_data_is_not_none:
+
+Do Not Store Data as ``None``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Specific to the :pynwb-docs:`PyNWB API <>`, when writing data to a :nwb-schema:ref:`sec-TimeSeries` object, or any
+descendant, do not use the native Python type of ``None``, as it is not a
+:nwb-schema:ref:`supported value <common-attributes>` in the NWB schema.
```
bendichter, line 158: best practices should not be language-specific. How would this sentence apply to someone using MATLAB?

docs/best_practices/time_series.rst
```diff
@@ -145,3 +145,20 @@ chunk individually. This is especially important when writing NWBFiles that are
...
+
+There is one exception to this for the :nwb-schema:ref:`imageseries` type: when reading and writing such objects in the
+:pynwb-docs:`PyNWB API <>`, if the ``external_file`` field was specified instead, the ``data`` field will be parsed as
+``None``.
```
bendichter, line 162: This isn't true generally...

docs/best_practices/time_series.rst
CodyCBakerPhD: Yes, this section is pretty specific to Python, since I have no idea how MATLAB handles situations like this; all I know is if the `data` is missing, and you read it using `PyNWB` (as the Inspector does), it shows up in the end as `None`.

I guess we could just refer to it as being 'missing' in the docs?

docs/best_practices/time_series.rst
CodyCBakerPhD: Ah, true; I was just reading the docstring of `ImageSeries`, but it seems it still handles missing input by turning it into an empty array.

CodyCBakerPhD:
This check is one of the ones from the old days: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/checks/time_series.py#L74-L87

But nowadays I don't seem to be able to actually create any kind of `TimeSeries` object with the `data` field actually `None`, or even containing `None` (with can be done on object creation just not `io.write`).

So is this check even needed any more?

PR #136, opened 2022-03-26
[Add Check]: TimeSeries unknown resolution

CodyCBakerPhD:
Uncommented check, added tests, added BP section.

### DIFF
--- a/nwbinspector/checks/time_series.py
+++ b/nwbinspector/checks/time_series.py
@@ -71,7 +71,7 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
 def check_resolution(time_series: TimeSeries):
     """Check the resolution value of a TimeSeries for proper format (-1.0 or NaN for unknown)."""
-    if not (np.isnan(time_series.resolution) or time_series.resolution == -1.0) and time_series.resolution <= 0:
+    if time_series.resolution != -1.0 and time_series.resolution <= 0:
         return InspectorMessage(
             message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
         )
--- a/nwbinspector/checks/time_series.py
+++ b/nwbinspector/checks/time_series.py
@@ -68,6 +68,15 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
         return InspectorMessage(f"{time_series.name} timestamps are not ascending.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_resolution(time_series: TimeSeries):
+    """Check the resolution value of a TimeSeries for proper format (-1.0 or NaN for unknown)."""
+    if not (np.isnan(time_series.resolution) or time_series.resolution == -1.0) and time_series.resolution <= 0:
+        return InspectorMessage(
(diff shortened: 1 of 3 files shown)


nwbinspector/checks/time_series.py
```diff
@@ -68,6 +68,15 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
...
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_resolution(time_series: TimeSeries):
+    """Check the resolution value of a TimeSeries for proper format (-1.0 or NaN for unknown)."""
+    if not (np.isnan(time_series.resolution) or time_series.resolution == -1.0) and time_series.resolution <= 0:
+        return InspectorMessage(
+            message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
+        )
+
```
bendichter: So this would catch if e.g. resolution = -2? Does that ever happen? Also, this logic is a little wonky. If `time_series.resolution <= 0` than it can't be NaN.

CodyCBakerPhD: > So this would catch if e.g. resolution = -2? Does that ever happen?

It would catch any negative `resolution` that is not specifically `-1.0` (which is also the PyNWB default), which I think is mostly there since `NaN` isn't possible in MatNWB? Otherwise I'd recommend `NaN` be the only allowed value to indicate 'unknown', and we disallow any negative resolution.

I don't know how often, or if it all it 'has happened' in observed DANDISets, but it is certainly possible:

```python
from pynwb import TimeSeries
TimeSeries(name="test", unit="test", rate=1.0, data=[], resolution=-5.)
```


> Also, this logic is a little wonky. If time_series.resolution <= 0 than it can't be NaN.

Logic taken directly from original code: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/136/files#diff-212c892af23049b5555f7484c72d41d3c7960d23aa1732d4bb553f1d9acd0cd0L89-L95

I can try to improve though

nwbinspector/checks/time_series.py
```diff
@@ -68,6 +68,15 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
         return InspectorMessage(f"{time_series.name} timestamps are not ascending.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_resolution(time_series: TimeSeries):
+    """Check the resolution value of a TimeSeries for proper format (-1.0 or NaN for unknown)."""
+    if not (np.isnan(time_series.resolution) or time_series.resolution == -1.0) and time_series.resolution <= 0:
```
CodyCBakerPhD:
```suggestion
    if np.isnan(time_series.resolution) or time_series.resolution == -1.0:
        return
    if time_series.resolution <= 0:
```

@bendichter Do you think this is less wonky?

nwbinspector/checks/time_series.py
bendichter:
```suggestion
def check_resolution(time_series: TimeSeries):
    """Check the resolution value of a TimeSeries for proper format (-1.0 or NaN for unknown)."""
    if time_series.resolution != -1.0 and time_series.resolution <= 0:
        return InspectorMessage(
            message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
        )

```

bendichter (approved):

PR #137, opened 2022-03-26
[Add Check]: missing 'unit' in TimeSeries

CodyCBakerPhD:
Uncommented check, added tests, linked to existing BP section.


bendichter:
Honestly this one is tricky because yes `unit` is a required field but is is often the case that the `TimeSeries` is dealing with unitless data, e.g. `DFOverF` or the unit that you'd want to use is not practical, e.g. I've never seen an`RoiResponseSeries` in actual lumens.

CodyCBakerPhD:
> Honestly this one is tricky because yes `unit` is a required field but is is often the case that the `TimeSeries` is dealing with unitless data, e.g. `DFOverF` or the unit that you'd want to use is not practical, e.g. I've never seen an`RoiResponseSeries` in actual lumens.

Makes sense, but we should have a Best Practice default recommendation for what to set `unit` to in those cases; such as `unit="a.u."`, which I'd say is the most common version I see in publications. This would communicate that they thought about it instead of ignoring it by just setting `unit=""`.

bendichter (changes requested):
see thread

CodyCBakerPhD:
> see thread

Sorry, which thread?

bendichter (approved):

PR #138, opened 2022-03-27
[Add Check]: attempt to check for 'default' start times

### LINKED ISSUE #20: [Add Check]: check session start time
Should probably check the session start time has been set properly within a file; not just as a `datetime` object, but as a valid date range for performing neurophysiology studies. E.g., if `session_start_time <= datetime(1970, 1, 1)` print `"Session start time may not be set to the true date of the recording."`

### LINKED ISSUE #84: [Add Check]: unknown or default start times
### What type of check would you like to see added to the NWBInspector?

At times in the past, if a `starting_time` for some `TimeSeries` was unknown we set it to `np.nan` to indicate this fact. I just wanted to raise the question here officially to see if this is indeed encouraged or if the default behavior makes sense.

It's not stated in the best practices, and default behavior seems to assume that it's always `0.0`, but especially in cases where multiple data streams are incoming to the NWBFile this may not be correct due to synchronizing offsets.

_No response_

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #20 

Added two similar checks, corresponding tests, and added an explicit blurb about the practice in the Best Practice section.


nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -25,6 +26,22 @@ def check_processing_module_name(processing_module: ProcessingModule):
         )
 
 
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBFile)
+def check_session_start_time(nwbfile: NWBFile):
+    """Check if the session_start_time was set to an appropriate value."""
+    if nwbfile.session_start_time <= datetime(1980, 1, 1).astimezone():
```
CodyCBakerPhD, line 22: Is Jan. 1, 1980 an OK threshold for determining a potentially invalid neurophysiology experiment? Or should we go back further?

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -11,18 +11,26 @@ File Organization
...
+``session_start_time``, but when writing multiple NWBFiles that are all designed to align to the same time reference,
+the ``session_start_time`` may be set separately from the explicitly specified common ``timestamps_reference_time``
+used across all of the NWBFiles.
 
 All time-related data in the NWBFile should be synchronized to the ``timestamps_reference_time`` so that future users
 are able to understand the timing of all events contained within the NWBFile.
 
+Given the importance of this field within an :nwb-schema:ref:`sec-NWBFile`, is it critical that it be set to a proper
+value. Default values should never be used for this field. If the true date and time are unknown, use your best guess
+or the date and time of writing the :nwb-schema:ref:`sec-NWBFile`.
```
bendichter: Careful with the word "never". There may be cases where this needs to be omitted for HIPAA reasons.

nwbinspector/checks/nwbfile_metadata.py
bendichter: I think that's fine.

nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -25,6 +26,22 @@ def check_processing_module_name(processing_module: ProcessingModule):
...
+        return InspectorMessage(message="The session_start_time may not be set to the true date of the recording.")
```
bendichter:
```suggestion
        return InspectorMessage(message=f"The session_start_time, {nwbfile.session_start_time}, may not be set to the true date of the recording.")
```

nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -25,6 +26,22 @@ def check_processing_module_name(processing_module: ProcessingModule):
...
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBFile)
+def check_timestamps_reference_time(nwbfile: NWBFile):
+    """Check if the timestamps_reference_time was set to an appropriate value."""
+    if nwbfile.timestamps_reference_time <= datetime(1980, 1, 1).astimezone():
+        return InspectorMessage(
+            message="The timestamps_reference_time may not be set to the true date of the recording."
+        )
```
bendichter: If this field is used, it's fairly common to use Jan 1 1970. Is that a problem?

CodyCBakerPhD: Hmmm I suppose for `timestamps_reference_time` we can let it slide since, as long as all the time values in the value are indeed aligned to whatever it happens to be, it could be anything they want.

docs/best_practices/nwbfile_metadata.rst
CodyCBakerPhD: Ah, true

bendichter:
We can also check for future dates

nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -29,7 +29,10 @@ def check_processing_module_name(processing_module: ProcessingModule):
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBFile)
 def check_session_start_time(nwbfile: NWBFile):
     """Check if the session_start_time was set to an appropriate value."""
-    if nwbfile.session_start_time <= datetime(1980, 1, 1).astimezone():
+    if (
+        nwbfile.session_start_time <= datetime(1980, 1, 1).astimezone()
+        or nwbfile.session_start_time >= datetime.now().astimezone()
+    ):
```
bendichter: I think these different violations should have different messages and different importance. Maybe they should be different checks entirely so we can control the importance independently.

CodyCBakerPhD: Proposing the future one to be `CRITICAL`, since it's an impossibility. Unless you think it's just a `VIOLATION`?

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -11,18 +11,26 @@ File Organization
...
+``session_start_time``, but when writing multiple NWBFiles that are all designed to align to the same time reference,
+the ``session_start_time`` may be set separately from the explicitly specified common ``timestamps_reference_time``
+used across all of the NWBFiles.
 
 All time-related data in the NWBFile should be synchronized to the ``timestamps_reference_time`` so that future users
 are able to understand the timing of all events contained within the NWBFile.
 
+Given the importance of this field within an :nwb-schema:ref:`sec-NWBFile`, is it critical that it be set to a proper
+value. Default values should generally not be used for this field. If the true date and time are unknown, use your
+best guess or the date and time of writing the :nwb-schema:ref:`sec-NWBFile`.
```
bendichter: Realistically, it is common for labs to not know the time of an experiment. Maybe we can speak to this.

CodyCBakerPhD: Changed to

"If the true date is unknown, use your best guess. If the exact start time is unknown, then it is fine to simply set it to midnight on that date."

oruebel: This sounds like we need an accuracy/error value for the start time.

bendichter: We could I suppose, but it's not as if we are often faced with sessions where we know a time plus-or-minus 2 hours. It's generally the case that we either know the time to the second or to the day. I think it would be a lot more straightforward to just allow for the session_start_time to be a date and not a datetime

CodyCBakerPhD: @oruebel I would second what @bendichter proposes. I forget the docval language, but in annotation typing it would make sense for

```python
NWBFile(
    session_start_time: Union[datetime.date, datetime.datetime],
    ...
)
```

Many of the Buzsaki DANDISets have this issue actually; either one of their acquisition systems grabbed the exact start time down to the second, or it didn't but they still recorded the date of the experiment in the file name.

bendichter: Of course best practice would be datetime.

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -11,18 +11,26 @@ File Organization
...
+Global Date and Time Reference
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-An :nwb-schema:ref:`sec-NWBFile` can have two primary time references. The global time reference for all objects in the
-:nwb-schema:ref:`sec-NWBFile` is the ``timestamps_reference_time``. By default, this is also set to the
-``session_start_time``, but when writing multiple NWBFiles that are all designed to align
-to the same time reference, the ``session_start_time`` may be set separately from the explicitly specified common
-``timestamps_reference_time`` used across all of the NWBFiles.
+An :nwb-schema:ref:`sec-NWBFile` can have two primary time references. The global date and time reference for all
+objects in the :nwb-schema:ref:`sec-NWBFile` is the ``timestamps_reference_time``. By default, this is also set to the
```
bendichter:
```suggestion
objects in the :nwb-schema:ref:`sec-NWBFile` is the ``timestamps_reference_time``. By default, this is set to the
```

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -11,18 +11,26 @@ File Organization
...
+``session_start_time``, but when writing multiple NWBFiles that are all designed to align to the same time reference,
+the ``session_start_time`` may be set separately from the explicitly specified common ``timestamps_reference_time``
+used across all of the NWBFiles.
```
bendichter:
```suggestion
the ``timestamp_reference_time`` used across all of the NWBFiles may be set separately from the ``session_start_time``.
```

bendichter (approved):

PR #143, opened 2022-03-30
Suppress cached namespace warnings

CodyCBakerPhD:
Currently, if you run the inspector on a directory of older files that didn't cache namespaces, you get a massive flood of warnings about that.

Example:

image

This not only clogs up the console for the final report display, but also seems to slow down the runtime by having the flush the `stdout` repeatedly.

This PR suppresses this simple warning (using `warnings.simplefilter`) but allows any other stronger warnings to continue through (as opposed to using `warnings.filterwarnings("ignore")`).

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -10,7 +10,7 @@ from enum import Enum
 from typing import Optional, List
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from types import FunctionType
-from warning import filterwarnings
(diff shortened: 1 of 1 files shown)


bendichter:
Is there a way to make this warning suppression more specific?

CodyCBakerPhD:
> Is there a way to make this warning suppression more specific?

@bendichter Made it hyper-specific to that warning by using `filterwarnings` combined with a regular expression for that message.

bendichter (approved):

PR #144, opened 2022-03-30
[Bug]: session_start_time not readable

CodyCBakerPhD:
A very, very strange bug that I detected while testing the new generalized formatting on real NWBFiles.

Took a while to pin down but I added some tests that show the unintended behavior at the level of core functions (unit tests, and core functions with unit-like usage, do not exhibit the error).

What appears to happen is that the [new check for the `description` field of an object](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/checks/general.py#L14-L23), if run before the [new checks on the `session_start_time`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/checks/nwbfile_metadata.py#L19-L38), corrupts the `session_start_time` of the in-memory `nwbfile` object by setting it to `None`, or something of that nature. Still trying to reduce the issue down to absolutely minimally reproducing components through PyNWB only.

This issue is specific only to the most recent version of the NWBInspector as of a day ago when these checks were merged together; the issue was not discoverable independently.

This PR includes a temporary band-aid fix by allowing the default ordering to run `nwbfile_metadata` checks before the `general` ones, but I'll continue working on a more sophisticated fix.

### DIFF
--- a/tests/unit_tests/test_ophys.py
+++ b/tests/unit_tests/test_ophys.py
@@ -11,7 +11,7 @@ from nwbinspector.checks.ophys import (
     check_roi_response_series_dims,
     check_roi_response_series_link_to_plane_segmentation,
 )
-from nwbinspector.register_checks import InspectorMessage, Importance, Severity
+from nwbinspector.register_checks import InspectorMessage, Importance
 
 
 class TestCheckRoiResponseSeries(TestCase):
@@ -84,7 +84,6 @@ class TestCheckRoiResponseSeries(TestCase):
             check_function_name="check_roi_response_series_dims",
             object_type="RoiResponseSeries",
             object_name="RoiResponseSeries",
-            location="/processing/ophys/",
         )
 
     def test_check_wrong_dims(self):
@@ -110,7 +109,6 @@ class TestCheckRoiResponseSeries(TestCase):
             check_function_name="check_roi_response_series_dims",
             object_type="RoiResponseSeries",
             object_name="RoiResponseSeries",
-            location="/processing/ophys/",
         )
 
     def test_pass_check_roi_response_series_dims(self):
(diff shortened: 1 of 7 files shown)


CodyCBakerPhD:
We figured out the issue; it was a particular behavior of some older code in `parse_location`, which the decorator was using to pop the `session_start_time` from the in-memory `nwbfile` object when running the `nwbfile_metadata` checks.

To-Do:
- [x] Test a bit more
- [x] Check that setting `location=None` is handled in display

CodyCBakerPhD:
@bendichter OK, this is fixed now. It seems the previous try/except code was mostly prototyped to function on `nwbfile.objects` that existed _before writing_ the NWBFile, so that locations could be parsed from the `nwbfile` while it was still in-memory; as we've proceeded more towards applying the inspector post-write, it makes sense to just remove it and rely solely on `h5py.Dataset` parent links to recover location. The exception, of course, being special static-location objects that are not truly `h5py.Groups` (and do not contain any `h5py.Dataset` objects either) such as `Subject`/`Device`/etc.

tests/unit_tests/test_ophys.py
```diff
@@ -84,7 +84,6 @@ def test_check_flipped_dims(self):
             check_function_name="check_roi_response_series_dims",
             object_type="RoiResponseSeries",
             object_name="RoiResponseSeries",
-            location="/processing/ophys/",
```
bendichter, line 87: wait, why is this being removed? Shouldn't this location be recoverable?

CodyCBakerPhD: In this particular test, the object being tested (and indeed the NWBFile it was added to) is still in-memory as `np.array`. It doesn't become a `h5py.Dataset` until we write the file and then read it back in, which is only currently being done in the `test_inspector.py` integration tests.

CodyCBakerPhD: Note that within those integration tests, `location` appears to be working correctly hence why I didn't have to delete those: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/fix_session_start_time_bug/tests/test_inspector.py#L159-L181

bendichter: ah ok that makes sense, thanks for clarifying

bendichter (approved):

PR #146, opened 2022-03-31
[Feature]: Adding header to JSON save (CLI)

CodyCBakerPhD:
Since it's independent of the primary text formatting in the other PR; this tool function grabs basic system/package info. This header is then included with the `--json-file-path` argument when dumping a raw message stream.

I also added, for consistency with the `--report-file-path` option, overwrite functionality (previously, would always overwrite) and a one-line console output at the bottom of the report displaying the successful save location (I find this useful when specifying a relative path as the target, _e.g._, `--json-file-path ../saving_one_level_up.json`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     url="https://github.com/NeurodataWithoutBorders/nwbinspector",
-    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema"],
+    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "importlib"],
     entry_points={"console_scripts": ["nwbinspector=nwbinspector.nwbinspector:inspect_all_cli"]},
     license="BSD-3-Clause",
 )
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     url="https://github.com/NeurodataWithoutBorders/nwbinspector",
-    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "importlib"],
+    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema"],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
     packages=find_packages(),
     include_package_data=True,
     url="https://github.com/NeurodataWithoutBorders/nwbinspector",
-    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema"],
+    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "importlib"],
```
bendichter: It looks to me like importlib is in the Python standard library so it doesn't need to be listed in requirements

CodyCBakerPhD:
Putting into draft until I work out the v3.7 issue

setup.py
CodyCBakerPhD: https://github.com/NeurodataWithoutBorders/nwbinspector/runs/5776230613?check_suite_focus=true

It appears as if the specific sub-module is not included in 3.7

CodyCBakerPhD: Yup, 'new to version 3.8': https://docs.python.org/3.10/library/importlib.metadata.html

bendichter: would this be a good use-case for `from __future__`, or perhaps should we make this value accessible via `nwbinspector.__version__`

CodyCBakerPhD: Personally I'm fine setting `nwbinspector.__version__`; but while reading up on the `importlib.metadata` and why it's better, I found old discussion regarding why https://peps.python.org/pep-0396/ was rejected

bendichter: huh, where's the discussion?

CodyCBakerPhD: Was mostly spread around multiple threads, but this one summarizes nicely: https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules

Also inspired currently testing solution through `pkg_resources`. Once our minimal supported version bumps to 3.8 we can remove the try/except

bendichter (approved):

PR #152, opened 2022-04-07
[Bug]: Fix display of similar files

CodyCBakerPhD:
Currently the formatted counts the file displayed in additional to 'other # of files'. This fixes it by counting the number of files minus the first one reported, and polishes the string a bit, too.

### DIFF
--- a/nwbinspector/inspector_tools.py
+++ b/nwbinspector/inspector_tools.py
@@ -217,7 +217,7 @@ class MessageFormatter:
                     message_header = self._get_message_header(message=message)
                     num_same = len(same_messages)
                     file_or_files = "s" if num_same > 2 else ""
-                    additional_file_str = f"and {num_same-1} other file{file_or_files}" if num_same > 1 else ""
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Example: it should be

```
2.15  sub-TWH103\sub-TWH103_ses-20181101_ecephys+image.nwb: check_column_binary_capability - 'Units' object with name 'units'
        Message: electrodes uses integers but has binary values [0 1]. Consider making it boolean instead and renaming the column to start with 'is_'; doing so will save 9.00B.
```

but it used to be

```
2.15  sub-TWH103\sub-TWH103_ses-20181101_ecephys+image.nwb and 1 other files: check_column_binary_capability - 'Units' object with name 'units'
        Message: electrodes uses integers but has binary values [0 1]. Consider making it boolean instead and renaming the column to start with 'is_'; doing so will save 9.00B.
```

CodyCBakerPhD:
- [x] Need to fix tests

bendichter (approved):

bendichter:
lgtm!

PR #162, opened 2022-04-08
[Modify Check]: allow single row check to skip special tables

CodyCBakerPhD:
As per discussion the other day, it is possible for Units tables to have only a single spiking Unit, so that's now allowed and will not trigger the check.

I also extended this to the ElectrodeTable, since it is likewise OK to only have a single electrode. Granted that, in practice, mostly applies to `icephys` where they should be using separate `IcephysElectrode` objects anyway, but I guess it's possible for an `ecephys` tetrode to have 3 bad or non-neural ones plugged in, too.

### DIFF
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -136,7 +136,9 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
 def check_single_row(
-    table: DynamicTable, exclude_types: Optional[list] = None, exclude_names: Optional[List[str]] = None
+    table: DynamicTable,
+    exclude_types: Optional[list] = (Units, ),
+    exclude_names: Optional[List[str]] = ("electrodes", ),
 ):
     """
     Check if DynamicTable has only a single row; may be better represented by another data type.
@@ -144,8 +146,6 @@ def check_single_row(
     Skips the Units table since it is OK to have only a single spiking unit.
     Skips the Electrode table since it is OK to have only a single electrode.
     """
-    exclude_types = [Units] if exclude_types is None else exclude_types
-    exclude_names = ["electrodes"] if exclude_names is None else exclude_names
     if any((isinstance(table, exclude_type) for exclude_type in exclude_types)):
         return
     if any((table.name == exclude_name for exclude_name in exclude_names)):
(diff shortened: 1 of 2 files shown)


nwbinspector/checks/tables.py
```diff
@@ -134,8 +135,15 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
 
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
-def check_single_row(table: DynamicTable):
-    """Check if DynamicTable has only a single row; may be better represented by another data type."""
+def check_single_row(table: DynamicTable, exclude_names: List[str] = ["Units", "electrodes"]):
```
CodyCBakerPhD: I wanted this to be `exclude_types = [pynwb.Units, pynwb.ElectrodeTable]` with `isinstance` logic checking if the `table` was a subtype of that exclusion. While this works perfectly fine for the `Units`, apparently the `ElectrodeTable` is not actually a subtype of a DynamicTable but rather a function that returns a DynamicTable of a particular structure.

bendichter: yes, there is no ElectrodeTable neurodata type. It's an oversight in the schema.

bendichter: can you create two lists, one for exclude neurodata types and one for exclude names?

CodyCBakerPhD: Done

nwbinspector/checks/tables.py
```diff
@@ -135,13 +135,19 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
...
     """
     Check if DynamicTable has only a single row; may be better represented by another data type.
 
     Skips the Units table since it is OK to have only a single spiking unit.
     Skips the Electrode table since it is OK to have only a single electrode.
     """
+    exclude_types = [Units] if exclude_types is None else exclude_types
+    exclude_names = ["electrodes"] if exclude_names is None else exclude_names
+    if any((isinstance(table, exclude_type) for exclude_type in exclude_types)):
+        return
```
```suggestion
def check_single_row(
    table: DynamicTable,
    exclude_types: Optional[list] = (Units, ),
    exclude_names: Optional[List[str]] = ("electrodes", ),
):
    """
    Check if DynamicTable has only a single row; may be better represented by another data type.

    Skips the Units table since it is OK to have only a single spiking unit.
    Skips the Electrode table since it is OK to have only a single electrode.
    """
    if any((isinstance(table, exclude_type) for exclude_type in exclude_types)):
        return
```

bendichter (approved):

PR #125, opened 2022-03-24
[Documentation]: Tables and Extensions BP sections

CodyCBakerPhD:
~~WIP~~

I think these might be the last two or 3 that need an initial wave of organizing and setup.

A final round of finishing touches should be done once all these broad-scale organization ones are merged.

PR #163, opened 2022-04-08
[Add Check]: string-valued table elements contain dictionaries

### LINKED ISSUE #156: [Add Check]: no data entry should be a dictionary

We've seen cases where a Python dictionary is dumped directly into a data field of a table field that was allowed to be a string. These should instead be added as separate custom rows/columns to the table.

While it would be possibly to detect this with a simple `try`/`except` around a `dict` cast of the `str` value, it might be better practice and more performant to use a simple regex like below. 

```shell
for data_element in data:
    re.findall(data_element, r"{.*?}")  # need to improve, obviously. Possibly "{".*?" : .*?}"? I'll test it a bit
```

Yes.

### Code of Conduct

CodyCBakerPhD:
As per discussion the other day, we've found cases where actual elements of a DynamicTable contained string casts of Python dictionary objects.

Technically this might be doable in MATLAB as well, but they would have to explicitly type out the string form to exactly match what Python would output. What I'm not sure of is if MATLAB has any other valid parallels to this, such as casting a struct as a string?

I'll add a Best Practice section highlighting the correct way to go about adding the dictionary information as additional rows and columns for this in a separate PR after #125.

### DIFF
--- a/tests/unit_tests/test_tables.py
+++ b/tests/unit_tests/test_tables.py
@@ -277,7 +277,7 @@ def test_check_table_values_for_dict():
     table = DynamicTable(name="test_table", description="")
     table.add_column(name="test_column", description="")
     table.add_row(test_column=str(dict(a=1)))
-    assert check_table_values_for_dict(table=table) == InspectorMessage(
+    assert check_table_values_for_dict(table=table)[0] == InspectorMessage(
         message=(
             "The column 'test_column' contains a string value that contains a dictionary! Please unpack "
             "dictionaries as additional rows or columns of the table."
--- a/tests/unit_tests/test_tables.py
+++ b/tests/unit_tests/test_tables.py
@@ -1,4 +1,5 @@
 import platform
+import json
 from unittest import TestCase
 
 import pytest
@@ -281,6 +282,24 @@ def test_check_table_values_for_dict():
         message=(
             "The column 'test_column' contains a string value that contains a dictionary! Please unpack "
             "dictionaries as additional rows or columns of the table."
+        ),
+        importance=Importance.BEST_PRACTICE_VIOLATION,
+        check_function_name="check_table_values_for_dict",
+        object_type="DynamicTable",
+        object_name="test_table",
+        location="/",
+    )
+
+
+def test_check_table_values_for_dict_json_case():
+    table = DynamicTable(name="test_table", description="")
+    table.add_column(name="test_column", description="")
+    table.add_row(test_column=json.dumps(dict(a=1)))
+    assert check_table_values_for_dict(table=table)[0] == InspectorMessage(
+        message=(
+            "The column 'test_column' contains a string value that contains a dictionary! Please unpack "
+            "dictionaries as additional rows or columns of the table. This string is also JSON loadable, so call "
+            "`json.loads(...)` on the string to unpack."
         ),
         importance=Importance.BEST_PRACTICE_VIOLATION,
         check_function_name="check_table_values_for_dict",
--- a/tests/unit_tests/test_tables.py
+++ b/tests/unit_tests/test_tables.py
@@ -13,6 +13,7 @@ from nwbinspector import (
     check_dynamic_table_region_data_validity,
     check_column_binary_capability,
     check_single_row,
+    check_table_values_for_dict,
 )
 from nwbinspector.register_checks import InspectorMessage, Importance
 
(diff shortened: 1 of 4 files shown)


bendichter:
I don't think this is specific enough and might trigger false positives. Why not use an actual regex for json?

bendichter:
I don't really think it makes sense to say that a string "is" a dictionary. If it's json, say it's json

CodyCBakerPhD:
> Why not use an actual regex for json?

> I don't really think it makes sense to say that a string "is" a dictionary. If it's json, say it's json

Thing is, not every dictionary is equivalent to JSON (JavaScript Object Notation). We talked about how how common it is to perform an operation such as `json.dump` and `json.load`, but that can result in a different output from calling `str(some_dict)` or including `some_dict` in a format-string like `f"some text: {some_dict}, more text"`.

In particular, calling `str(some_dict)` is more general than `json.dump` for two reasons: (i) JSON only supports strings as keys - Python supports any hashable objects as keys (including tuples, frozensets, custom classes, etc.) and (ii) JSON values can only be of a valid JSON type (list, dictionary, string, integer, float, boolean, Null) - Python values can be anything that can be constructed in Python.

As such, all the JSON-specific regex I tried only catch those specific subcases where the given `str` version of the `dict` was in perfect synchrony with JSON. But the spirit of this check goes beyond that. See https://github.com/NeurodataWithoutBorders/nwbinspector/pull/163/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R69-R72 for a new test illustrating this.


> I don't think this is specific enough and might trigger false positives.

I've tweaked it a bit and added more tests: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/163/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R34-R41

Including one that currently fails: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/163/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R44-R46

I'll think of how the regex can be improved to include this case, which upon careful inspection is not an object that could have been the result of any string-casting operation.

If you can think of any other test cases to add to avoid accidental false positives, just mention them and I can add them to the tests. I'll also run the check on various DANDISets to see if it ever triggers.

Likewise, if you can think of a valid reason some table element would have something like this inside its string value, we can try to think about how to account for that. But I'm pretty sure most structures, even jagged ones, can be taken care of by the rich DynamicTable functionality and would more appropriately be included as extra rows/cols or as indexed rows/cols.

bendichter:
calling `str(some_dict)` is not a standard way of storing dictionary-structured data as text, and I do not recall ever seeing an NWB file that has used this technique.

bendichter:
The generality of python dictionaries over json is not really a fair comparison here because you would not be storing python dictionaries, you would be storing the string representations of those objects, which for complex objects often wouldn't  enable you to do a round trip

CodyCBakerPhD:
> calling str(some_dict) is not a standard way of storing dictionary-structured data as text, and I do not recall ever seeing an NWB file that has used this technique.

We've done it ourselves for storing spike sorting provenance for the Feldman Lab - granted, we knew it wasn't 100% ideal then (an extension would have been the better way to go) but at least we put it into the `description` field for the table instead of copying it over row values.

> The generality of python dictionaries over json is not really a fair comparison here because you would not be storing python dictionaries, you would be storing the string representations of those objects, which for complex objects often wouldn't enable you to do a round trip

It's true, you can call `some_str=str(some_dict)` but not `dict(some_str)` without custom parsing. Thus where `json.loads(some_str)` would be the ideal call; however, my point is you cannot always call `json.dumps(some_dict)` if it contains non-JSON objects (which can happen fairly often, actually, classic example being a `pathlib.Path` instead of the `str` of that `Path`).


For easy reference, this is some code for grabbing the specific item that caused interest in this kind of check.

```python
from pynwb import NWBHDF5IO

with NWBHDF5IO(path="https://dandiarchive.s3.amazonaws.com/blobs/0ae/41b/0ae41b54-d2c2-4b4b-b32f-bbb631d7bf19", mode="r", driver="ros3", load_namespaces=True) as io:
    nwbfile = io.read()
    dicts_as_strs = nwbfile.epochs.arena.data[:]
```

The result is indeed JSON loadable, but the output is equivalent to either `json.dump` or `str(original_dict)` so without the original code there's no way to tell how they got it in there.

bendichter:
Let's discuss this in our meeting today

tests/unit_tests/test_tables.py
```diff
@@ -258,6 +258,37 @@ def test_check_single_row_fail():
     )
 
 
+def check_table_values_for_dict_non_str():
```
bendichter:
```suggestion
def test_check_table_values_for_dict_non_str():
```

tests/unit_tests/test_tables.py
```diff
@@ -258,6 +258,37 @@ def test_check_single_row_fail():
...
+    table = DynamicTable(name="test_table", description="")
+    table.add_column(name="test_column", description="")
+    table.add_row(test_column=123)
+    assert check_table_values_for_dict(table=table) is None
+
+
+def check_table_values_for_dict_pass():
```
bendichter:
```suggestion
def test_check_table_values_for_dict_pass():
```

tests/unit_tests/test_tables.py
```diff
@@ -258,6 +258,37 @@ def test_check_single_row_fail():
...
+    table = DynamicTable(name="test_table", description="")
+    table.add_column(name="test_column", description="")
+    table.add_row(test_column="123")
+    assert check_table_values_for_dict(table=table) is None
+
+
+def check_table_values_for_dict():
```
bendichter:
```suggestion
def test_check_table_values_for_dict():
```

CodyCBakerPhD:
@bendichter Added JSON loading aspect to the check as well as a test for that sub-case.

bendichter (approved):

PR #166, opened 2022-04-11
[Feature]: Direct inspection on Dandisets

### LINKED ISSUE #154: [Feature]: helper tool for running directly on a DANDISet ID #

We've been running the inspector on DANDISets using the `ros3` mode for a while now, but a lot of it is still manually specified code for resolving the s3 paths. It's pretty easy to do automatically though so might as well port over the code so we can just call a nice simple function directly each time.

Only concern is if it might produce circular dependencies when the dandi-cli merges the integration with the inspector.

Yes.

### Code of Conduct

### LINKED ISSUE #165: [Feature]: run command line on dandi hub

I'd like to run in Terminal:

```bash
nwbinspector https://dandiarchive.org/dandiset/000231/draft
```

and have nwbinspector automatically set up the streaming to inspect the dandiset in the cloud

No.

### Code of Conduct

CodyCBakerPhD:
fixes #165 and #154 

- Adds `dandi` as a dependency to use their API tools for automatically resolving content urls given a particular `dandiset_id`.

- Adds a helper function in `tools` for retrieving S3 paths together with the 'file paths' as they are displayed on the DANDI web front end.

- Adds a separate core function for performing inspection on a dandiset rather than a local path; there are two reasons for this...

(i) the generator behavior with parallelization is a bit different than the other functions (it will trigger iteration over all messages on a per-file basis instead of running a single check at each step). This is different is due to the second reason, which is...

(ii) the S3 streaming appears to become I/O bound unlike when running on local disk, which makes the previous multithreading very inefficient. Thankfully, distributing over processes works very well with the caveat that the setup for the `ProcessPoolExecutor` has different requirements in how it needs to communicate between workers.

(iii) The parallelization is extremely important. Here are some performance details:

Local copy of dataset `000004`: 45s for 1 job, 14s with 4 jobs (multithreading)
S3 streaming of `000004` on DANDIHub: 542s for 1 job, 11s for -1 jobs (I *think* 48, but not entirely sure since Satra enabled autoscaling) using parallel processes

- Added a new flag to the CLI indicating whether to `--stream` from S3. I had first set this to `--driver ros3` but since `ros3` is the only driver we'd really support with that, I thought  it would make more sense (and be simpler to code) to just have a boolean flag that equates to it. The CLI `path` in this case can take either a copy/paste of the full dandiarchive link, or just the six-digit ID and optional `version_id`.

- Also added description of the `path` argument to the CLI `help` by including it in the docstring. It looks like

image

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -314,7 +314,7 @@ def inspect_all(
                     yield message
     else:
         for nwbfile_path in nwbfiles:
-            for message in inspect_nwb(nwbfile_path=nwbfile_path, checks=checks):
+            for message in inspect_nwb(nwbfile_path=nwbfile_path, checks=checks, driver=driver):
                 yield message
 
 
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from collections.abc import Iterable
 from enum import Enum
 from typing import Optional, List
-from concurrent.futures import ThreadPoolExecutor, as_completed, ProcessPoolExecutor
+from concurrent.futures import ProcessPoolExecutor, as_completed
 from types import FunctionType
 from warnings import filterwarnings, warn
 from distutils.util import strtobool
@@ -187,6 +187,7 @@ def configure_checks(
     ),
     is_flag=True,
 )
+@click.option("--progress-bar", help="Set this flag to False to disable display of the progress bar.")
 @click.option(
     "--stream",
     help=(
@@ -202,7 +203,6 @@ def configure_checks(
         "When 'path' is a six-digit DANDISet ID, this further specifies which version of " "the DANDISet to inspect."
     ),
 )
-@click.option("--progress-bar", help="Sett this flag to False to disable the display of the progress bar.")
 def inspect_all_cli(
     path: str,
     modules: Optional[str] = None,
@@ -219,9 +219,9 @@ def inspect_all_cli(
     n_jobs: int = 1,
     skip_validate: bool = False,
     detailed: bool = False,
+    progress_bar: Optional[str] = None,
     stream: bool = False,
     version_id: Optional[str] = None,
-    progress_bar: Optional[str] = None,
 ):
     """
     Run the NWBInspector via the command line.
@@ -360,38 +360,22 @@ def inspect_all(
     checks = configure_checks(config=config, ignore=ignore, select=select, importance_threshold=importance_threshold)
 
     if n_jobs != 1:
-        # max_workers for threading is a different concept to number of processes; from the documentation
-        # https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor
-        # we can multiply the specified number of jobs by 5
-        # if n_jobs != -1:
-        #     max_workers = n_jobs * 5
-        # else:
-        #     max_workers = None  # concurrents doesn't have a -1 flag like joblib; set to None to achieve this
         progress_bar_options.update(total=len(nwbfiles))
         futures = []
-        n_jobs = None if n_jobs == -1 else n_jobs
+        n_jobs = None if n_jobs == -1 else n_jobs  # concurrents uses None instead of -1 for 'auto' mode
         with ProcessPoolExecutor(max_workers=n_jobs) as executor:
             for nwbfile_path in nwbfiles:
-                futures.append(executor.submit(_run_checks, nwbfile_path=nwbfile_path, checks=checks))
+                futures.append(
+                    executor.submit(
+                        _pickle_inspect_nwb, nwbfile_path=nwbfile_path, checks=checks, skip_validate=skip_validate
+                    )
+                )
+            completed_futures = as_completed(futures)
             if progress_bar:
-                for future in tqdm(as_completed(futures), **progress_bar_options):
-                    for message in future.result():
-                        yield message
-            else:
-                for future in as_completed(futures):
-                    for message in future.result():
-                        yield message
-        # with ThreadPoolExecutor(max_workers=max_workers) as executor:
-        #     for nwbfile_path in nwbfiles:
-        #         futures.append(executor.submit(inspect_nwb, nwbfile_path=nwbfile_path, checks=checks))
-        #     if progress_bar:
-        #         for future in tqdm(as_completed(futures), **progress_bar_options):
-        #             for message in future.result():
-        #                 yield message
-        #     else:
-        #         for future in as_completed(futures):
-        #             for message in future.result():
-        #                 yield message
+                completed_futures = tqdm(completed_futures, **progress_bar_options)
+            for future in completed_futures:
+                for message in future.result():
+                    yield message
     else:
         if progress_bar:
             nwbfiles = tqdm(nwbfiles, **progress_bar_options)
@@ -400,7 +384,8 @@ def inspect_all(
                 yield message
 
 
-def _run_checks(nwbfile_path: str, checks: list = available_checks, skip_validate: bool = False):
+def _pickle_inspect_nwb(nwbfile_path: str, checks: list = available_checks, skip_validate: bool = False):
+    """Auxilliary function for inspect_all to run in parallel using the ProcessPoolExecutor."""
     return list(inspect_nwb(nwbfile_path=nwbfile_path, checks=checks, skip_validate=skip_validate))
 
 
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -149,15 +149,7 @@ def configure_checks(
 
 
 @click.command()
-@click.argument(
-    "path",
-    help=(
-        "Relative path to either a local NWBFile, a local folder containing multiple NWBFiles, a link to a dataset on "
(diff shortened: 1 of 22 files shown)


nwbinspector/nwbinspector.py
```diff
@@ -200,24 +217,60 @@ def inspect_all_cli(
...
     detailed: bool = False,
+    stream: bool = False,
+    version_id: Optional[str] = None,
 ):
-    """Primary CLI usage of the NWBInspector."""
+    """
+    Run the NWBInspector via the command line.
+
+    path :
+    Relative path to either a local NWBFile, a local folder containing multiple NWBFiles, a link to a dataset on
```
bendichter: I'm pretty sure an absolute path would work too

```suggestion
    Path to either a local NWBFile, a local folder containing multiple NWBFiles, a link to a dataset on
```

bendichter: the folder could contain just one nwb file

```suggestion
    Relative path to either a local NWBFile, a local folder containing NWBFiles, a link to a dataset on
```

nwbinspector/nwbinspector.py
```diff
@@ -200,24 +217,60 @@ def inspect_all_cli(
...
+    DANDI archive (i.e., https://dandiarchive.org/dandiset/dandiset_id/version_id), or a six-digit DANDISet ID.
```
bendichter: while all dandiset ids are currently 6 digits, they are not constrained to always use that form

```suggestion
    DANDI archive (i.e., https://dandiarchive.org/dandiset/{dandiset_id}/{version_id}), or a Dandiset ID.
```

nwbinspector/nwbinspector.py
```diff
@@ -200,24 +217,60 @@ def inspect_all_cli(
...
+            dandiset_id, version_id = url_path.split("/")[-2:]
+        else:
+            dandiset_id = path  # version_id will get set automatically to most recent version if unspecified
+        assert url_path or re.fullmatch(
+            pattern="^[0-9]{6}$", string=dandiset_id
+        ), "'--stream' flag was enabled, but 'path' is neither a full link to the DANDI archive nor a DANDISet ID."
+        if Path(path).is_dir():
+            warn(
+                f"The local DANDISet '{path}' exists, but the '--stream' flag was used. "
+                "NWBInspector will use S3 streaming from DANDI. To use local data, remove the '--driver' flag."
```
```suggestion
                "NWBInspector will use S3 streaming from DANDI. To use local data, remove the '--stream' flag."
```

CodyCBakerPhD: Oops, good catch

bendichter:
This looks great! Could you put together a simple test? Once that is up I think this is good to go!

CodyCBakerPhD:
> This looks great! Could you put together a simple test? Once that is up I think this is good to go!

Yeah, so for tests I was wondering how to go about that... I guess I'll just setup the CI to use `ros3` and maybe point it to one of our staged test files...

Issue is, that means that running tests locally would also require 'ros3' to be enabled on any device trying to run those tests, unless I can figure a lazy way of skipping such a test. I'll look into it

nwbinspector/nwbinspector.py
```diff
@@ -419,5 +470,93 @@ def run_checks(nwbfile: pynwb.NWBFile, checks: list):
...
+            for s3_path, dandi_path in s3_urls_to_dandi_paths.items():
+                futures.append(
+                    executor.submit(_run_s3_checks, s3_path=s3_path, checks=checks, skip_validate=skip_validate)
+                )
+            for future in as_completed(futures):
+                for message in future.result():
+                    message.file_path = dandi_path
+                    yield message
+    else:
+        for s3_path, dandi_path in s3_urls_to_dandi_paths.items():
```
bendichter: could we add a progress bar here?

nwbinspector/nwbinspector.py
```diff
@@ -419,5 +470,93 @@ def run_checks(nwbfile: pynwb.NWBFile, checks: list):
...
+        The server type to use in resolving S3 paths.
+        Mostly used by tests that use https://api-staging.dandiarchive.org/api
+        Defaults to the standard archive.
+    """
+    s3_urls_to_dandi_paths = get_s3_urls_and_dandi_paths(dandiset_id=dandiset_id, version_id=version_id, n_jobs=n_jobs)
+    checks = configure_checks(config=config, ignore=ignore, select=select, importance_threshold=importance_threshold)
+
+    n_jobs = n_jobs if n_jobs > 0 else None
+    if n_jobs != 1:
+        with ProcessPoolExecutor(max_workers=n_jobs) as executor:
```
bendichter: is it possible to have a progress bar with ProcessPoolExecutor?

CodyCBakerPhD: Yep, I'll just do what SpikeInterface does: https://github.com/SpikeInterface/spikeinterface/blob/ce2af6098ea4e9caf2410bf5f3d8a5708bc56841/spikeinterface/core/job_tools.py

bendichter:
this might be helpful:

https://github.com/dandi/dandi-hub/blob/226656d33f0429cd912dfd1e123bda50d241c98a/docker/Dockerfile#L39

CodyCBakerPhD:
@bendichter Voila, got the ros3 working in the CI and the streaming test is now passing.

Note I had to make a small change to the new `__repr__` for `InspectorMessages` because they were causing a syntax issue; they are now displayed as

image


The [standard instructions from PyNWB](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/streaming.html?highlight=ros3#streaming-from-an-s3-bucket) were all that was needed to get the ros3 setup, which is lovely.

For reference on what was wrong with the CI, though, if anyone looks at this again in the future - the issue was we were accidentally overriding the conda virtual environment with a secondary python setup so it was never finding the conda installation of `h5py` that comes with `ros3` configured.

Anyway I'll refactor this PR to become a part of `inspect_dandi` so we don't have to duplicate arguments and documentation all over the place since they're pretty minimally different now.

nwbinspector/tools.py
```diff
@@ -22,3 +26,39 @@ def get_nwbfile_path_from_internal_object(obj):
...
+    ), "The specified 'path' is not a proper DANDISet ID. It should be a six-digit numeric identifier."
+
+    s3_urls_to_dandi_paths = dict()
+    if n_jobs != 1:
+        with DandiAPIClient() as client:
+            dandiset = client.get_dandiset(dandiset_id=dandiset_id, version_id=version_id)
+            max_workers = n_jobs if n_jobs > 0 else None
+            with ProcessPoolExecutor(max_workers=max_workers) as executor:
+                futures = []
+                for asset in dandiset.get_assets():
```
bendichter, line 48:
```suggestion
                for asset in dandiset.get_assets():
                    if asset.path.split(".")[-1] == "nwb"
```

NWB datasets can contain non-NWB assets, so let's skip those

nwbinspector/tools.py
```diff
@@ -22,3 +26,39 @@ def get_nwbfile_path_from_internal_object(obj):
...
+                    futures.append(
+                        executor.submit(_get_content_url_and_path, asset=asset, follow_redirects=1, strip_query=True)
+                    )
+                for future in as_completed(futures):
+                    s3_urls_to_dandi_paths.update(future.result())
```
bendichter:
```suggestion
                for asset in dandiset.get_assets():
                    if asset.path.split(".")[-1] == "nwb"
                        futures.append(
                            executor.submit(_get_content_url_and_path, asset=asset, follow_redirects=1, strip_query=True)
                        )
                    for future in as_completed(futures):
                        s3_urls_to_dandi_paths.update(future.result())
```

nwbinspector/tools.py
```diff
@@ -22,3 +26,39 @@ def get_nwbfile_path_from_internal_object(obj):
...
+    else:
+        with DandiAPIClient() as client:
+            dandiset = client.get_dandiset(dandiset_id=dandiset_id, version_id=version_id)
+            for asset in dandiset.get_assets():
+                s3_urls_to_dandi_paths.update(_get_content_url_and_path(asset=asset))
```
bendichter:
```suggestion
            for asset in dandiset.get_assets():
                if asset.path.split(".")[-1] == "nwb"
                    s3_urls_to_dandi_paths.update(_get_content_url_and_path(asset=asset))
```

nwbinspector/tools.py
```diff
@@ -22,3 +26,39 @@ def get_nwbfile_path_from_internal_object(obj):
     if isinstance(obj, NWBFile):
         return obj.container_source
     return obj.get_ancestor("NWBFile").container_source
+
```
bendichter, line 29:
```suggestion
def is_nwb_dandiset(dandiset):
    return any(
        x['identifier'] == 'RRID:SCR_015242'
        for x in dandiset.get_raw_metadata()['assetsSummary'].get('dataStandard', {})
    )
```

nwbinspector/tools.py
bendichter: to account for non-nwb files in nwb dandisets

bendichter:
wow, this looks great!

bendichter (approved):

CodyCBakerPhD:
> wow, this looks great!

Thanks, really excited about this. Added a few more tests to the CI to improve coverage, soon as they pass I'll merge

Issue #87, opened 2022-03-03
[Add Check]: validating video files under external_file attribute. 


Saksham20:
### What would you like to see added to the NWBInspector?

Video files that are linked to nwbfiles under `ImageSeries.external_file` should be inspected for:
- Checking their existence on disk with the path specified. (warning importance high). This check would defer for paths for local disk location vs urls. 
- Whether video files are of the common formats: .avi, .mp4, .mov, .wmv, .flv. This can be a mild recommendation (low importance)
- if external_file is a local path, checking if its an absolute or relative to nwbfile location (low importance)

### Pseudocode or other logic of the check

```shell
from pathlib import Path
import requests

video_file_path = "documents/folder0/video0.mp4"
video_file_url = "https://dandiarchive.s3.amazonaws.com/blobs/a52/253/a5225398-ec1c-4325-8bbe-83dc73fa8c87"

def video_exist_check(video_path, nwbfile_path):
   video_path_type = video_pathtype_check_absolute(video_path)
   if  video_path_type == "relative":
      pt = Path(nwbfile_path)/Path(video_path)
      return pt.exists()
   elif video_path_type == "url":
      return requests.head(video_path).ok
   else:
      return Path(video_path).exists()

def video_format_check(video_path):
   if Path(video_path).suffix in [".mp4", ".avi", ".wmv", ".mov", ".flv"]:
      return True
   else:
      return False

def video_pathtype_check_absolute(video_path):
   if os.path.isabs(video_path):
      return "absolute"
   elif any([str(video_path).startswith(i) for i in ["ftp://", "http://", "https://"]]):
      return "url"
   else:
      return "relative"
```

### Do you have any interest in helping implement the check function?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this check was not already [requested](https://github.com/neurodatawithoutborders/nwbinspector/issues)?

CodyCBakerPhD:
For starters, I think we can extend a couple of these checks to any object that uses external links.

> Checking their existence on disk with the path specified. (warning importance high). This check would defer for paths for local disk location vs urls.

Yeah, that would be great for any object that can and does have external links specified. What are some other objects that can currently support external links?

> if external_file is a local path, checking if its an absolute or relative to nwbfile location (low importance)

[...]

PR #167, opened 2022-04-11
[Add Check]: external file existence for ImageSeries

CodyCBakerPhD:
First step to resolving #87 

If an ImageSeries uses `external_file` rather than `data`; this value can either be an external link or a local uri. If local, checks that the file referenced exists. A future check will also verify that it is a relative instead of absolute path. If a url, then verifies the url can be accessed.

### DIFF
--- a/nwbinspector/checks/image_series.py
+++ b/nwbinspector/checks/image_series.py
@@ -7,7 +7,7 @@ from ..register_checks import register_check, Importance, InspectorMessage
 
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=ImageSeries)
-def check_image_series_external_file(image_series: ImageSeries):
+def check_image_series_external_file_valid(image_series: ImageSeries):
     """Check if the external_file specified by an ImageSeries actually exists at the relative location."""
     if image_series.external_file is None:
         return
--- a/nwbinspector/checks/image_series.py
+++ b/nwbinspector/checks/image_series.py
@@ -16,5 +16,8 @@ def check_image_series_external_file_valid(image_series: ImageSeries):
     for file_path in image_series.external_file:
         if not Path(file_path).is_absolute() and not (nwbfile_path / file_path).exists():
             yield InspectorMessage(
-                message=f"The external file {file_path} does not exist. Please confirm the relative location to the NWBFile."
+                message=(
+                    f"The external file '{file_path}' does not exist. Please confirm the relative location to the"
+                    " NWBFile."
+                )
             )
--- a/nwbinspector/checks/image_series.py
+++ b/nwbinspector/checks/image_series.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from pynwb.image import ImageSeries
 
 from ..register_checks import register_check, Importance, InspectorMessage
+from ..tools import get_nwbfile_path_from_internal_object
 
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=ImageSeries)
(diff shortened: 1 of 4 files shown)


nwbinspector/checks/image_series.py
```diff
@@ -0,0 +1,18 @@
...
+from ..register_checks import register_check, Importance, InspectorMessage
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=ImageSeries)
+def check_image_series_external_file_valid(image_series: ImageSeries):
+    """Check if the external_file specified by an ImageSeries actually exists at the relative location."""
+    if image_series.external_file is None:
+        return
+    for file_path in image_series.external_file:
+        if not Path(file_path).exists():
```
bendichter: path should be relative wrt nwbfile location, not current working directory. Would this work if the NWB file was in nested directories?

CodyCBakerPhD: Oh, uhh, didn't think about that. Lemme work on that then

CodyCBakerPhD: OK added a tool function for determining NWBFile path on disk from only the input object.

CodyCBakerPhD: Also added tests for different file locations relative to the NWBFile location.

nwbinspector/checks/image_series.py
```diff
@@ -0,0 +1,20 @@
...
+@register_check(importance=Importance.CRITICAL, neurodata_type=ImageSeries)
+def check_image_series_external_file_valid(image_series: ImageSeries):
+    """Check if the external_file specified by an ImageSeries actually exists at the relative location."""
+    if image_series.external_file is None:
+        return
+    nwbfile_path = Path(get_nwbfile_path_from_internal_object(obj=image_series))
+    for file_path in image_series.external_file:
+        if not Path(file_path).is_absolute() and not (nwbfile_path / file_path).exists():
+            yield InspectorMessage(
+                message="The external file does not exist. Please confirm the relative location to the NWBFile."
```
bendichter:
```suggestion
                message=f"The external file {file_path} does not exist. Please confirm the relative location to the NWBFile."
```

nwbinspector/tools.py
```diff
@@ -15,3 +15,13 @@ def all_of_type(nwbfile: NWBFile, neurodata_type):
...
+
+
+def get_nwbfile_path_from_internal_object(obj):
+    """Determine the file path on disk for a NWBFile given only an internal object of that file."""
+    ancestor = obj.get_ancestor()
+    if ancestor is None:
+        return
+    while ancestor.get_ancestor() is not None:
+        ancestor = ancestor.get_ancestor()
+    return ancestor.container_source
```
bendichter: I think this function could be optimized. It doesn't currently work if given an `NWBFile`.

bendichter:
```suggestion
from pynwb.file import NWBFile

def get_nwbfile_path_from_internal_object(obj):
    """Determine the file path on disk for a NWBFile given only an internal object of that file."""
    if isinstance(obj, NWBFile):
        return obj.container_source
    else:
        return obj.get_ancestor("NWBFile").container_source
```

nwbinspector/tools.py
```diff
@@ -15,3 +15,11 @@ def all_of_type(nwbfile: NWBFile, neurodata_type):
...
         if isinstance(obj, neurodata_type):
             yield obj
+
+
+def get_nwbfile_path_from_internal_object(obj):
+    """Determine the file path on disk for a NWBFile given only an internal object of that file."""
+    if isinstance(obj, NWBFile):
+        return obj.container_source
+    else:
+        return obj.get_ancestor("NWBFile").container_source
```
bendichter:
```suggestion
    return obj.get_ancestor("NWBFile").container_source
```

bendichter (approved):

PR #168, opened 2022-04-11
[Add Check]: Spike times not in unobserved intervals

### LINKED ISSUE #5: [Add Check]: check for spike times during unobserved intervals

CodyCBakerPhD:
fix #5 

Pretty self-explanatory. Note that all logic is done with generators so it will scan each spike time separately and terminate the first time it finds the issue.

That said, our standard `200` elements to extract seems a tad low for typical units tables. Even a single neuron firing at 10 Hz would produce 200 spikes after only 20 seconds, and most experiments go much longer than that and have more detected units. Note if this was manually increased, we are still in some sense scanning each unit in order (not the first # of spikes of each unit, which would have separate slicing logic).

### DIFF
--- a/nwbinspector/checks/ecephys.py
+++ b/nwbinspector/checks/ecephys.py
@@ -55,7 +55,8 @@ def check_electrical_series_reference_electrodes_table(electrical_series: Electr
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
 def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int = 200):
     """Check if a Units table has spike times that occur outside of observed intervals."""
-    if units_table.obs_intervals:
+    if not units_table.obs_intervals:
+        return
         for spike_time in units_table.spike_times.data[:nelems]:
             if not any((obs_int[0] <= spike_time <= obs_int[1] for obs_int in units_table.obs_intervals.data)):
                 return InspectorMessage(
--- a/nwbinspector/checks/ecephys.py
+++ b/nwbinspector/checks/ecephys.py
@@ -50,3 +50,17 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
 def check_electrical_series_reference_electrodes_table(electrical_series: ElectricalSeries):
     if electrical_series.electrodes.table.name != "electrodes":
         return InspectorMessage(message="electrodes does not  reference an electrodes table.")
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int = 200):
+    """Check if a Units table has spike times that occur outside of observed intervals."""
+    if units_table.obs_intervals:
+        for spike_time in units_table.spike_times.data[:nelems]:
+            if not any((obs_int[0] <= spike_time <= obs_int[1] for obs_int in units_table.obs_intervals.data)):
+                return InspectorMessage(
+                    message=(
+                        "This Units table contains spike times that occur during periods of time not labeled as being "
+                        "observed intervals."
+                    )
+                )
--- a/nwbinspector/checks/ecephys.py
+++ b/nwbinspector/checks/ecephys.py
@@ -58,7 +58,7 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int
     if not units_table.obs_intervals:
         return
     for spike_time in units_table.spike_times.data[:nelems]:
-        if not any((obs_int[0] <= spike_time <= obs_int[1] for obs_int in units_table.obs_intervals.data)):
+        if not any((start <= spike_time <= stop for start, stop in units_table.obs_intervals.data)):
             return InspectorMessage(
                 message=(
                     "This Units table contains spike times that occur during periods of time not labeled as being "
--- a/nwbinspector/checks/ecephys.py
+++ b/nwbinspector/checks/ecephys.py
@@ -53,12 +53,22 @@ def check_electrical_series_reference_electrodes_table(electrical_series: Electr
 
 
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
-def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int = 200):
+def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int = 4):
     """Check if a Units table has spike times that occur outside of observed intervals."""
     if not units_table.obs_intervals:
         return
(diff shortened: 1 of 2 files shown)


nwbinspector/checks/ecephys.py
```diff
@@ -50,3 +50,17 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
 def check_electrical_series_reference_electrodes_table(electrical_series: ElectricalSeries):
     if electrical_series.electrodes.table.name != "electrodes":
         return InspectorMessage(message="electrodes does not  reference an electrodes table.")
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int = 200):
+    """Check if a Units table has spike times that occur outside of observed intervals."""
+    if units_table.obs_intervals:
```
bendichter: I prefer early stopping. Less indented code and easier to read as you don't need to check far down the line for an `else` clause.

```suggestion
    if not units_table.obs_intervals:
        return
```

nwbinspector/checks/ecephys.py
```diff
@@ -50,3 +50,17 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
...
+        for spike_time in units_table.spike_times.data[:nelems]:
+            if not any((obs_int[0] <= spike_time <= obs_int[1] for obs_int in units_table.obs_intervals.data)):
```
bendichter: A single row of obs_intervales in a matrix of shape (n_intervals, 2)

CodyCBakerPhD: Yep. The iteration of `obs_int` over `units_table.obs_intervals.data` by default iterates over the `n_intervals`. The lower and upper bounds are sliced from the `0` and `1` indices of each length 2 `obs_int`

nwbinspector/checks/ecephys.py
```diff
@@ -50,3 +50,18 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
...
         return InspectorMessage(message="electrodes does not  reference an electrodes table.")
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_spike_times_not_in_unobserved_interval(units_table: Units, nelems: int = 200):
+    """Check if a Units table has spike times that occur outside of observed intervals."""
+    if not units_table.obs_intervals:
+        return
+    for spike_time in units_table.spike_times.data[:nelems]:
+        if not any((obs_int[0] <= spike_time <= obs_int[1] for obs_int in units_table.obs_intervals.data)):
```
bendichter: This checks if the spike times are *exclusively* outside of the observation interval. It should check whether any spikes exist outside the obs intervals

CodyCBakerPhD: Sorry, having trouble understanding you. Can you provide an example of a spike time and observation intervals that you're thinking of?

bendichter: spike_times = [1,2,3,4,5,6] and obs_intervals=[[0,2.5],[3.5,7]] should fail because there is a spike at 3 which is outside the observation intervals

CodyCBakerPhD: That's essentially the case captured by this test: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/168/files#diff-c94c2af000eb18300d99f65431873ae8b02d1d924c9c2a726aa1738783ba51d3R170-R183

But I went ahead and added your example as well (can never have too many tests!)

nwbinspector/checks/ecephys.py
bendichter: oh ok I misread this before. I think this could be vectorized a bit better but this should be fine for now

nwbinspector/checks/ecephys.py
bendichter:
```suggestion
        if not any((start <= spike_time <= stop for start, stop in units_table.obs_intervals.data)):
```

nwbinspector/checks/ecephys.py
CodyCBakerPhD: Yeah, it's optimized for low-memory usage and early break as opposed to fast speed of evaluating the total logic for all elements

bendichter:
I am not sure this would work for more than one unit as it is currently written. I think you'll need to use `units["obs_intervals"][:]`

CodyCBakerPhD:
> I am not sure this would work for more than one unit as it is currently written.

Multiple units would require indexing utilizing the `spike_times_index`, but the `spike_times` data itself is just a length-one vector so we get as many units as fit in the first 200 elements, which as I mention in PR description is pretty unlikely to be more than the first unit; however, if we grab the first 200/n_units number of elements of each unit, we likewise won't span very much time for even a single unit.

bendichter:
let's discuss this in our meeting today

CodyCBakerPhD:
@bendichter OK logic for this should be fixed now. Truncates based on `nunits` instead of `nelems`, scans ALL spike times for each unit and compares each to its own observation intervals. Also uses numpy vectorized logic:

Note `logical_and` is needed for vectorized left/right bound since the `start <= element <= stop` syntax doesn't seem to play nice with `numpy`.

Also note `logical_or` for combining results of the bound check only works if you have exactly two observation intervals, whereas `sum` will aggregate the logic over an arbitrary number.

nwbinspector/checks/ecephys.py
```diff
@@ -50,3 +50,28 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
...
+    for unit_spike_times, unit_obs_intervals in zip(
+        units_table["spike_times"][:nunits], units_table["obs_intervals"][:nunits]
+    ):
+        spike_times_array = np.array(unit_spike_times)
+        if not all(
+            sum(
+                [
+                    np.logical_and(start <= spike_times_array, spike_times_array <= stop)
+                    for start, stop in unit_obs_intervals
+                ]
```
bendichter, line 69: I don't think you need the list brackets here

nwbinspector/checks/ecephys.py
```diff
@@ -50,3 +50,28 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries):
...
+def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int = 4):
+    """Check if a Units table has spike times that occur outside of observed intervals."""
+    if not units_table.obs_intervals:
+        return
+    for unit_spike_times, unit_obs_intervals in zip(
+        units_table["spike_times"][:nunits], units_table["obs_intervals"][:nunits]
+    ):
+        spike_times_array = np.array(unit_spike_times)
+        if not all(
+            sum(
```
bendichter, line 65:
```suggestion
            np.sum(
```

PR #173, opened 2022-04-12
[Feature]: swap to ProcessPool and add progress bar

CodyCBakerPhD:
# Motivation

While adding a progress bar, it was revealed that my previous timings comparing normal usage to multithreading were biased to running on files not subject to a disk bottle-neck (original files were located on a built-in SSD). When moved to an external USB drive, this became I/O bound and behaved similarly to what we see with S3 streaming in PR #166. Easy fix for this is just swap to using the `ProcessPoolExecutor` all the time, since it gives nearly equivalent speed increase without relying on system properties.

The first attached video ('inspector_progress_bar_demo.wmv') shows how there is no noticeable speed difference between non-threading and multithreading in this case. The second video ('inspector_progress_bar_demo_process_instead_of_threading') confirms that multiple processes speeds up inspection, similar to what we saw in PR #166. The movies also demonstrate the progress bar.

### Small change to behavior
Another original reason for using threading was to have allow message-by-message generator behavior  even when `n_jobs` is specified, which is not possible with the `ProcessPoolExecutor` (generators cannot be pickled, so we have to scan an entire individual NWBFile at once within each process).

However, I would say that if a true message-by-message generator is what someone really wants to do (*e.g.* to find if there is a single error or not; `next(inspect_all(..., n_jobs=-1))`), they should not be using parallelization unless they're looking for a very small needle (like a single check) in a very large haystack (a very large number of NWBFiles). And in that case, they might as well just grab all checks instead of doing lazy iteration.  Otherwise, so many of our tests have early exit conditions that violation-free files take next to no time to scan. Instead, they'd have better performance just doing `next(inspect_all(..., n_jobs=1))`.

https://user-images.githubusercontent.com/51133164/163027835-44681382-0b04-4b85-ad3a-cb1e39f0c050.mp4

https://user-images.githubusercontent.com/51133164/163027837-b7dd3f25-9980-45ab-8796-3ca0cbf17388.mp4

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from collections.abc import Iterable
 from enum import Enum
 from typing import Optional, List
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ProcessPoolExecutor, as_completed
 from types import FunctionType
 from warnings import filterwarnings
 from distutils.util import strtobool
@@ -16,6 +16,7 @@ from distutils.util import strtobool
 import click
 import pynwb
 import yaml
+from tqdm import tqdm
 
 from . import available_checks
 from .inspector_tools import (
@@ -184,6 +185,7 @@ def configure_checks(
     ),
     is_flag=True,
 )
+@click.option("--progress-bar", help="Sett this flag to False to disable the display of the progress bar.")
 def inspect_all_cli(
     path: str,
     modules: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
This would also open the door again to absorbing the `inspect_dandiset` function from PR #166 into `inspect_all` since that parallelization type was the biggest difference between the two.

nwbinspector/nwbinspector.py
```diff
@@ -292,28 +309,39 @@ def inspect_all(
...
+                    )
+                )
+            if progress_bar:
+                for future in tqdm(as_completed(futures), **progress_bar_options):
+                    for message in future.result():
+                        yield message
+            else:
+                for future in as_completed(futures):
+                    for message in future.result():
+                        yield message
```
```suggestion
            completed_futures = as_completed(futures)
            if progress_bar:
                completed_futures = tqdm(completed_futures, **progress_bar_options)
            for future in completed_futures:
                for message in future.result():
                    yield message
```

bendichter:
This looks great! Any chance we could also have a progress bar when used as a Python library?

CodyCBakerPhD:
> Any chance we could also have a progress bar when used as a Python library?

Comes for free - do you want a video of it in an iPython console? It looks exactly the same though

bendichter:
ah yes, ok perfect

nwbinspector/nwbinspector.py
CodyCBakerPhD: Ah yes, thought about this but wasn't sure if calling `as_completed()` acted lazily; tried it ought and it seems to.

bendichter (approved):

PR #174, opened 2022-04-12
[Add Check]: feasible values of lambda

### LINKED ISSUE #17: [Add Check]: ophys checks to add
* [ ]  image_mask dimensions do not match dimensions of associated ImageSeries
* [x] wavelength is not in nm

bendichter:


CodyCBakerPhD:
@bendichter Thanks, is this something in the NWB schema that is hard to check for in PyNWB, or should there be a Best Practice section for it?

bendichter:
I think it's fine to not have in best practices, given that it is clearly and explicitly stated in the schema

CodyCBakerPhD (approved):

PR #178, opened 2022-04-13
add check and tests for spatial_series with improper dimensions

bendichter:


nwbinspector/checks/behavior.py
```diff
@@ -0,0 +1,9 @@
+from pynwb.behavior import SpatialSeries
+
+from ..register_checks import register_check, Importance, InspectorMessage
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=SpatialSeries)
+def check_spatial_series_dims(spatial_series: SpatialSeries):
+    if len(spatial_series.data.shape) > 1 and spatial_series.data.shape[1] > 3:
+        return InspectorMessage("SpatialSeries should have either 2 columns (x,y) or 3 columns (x,y,z).")
```
CodyCBakerPhD: What about 1 column (x) for linear tracks?

bendichter: yes, those will be false with `if len(spatial_series.data.shape) > 1`, so they will pass the check

bendichter: I just added a test for this to make sure it works as intended

CodyCBakerPhD: Right, I just mean in the message string itself (sorry, I selected only that one line when I made the comment so I thought it would only highlight that line only)

CodyCBakerPhD: I.e., "SpatialSeries should have either 1 column (x), 2 columns (x,y)..."

bendichter: ah ok you're right

bendichter:
```suggestion
        return InspectorMessage("SpatialSeries should have 1 column (x, ), 2 columns (x, y), or 3 columns (x, y, z).")
```

CodyCBakerPhD (approved):

PR #182, opened 2022-04-15
[Reduction]: remove coloration

CodyCBakerPhD:
As per discussion, coloration has been broken since the use of generalized formatting (technically it might still work if you use `levels=["file_path", "importance"]`). Given what a minor feature this is, even though it's 'nice' to have, it is proving to be quite the time investment to actually maintain.

If we end up with some free time we can always revisit getting this to work and maybe even how to make nice tests for it, but until then I think it's best to remove from the main code base. It'll always be here in these commit hashes and previous versions if we need to refer to it again.

### DIFF
--- a/nwbinspector/inspector_tools.py
+++ b/nwbinspector/inspector_tools.py
@@ -269,7 +269,7 @@ def format_messages(
     return formatted_messages
 
 
-def print_to_console(formatted_messages: List[str], no_color: bool = False):
+def print_to_console(formatted_messages: List[str]):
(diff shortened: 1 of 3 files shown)


bendichter:
we need to remove this from the cli too, right?

CodyCBakerPhD:
> we need to remove this from the cli too, right?

Ah, right.

CodyCBakerPhD:
@bendichter removed from CLI

bendichter (approved):

PR #193, opened 2022-04-20
[Feature]: Allow 'importance_threshold' string inputs

### LINKED ISSUE #192: [Feature]: Allow string input to 'importance_threshold'
### What happened?

When passing `importance_threshold` as a string, it can't be checked against Importance members.
It could be specific to Python 3.8, from what I gathered googling it.

### Steps to Reproduce

```shell
from nwbinspector import inspect_nwb

inspection_generator = inspect_nwb(
    nwbfile_path=file_path, 
    importance_threshold="BEST_PRACTICE_VIOLATION"
)
inspection_messages = list(inspection_generator)
```

### Traceback

```shell
Traceback (most recent call last):
  File "convert_all.py", line 241, in <module>
    inspection_messages = list(inspection_generator)
  File "/media/luiz/storage/Github/nwbinspector/nwbinspector/nwbinspector.py", line 451, in inspect_nwb
    checks = configure_checks(
  File "/media/luiz/storage/Github/nwbinspector/nwbinspector/nwbinspector.py", line 123, in configure_checks
    if importance_threshold not in Importance:
  File "/home/luiz/anaconda3/envs/env_oconnor/lib/python3.8/enum.py", line 352, in __contains__
    raise TypeError(
TypeError: unsupported operand type(s) for 'in': 'str' and 'EnumMeta'
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.8

### Package Versions

_No response_

### Code of Conduct

luiztauffer:
## Motivation

Fix #192 

## How to test the behavior?
```
```

## Checklist


nwbinspector/nwbinspector.py
```diff
@@ -447,6 +447,13 @@ def inspect_nwb(
         Skip the PyNWB validation step. This may be desired for older NWBFiles (< schema version v2.10).
         The default is False, which is also recommended.
     """
+    # Check importance_threshold passed as string
+    if isinstance(importance_threshold, str):
+        if importance_threshold in Importance.__members__:
+            importance_threshold = getattr(Importance, importance_threshold)
+        else:
+            raise ValueError(f"'{importance_threshold}' is not a valid importance_threshold value.")
+
```
CodyCBakerPhD, line 456: This is a bit more complicated than we actually need for this: [note how the CLI handles it](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/nwbinspector.py#L252) (`Enum` objects like our `Importance` are indexable just like dictionaries)

```python
importance_threshold = Importance[importance_threshold] if isinstance(importance_threshold, str) else importance_threshold
```

should be sufficient. Also not going to get into explicit `ValueError` type checking for every single argument - supported input types are specified in the type annotations and docstrings of every function and should be followed to avoid unintended behaviors.

CodyCBakerPhD: Also, might as well copy that same one-liner logic (assuming it works) to `inspect_all` as well. And maybe `configure_checks`?

luiztauffer: the one-liner might fail for string inputs with any typo, right? What's wrong with checking for it?

> Also, might as well copy that same one-liner logic (assuming it works) to inspect_all as well. And maybe configure_checks?

sure!

CodyCBakerPhD: >the one-liner might fail for string inputs with any typo, right? What's wrong with checking for it?

The error message from the key error will be enough of an error message in that respect. Not our responsibility to be a spell-checker, which is actually the original reasoning for not even supporting string inputs like this to begin with.

CodyCBakerPhD: Oh, also, be sure to update the annotation type to `Optional[Union[str, Importance]]`

CodyCBakerPhD:
Implemented and merged in #221

PR #195, opened 2022-04-20
handle if strings are returned as bytes

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist


CodyCBakerPhD:
Most annoying thing about this issue is we may have to do it for all checks with any field that *could* be written that way for older files. Any way we can generalize this as an auxiliary function?

bendichter:
@CodyCBakerPhD this doesn't have to do with the age of the file, it has to do with the version of h5py being used to read the file. This could apply to any vector of strings, including for instance the experimenter field. I suppose we could refactor this into a function, but it's 2 lines right now. Refactoring it could reduce it from 2 lines to 1 line, but I don't know if that's really worth it.

CodyCBakerPhD (approved):

PR #196, opened 2022-04-21
Add copy button to code blocks

weiglszonja:
## Motivation

Add a copy button to the right side of the code blocks. 
Size, opacity and other properties can be changed in `custom.css`

## How to test the behavior?

Check the appearance of the button with `sphinx-build`

## Checklist

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -14,3 +14,12 @@ dl.class em:not([class]):last-of-type::after {
 dl.class > dt:first-of-type {
     display: block !important;
 }
+
+button.copybtn {
+    height:25px;
+    width:25px;
+    opacity: 0.5;
+    padding: 0;
+    border: none;
+    background: none;
+}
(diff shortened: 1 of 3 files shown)


docs/_static/css/custom.css
```diff
@@ -14,3 +14,12 @@ dl.class em:not([class]):last-of-type::after {
 dl.class > dt:first-of-type {
     display: block !important;
 }
+
+button.copybtn {
+    height:25px;
+    width:25px;
+    opacity: 0.5;
```
weiglszonja, line 21: This way the button is always visible but faded (50% visibility) even when the mouse is not hovered over the code block. I can change it to be only visible when hovered over, background etc. if you prefer it differently.

bendichter:
@weiglszonja Thanks! Can you include a screenshot for changes like this?

weiglszonja:
This is how the copy button appears: 
button_screenshot

And when it is pressed:
button_press

CodyCBakerPhD:
Looks great! Thanks @weiglszonja!

CodyCBakerPhD (approved):

PR #204, opened 2022-04-26
[Bug] encapsulate nwbfile after io.read() in same try/except

CodyCBakerPhD:
Small issue came up recently when running NWBInspector on a file that was not able to be opened through PyNWB due to a `ConstructionError`.

We previously had this try/except wrapped around all our calls to `nwbfile = io.read()`, however, this particular instance of the error did not occur during those try/except clauses, it occurred during the `pynwb.validation` step (which acts directly on the open `io`, but can still cause actual `Errors` instead of just returning messages of those errors.

Unable to write CI tests for this since I can't really 'create' a simple unreadable NWBFile like that (also, the file it was triggered on is not openly shared yet). Open to suggestions on this.

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -454,8 +454,8 @@ def inspect_nwb(
     nwbfile_path = str(nwbfile_path)
     filterwarnings(action="ignore", message="No cached namespaces found in .*")
     filterwarnings(action="ignore", message="Ignoring cached namespace .*")
-    if not skip_validate:
-        with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True, driver=driver) as io:
+    with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True, driver=driver) as io:
+        if not skip_validate:
             validation_errors = pynwb.validate(io=io)
             for validation_error in validation_errors:
                 yield InspectorMessage(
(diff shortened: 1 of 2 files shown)


bendichter:
This sounds like a bug in `pynwb.validation`.

CodyCBakerPhD:
> This sounds like a bug in `pynwb.validation`.

If so, then it seems it could be a common one with many potential causes (any time a content fails to build for whatever reason), so we ought to account for it here as much as possible.

CodyCBakerPhD:
- [x] Break out use of try/except

CodyCBakerPhD:
@bendichter Try/except is now limited to the `io.read()` call and deployment over checks: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/fix_try_except_pynwb_validation/nwbinspector/nwbinspector.py#L469-L481. Can confirm this fixes the issue specific to the problematic file.

I tried using a single open `io` for both validate and then, and then `io.close()` after all exceptions are captured, but Windows CI struggles to properly close it apparently, so just broke each step into a separate context.

nwbinspector/nwbinspector.py
```diff
@@ -454,30 +454,31 @@ def inspect_nwb(
...
+            for inspector_message in run_checks(nwbfile=nwbfile, checks=checks):
+                inspector_message.file_path = nwbfile_path
+                yield inspector_message
+    except Exception as ex:
+        yield InspectorMessage(
+            message=traceback.format_exc(),
+            importance=Importance.ERROR,
+            check_function_name=f"During io.read() - {type(ex)}: {str(ex)}",
+            file_path=nwbfile_path,
+        )
```
rly: Just a suggestion: I am pretty sure you can combine the NWBHDF5IO calls into one.

This code:
```python
with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True, driver=driver) as io:
        if not skip_validate:
            validation_errors = pynwb.validate(io=io)
            for validation_error in validation_errors:
                yield InspectorMessage(
                    message=validation_error.reason,
                    importance=Importance.PYNWB_VALIDATION,
                    check_function_name=validation_error.name,
                    location=validation_error.location,
                    file_path=nwbfile_path,
                )

        try:
            nwbfile = io.read()
            for inspector_message in run_checks(nwbfile=nwbfile, checks=checks):
                inspector_message.file_path = nwbfile_path
                yield inspector_message
        except Exception as ex:
            yield InspectorMessage(
                message=traceback.format_exc(),
                importance=Importance.ERROR,
                check_function_name=f"During io.read() - {type(ex)}: {str(ex)}",
                file_path=nwbfile_path,
            )
```
passed pytest on my local Windows machine.

CodyCBakerPhD: Yup, this works. Thanks!

rly:
This doesn't address the original issue of an error occurring during validate (which would probably be due to a bug in pynwb). I hope those are rare, but perhaps they are worth catching. If so, you could wrap `pynwb.validate` with a different (or the same) try/except.

CodyCBakerPhD:
> This doesn't address the original issue of an error occurring during validate (which would probably be due to a bug in pynwb). I hope those are rare, but perhaps they are worth catching. If so, you could wrap pynwb.validate with a different (or the same) try/except.

It turns out that wasn't the actual problem after all, this is the file that passes validate but fails read. The 'bug' is that validate doesn't return anything for that.

rly:
Ah, got it. Thanks for explaining

rly (approved):

PR #205, opened 2022-04-27
[Dependencies] Make dandi a lazy import

CodyCBakerPhD:
DANDI team is complaining that the circular dependency caused by the S3 helper function might cause a problem for them in the future. Making it a lazy import so it's not a strict dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ setup(
     include_package_data=True,
     url="https://github.com/NeurodataWithoutBorders/nwbinspector",
     install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "tqdm"],
+    extras_require=dict(dandi=["dandi>=0.39.2"]),
     entry_points={"console_scripts": ["nwbinspector=nwbinspector.nwbinspector:inspect_all_cli"]},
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
     packages=find_packages(),
     include_package_data=True,
     url="https://github.com/NeurodataWithoutBorders/nwbinspector",
-    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "dandi>=0.39.2", "tqdm"],
+    install_requires=["pynwb", "natsort", "click", "PyYAML", "jsonschema", "tqdm"],
```
yarikoptic, line 17: minor: I would recommend to formalize as `extra_requires={'dandi': ["dandi>=0.39.2"]}` (or whatever format it is) and then use `pip install -e .[dandi]` in your workflow.

CodyCBakerPhD: Not a bad idea, we use this on the CT for including non-minimal reqs

CodyCBakerPhD:
@bendichter good to merge?

bendichter (approved):

PR #218, opened 2022-07-07
[Bug]: Fix None fields with config

CodyCBakerPhD:
fix #216 

I swear this was working when I originally implemented the `config` feature, but possible that PR #96 introduced it (though could also swear I would have noticed that when originally testing it out in practice) but either way it evidently wasn't tested thoroughly enough.

I've now added an explicit test to ensure message properties are not corrupted by the config setting.

If interested in a technical reason why this happened, fix #96 worked by copying the check function from the registry, but neglected to also copy the auto-parsing wrapper applied to those checks after they are triggered (or more precisely, if they do not return `None`).

PR #219, opened 2022-07-07
Fix testing suite for PyNWB v2.1.0

CodyCBakerPhD:
Discovered from #218, fixed here. Raised also in https://github.com/NeurodataWithoutBorders/pynwb/issues/1497

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pynwb>=1.4.0,<2.1.0  # TODO, debug 2.1.0?
+pynwb
 PyYAML
 jsonschema
 packaging
--- a/tests/unit_tests/test_tables.py
+++ b/tests/unit_tests/test_tables.py
@@ -229,14 +229,18 @@ def test_check_single_row_pass():
 
 
 def test_check_single_row_ignore_units():
-    table = Units(name="Units",)  # default name when building through nwbfile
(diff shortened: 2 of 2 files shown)


bendichter:
If you want to remove x, y, and z, you'll need to set the pynwb requirement to >= 2.1.0

CodyCBakerPhD:
> If you want to remove x, y, and z, you'll need to set the pynwb requirement to >= 2.1.0

Note that the removal is purely for testing purposes, and testing always performs using latest requirements without having to enforce the minimal install on others. That said it's also not intended behavior so I expect them to fix it quite soon anyway. No change to core behavior is expected for our checks w.r.t <= 2.1.0

bendichter:
>  testing always performs using latest requirements without having to enforce the minimal install on others

what do you mean?

If I run `pytest` on my local machine, it won't be guaranteed to run the latest version of every package

CodyCBakerPhD:
> If I run pytest on my local machine, it won't be guaranteed to run the latest version of every package

I was narrowly thinking about only the CI workflows that use fresh installs of all the most recent dependencies unless an upper bound is applied.

I can see how that would be limiting for local testing, though, so swapped with a `pytest.skipIf` condition that will be removed once the upstream issue is fixed.

PR #223, opened 2022-07-11
Add retry for curl requests

### LINKED ISSUE #197: [Bug]: Can't read data (curl cannot perform request)
### What happened?

As I parse the remaining errors from the first attempt to inspect the entire archive, a number of repetitions of this issue occur. One basic example is in the traceback section below, but I've also attached all report files where it occurs, which spans multiple objects.

There also seem to be some related issues such as `ConstructErrors` and `KeyError: 'Unable to open object by token (unable to determine object type)'`.

In each of these cases, I am streaming the data via `ros3`, which could very well be a part of the problem. Interestingly, sometimes these curl errors occur outside of the `nwbfile = io.read() step` which causes the the dandiset to not be inspectable at all (which I'm considering a separate issue).

Anyway here is the full set of tracebacks available in the ERROR sections of each report...

[000008.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543040/000008.txt)
[000011.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543041/000011.txt)
[000035.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543042/000035.txt)
[000041.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543043/000041.txt)
[000043.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543044/000043.txt)
[000054.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543045/000054.txt)
[000121.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543047/000121.txt)
[000148.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543048/000148.txt)
[000211.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543051/000211.txt)
[000223.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/8543053/000223.txt)

### Steps to Reproduce

_No response_

### Traceback

```shell
0.0  <class 'OSError'>: Can't read data (curl cannot perform request)
---------------------------------------------------------------------

0.0.0  sub-mouse-GHNSF/sub-mouse-GHNSF_ses-20190828-sample-3_slice-20190828-slice-3_cell-20190828-sample-3_icephys.nwb: 'None' object with name 'None'
         Message: Traceback (most recent call last):
  File "/home/jovyan/GitHub/nwbinspector/nwbinspector/nwbinspector.py", line 476, in inspect_nwb
    nwbfile = io.read()
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 498, in read
    return call_docval_func(super().read, kwargs)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/utils.py", line 424, in call_docval_func
    return func(*fargs, **fkwargs)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/io.py", line 37, in read
    f_builder = self.read_builder()
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/utils.py", line 583, in func_call
    return func(args[0], **pargs)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 515, in read_builder
    f_builder = self.__read_group(self.__file, ROOT_NAME, ignore=ignore)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 656, in __read_group
    builder = read_method(sub_h5obj)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 656, in __read_group
    builder = read_method(sub_h5obj)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 656, in __read_group
    builder = read_method(sub_h5obj)
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/hdmf/backends/hdf5/h5tools.py", line 683, in __read_dataset
    scalar = h5obj[()]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/home/jovyan/my-conda-envs/nwbct_3_9/lib/python3.9/site-packages/h5py/_hl/dataset.py", line 775, in __getitem__
    self.id.read(mspace, fspace, arr, mtype)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 192, in h5py.h5d.DatasetID.read
  File "h5py/_proxy.pyx", line 112, in h5py._proxy.dset_rw
OSError: Can't read data (curl cannot perform request)
```

### Operating System

Linux

### Python Executable

Python

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #197 

Attempting to resolve curl requests errors to S3 paths, up to `max_retries` using the suggested exponential back-off

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -459,7 +459,7 @@ def inspect_nwb(
     filterwarnings(action="ignore", message="No cached namespaces found in .*")
     filterwarnings(action="ignore", message="Ignoring cached namespace .*")
 
-    def _collect_all_messages(nwbfile_path: FilePathType, driver: Optional[str] = None, skip_validate: bool = False):
+    def _collect_all_messages(nwbfile_path: FilePathType, checks: list, driver: Optional[str] = None, skip_validate: bool = False):
         with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True, driver=driver) as io:
             if not skip_validate:
                 validation_errors = pynwb.validate(io=io)
@@ -477,6 +477,8 @@ def inspect_nwb(
                 for inspector_message in run_checks(nwbfile=nwbfile, checks=checks):
                     inspector_message.file_path = nwbfile_path
                     yield inspector_message
+            except OSError as ex:
+                raise ex  # propagate outside private
             except Exception as ex:
                 yield InspectorMessage(
                     message=traceback.format_exc(),
@@ -486,14 +488,14 @@ def inspect_nwb(
                 )
 
     if driver != "ros3":
-        yield _collect_all_messages(nwbfile_path=nwbfile_path, driver=driver, skip_validate=skip_validate)
+        yield _collect_all_messages(nwbfile_path=nwbfile_path, checks=checks, driver=driver, skip_validate=skip_validate)
     else:
         retries = 0
 
         while retries < max_retries:
             try:
                 retries += 1
-                yield _collect_all_messages(nwbfile_path=nwbfile_path, driver=driver, skip_validate=skip_validate)
+                yield _collect_all_messages(nwbfile_path=nwbfile_path, checks=checks, driver=driver, skip_validate=skip_validate)
             except OSError:  # Cannot curl request
                 sleep(0.1 * 2 ** retries)
 
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -493,7 +493,7 @@ def inspect_nwb(
                 )
 
     if driver != "ros3":
-        yield _collect_all_messages(
+        return _collect_all_messages(
             nwbfile_path=nwbfile_path, checks=checks, driver=driver, skip_validate=skip_validate
         )
     else:
@@ -502,7 +502,7 @@ def inspect_nwb(
         while retries < max_retries:
             try:
                 retries += 1
-                yield _collect_all_messages(
+                return _collect_all_messages(
                     nwbfile_path=nwbfile_path, checks=checks, driver=driver, skip_validate=skip_validate
                 )
             except OSError:  # Cannot curl request
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -13,6 +13,7 @@ from concurrent.futures import ProcessPoolExecutor, as_completed
 from types import FunctionType
 from warnings import filterwarnings, warn
 from distutils.util import strtobool
+from time import sleep
 
 import click
 import pynwb
@@ -153,10 +154,7 @@ def configure_checks(
 @click.argument("path")
 @click.option("--modules", help="Modules to import prior to reading the file(s).")
 @click.option(
-    "--report-file-path",
-    default=None,
-    help="Save path for the report file.",
-    type=click.Path(writable=True),
+    "--report-file-path", default=None, help="Save path for the report file.", type=click.Path(writable=True),
 )
 @click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("--levels", help="Comma-separated names of InspectorMessage attributes to organize by.")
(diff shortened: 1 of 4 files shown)


bendichter:
I think this looks good but perhaps we need an importable function that can help users do this

CodyCBakerPhD:
> I think this looks good but perhaps we need an importable function that can help users do this

I was toying around with some ways to do that; it was proving a tad tricky since the user (or ourselves) might wish to call any executable code on the open `io` handling the S3 requests - but my idea was to make a context class that does something like

```python
with retry_s3_call(max_retries=10):
    # try to do something on an open io
```

where the thing you try to do could be as simple as `nwbfile = io.read()` or `run_some_check_on_nwbfile(nwbfile=nwbfile)` etc, but wasn't sure how to construct such a context capable of running arbitrary code multiple times (or a function for that matter).

CodyCBakerPhD:
Though I suppose that a context is simply not possible for that kind of thing... https://stackoverflow.com/a/64770597

Perhaps a simple function like

```python
def try_until_success(command, attempts=1):
    for _ in range(attempts):
        try:
            return command()
        except Exception as exc:
            err = exc

    raise err
```

might work (from https://stackoverflow.com/a/64770679)

bendichter:
yeah that would be cool but I don't know if that's possible. Usually contexts pass in an object that is used within the context, e.g.

```python
with open("fpath", "r") as f:
    # use f here
```

I think the best approach might be to do two things:
1. adjust `NWBHDF5IO` to use retries if driver == "ros3"
2. create a `robust_read` function that can be used like 

```python
robust_read(nwbfile.acquisition["ElectricalSeries"].data, slice(1, 10), max_retries=5)
```
what do you think?

bendichter:
One tricky thing is that the HDF5 Datasets themselves don't have information about whether they are s3 or local. I can't think of a way to automatically switch to retrying if ros3 is being used. I suppose we could just always retry, but even that is tricky because we are interfacing with the h5py.Dataset objects directly. There are lots of ways we could potentially approach this. Let's discuss in our meeting today.

CodyCBakerPhD:
> I suppose we could just always retry

That's what I've done in latest commit using a very similar function to what you outlined in `2.` above. It shouldn't hurt in the non-S3 case since it will just execute and return the output nonetheless on first try (or first `retry`).

Could also put in switch at the corresponding deployment levels to determine whether to do it at the classic way or S3 way.

Will discuss this later, then.

bendichter (approved):

CodyCBakerPhD:
Summary: this patch is fine for now, it hits most of the major error points where this can occur - long term solution is to support `s3fs` through PyNWB and that should automatically handle this kind of thing much better (and offer improved performance) but only downside is Windows would still have to go through `ros3`.

Otherwise, a more sophisticated approach would be a decorator that wraps functions in the registry (and `io.read()`) and does more or less the same thing as this utility but would therefore allow less awkward function `arg`/`kwarg` passing.

PR #225, opened 2022-07-11
[New Check]: Non-required Electrode columns (nwb-schema >= 2.5.0)

### LINKED ISSUE #220: [Add Check]: add check for if electrodes table x, y, z are NaNs

With the new pynwb release, 2.1.0, you no longer need to fill x, y, and z with `np.nan`. We should make a suggestion to remove these if they are present. Maybe this suggestion should only show up to the nwb schema version is >= 2.5.0. What do you think, @CodyCBakerPhD? Should we make version-specific messages? 

_No response_

No.

### Code of Conduct

CodyCBakerPhD:
fix #220

Early work on adding this check. Currently relies (incorrectly) on grabbing this information from the PyNWB version being used to read the files (>= 2.1.0 release) however that would not correctly apply to a file written using a cached namespace corresponding to any `nwb-schema<2.5.0`.

A couple of design considerations...

i) how to get the cached namespace core version? I tried the `NWBHDF5IO` nwbfile and such for this but didn't have any luck - `nwbfile.namespaces` just reports the string `"core"` and not the version. I was able to get this information however by the following

```python
str(next(iter(file["specifications"]["core"])))
```

applied to a `file = h5py.File(...)`

ii) how best to communicate this version information from the io/file reading level of the `inspect_nwb`/`run_checks` level of the inspector and the ultimate check function (which is really only designed to receive one input, that being the object being checked.


nwbinspector/checks/ecephys.py
```diff
@@ -79,3 +81,17 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
...
             )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=ElectrodeTable)
+def check_optional_columns_electrode_table(electrode_table: ElectrodeTable):
+    """Check that the ElectrodeTable columns which are optional for nwb-schema>=2.5.0 are not specified as NaN."""
+    if get_package_version(name="pynwb") < version.Version("2.1.0"):
+        return
+    if any(x.isnan() in electrode_table for x in ["x", "y", "z", "imp", "filtering"]):
+        return InspectorMessage(
```
bendichter, line 92: I would yield here and say in the message which column is Nan. Also, does `isnan` work for an array? Maybe we could say `if np.all(np.isnan(col))`

tests/unit_tests/test_ecephys.py
```diff
@@ -220,3 +225,43 @@ def test_check_spike_times_not_in_unobserved_interval_multiple_units():
...
+        x=np.nan,
+        y=np.nan,
+        z=np.nan,
+        imp=np.nan,
+        location="unknown",
+        filtering="unknown",
+        group=ElectrodeGroup(name="test_group", description="", device=Device(name="test_device"), location="unknown"),
+        group_name="test_group",
+    )
+    if get_package_version(name="pynwb") >= version.Version("2.1.0"):
```
bendichter, line 254: I think this logic should go inside the skip statement. From the comments it seems like maybe you agree?

CodyCBakerPhD: This one is temporary until they fix the bug - this PR will have to stay in draft until then.

nwbinspector/checks/ecephys.py
CodyCBakerPhD: Sure, do you want it to report the exact index(ices) that are `NaN` or just the column that it was found in?

PR #227, opened 2022-07-11
[New Check]: Add experimenter form check using DANDI regex

### LINKED ISSUE #33: [Add Check]: Check that if experimenters string does not have comma
If a user provides a comma-separated scalar string for `nwbfile.experimeter`, the inspector should recommend that they instead store an array of experimenter names. Same for keywords and related publications.

CodyCBakerPhD:
fix #33

Checks the text form of each experimenter against the DANDI regex pattern (imported directly from there in case they ever update it).

Check has early return in the event that DANDI is not installed (with informative warning message).

Also incorporates that byte-string fix we needed for publications, which has a similar kind of structure to this.

### DIFF
--- a/nwbinspector/checks/nwbfile_metadata.py
+++ b/nwbinspector/checks/nwbfile_metadata.py
@@ -1,11 +1,13 @@
 """Check functions that examine general NWBFile metadata."""
 import re
 from datetime import datetime
+from warnings import warn
 
 from pynwb import NWBFile, ProcessingModule
 from pynwb.file import Subject
 
 from ..register_checks import register_check, InspectorMessage, Importance
+from ..utils import is_module_installed
 
 duration_regex = (
     r"^P(?!$)(\d+(?:\.\d+)?Y)?(\d+(?:\.\d+)?M)?(\d+(?:\.\d+)?W)?(\d+(?:\.\d+)?D)?(T(?=\d)(\d+(?:\.\d+)?H)?(\d+(?:\.\d+)"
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Oh yeah, SS of how the warning appears (in a small window anyway):

image

EDIT: Not applicable any more

nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -38,12 +40,35 @@ def check_session_start_time_future_date(nwbfile: NWBFile):
...
+
+    for experimenter in nwbfile.experimenter:
+        experimenter = experimenter.decode() if isinstance(experimenter, bytes) else experimenter
+        if re.match(string=experimenter, pattern=NAME_PATTERN) is None:
+            yield InspectorMessage(
+                message=(
+                    f"The name of experimenter '{experimenter}' does not match the DANDI form "
+                    "(Last, First Middle or Last, First M.)."
+                )
+            )
```
bendichter: Maybe if dandi isn't installed we can default to implementing this pattern locally

CodyCBakerPhD: Sure thing, added.

nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -1,11 +1,13 @@
 """Check functions that examine general NWBFile metadata."""
 import re
 from datetime import datetime
+from warnings import warn
```
bendichter:
```suggestion
```

bendichter (approved):

PR #230, opened 2022-07-13
Add caching for repeated table data calls

### LINKED ISSUE #177: [Feature]: lru caching for repeated table column data access

Not a huge deal, but I notice it might increase performance on various dandisets to reduce the number of repeated data access operations by the table checks (or anything else that extracts the same limited amount of data across multiple checks).

We had previously hypothesized lru caching for this, so adding it to the to-do list.

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #177 

Many of the checks we have for tables rely on extracting a specifiable amount of data (`nelems`) from the columns to enact some logic upon. This results in the same pattern of data extraction being applied to the same table column potentially multiple times.

This PR adds caching for the input/output of these calls to allow instant retrieval. The default cache 'size' is `128` function items (pairs of input/output values), which I've bumped to `1000` to get some benefit for larger tables or files with many tables. Each of these pairs should be approximately 1.6 KB in size (assuming `float64` values) for a 1.6 MB cache when fully used.

### DIFF
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -9,7 +9,7 @@ from pynwb.file import TimeIntervals, Units
 
 from ..register_checks import register_check, InspectorMessage, Importance
 from ..utils import (
-    _subsample_data,
+    _cache_data_selection,
     format_byte_size,
     is_ascending_series,
     is_dict_in_string,
@@ -56,7 +56,7 @@ def check_time_interval_time_columns(time_intervals: TimeIntervals, nelems: int 
     unsorted_cols = []
     for column in time_intervals.columns:
         if column.name[-5:] == "_time":
-            if not is_ascending_series(column, nelems):
+            if not is_ascending_series(column.data, nelems):
                 unsorted_cols.append(column.name)
     if unsorted_cols:
         return InspectorMessage(
@@ -81,8 +81,8 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
         load the entire arrays.
     """
     if np.any(
-        np.asarray(_subsample_data(data=time_intervals["stop_time"], nelems=nelems))
-        - np.asarray(_subsample_data(data=time_intervals["start_time"], nelems=nelems))
+        np.asarray(_cache_data_selection(data=time_intervals["stop_time"].data, selection=slice(nelems)))
+        - np.asarray(_cache_data_selection(data=time_intervals["start_time"].data, selection=slice(nelems)))
         < 0
     ):
         return InspectorMessage(
@@ -111,7 +111,7 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
             if np.asarray(column.data[0]).itemsize == 1:
                 continue  # already boolean, int8, or uint8
             try:
-                unique_values = np.unique(_subsample_data(data=column, nelems=nelems))
+                unique_values = np.unique(_cache_data_selection(data=column.data, selection=slice(nelems)))
             except TypeError:  # some contained objects are unhashable or have no comparison defined
                 continue
             if unique_values.size != 2:
@@ -179,7 +179,7 @@ def check_table_values_for_dict(table: DynamicTable, nelems: int = 200):
     for column in table.columns:
         if not hasattr(column, "data") or isinstance(column, VectorIndex) or not isinstance(column.data[0], str):
             continue
-        for string in _subsample_data(data=column, nelems=nelems):
+        for string in _cache_data_selection(data=column.data, selection=slice(nelems)):
             if is_dict_in_string(string=string):
                 message = (
                     f"The column '{column.name}' contains a string value that contains a dictionary! Please "
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -153,7 +153,9 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
 def check_single_row(
-    table: DynamicTable, exclude_types: Optional[list] = (Units,), exclude_names: Optional[List[str]] = ("electrodes",),
+    table: DynamicTable,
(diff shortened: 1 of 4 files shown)


nwbinspector/utils.py
```diff
@@ -2,18 +2,30 @@
...
 PathType = TypeVar("PathType", str, Path)  # For types that can be either files or folders
 FilePathType = TypeVar("FilePathType", str, Path)
 OptionalListOfStrings = Optional[List[str]]
 
 dict_regex = r"({.+:.+})"
+MAX_CACHE_ITEMS = 1000  # lru_cache default is 128 calls of matching input/output, but might need more to get use here
+
+
+@lru_cache(maxsize=MAX_CACHE_ITEMS)
+def _subsample_data(data: np.ndarray, nelems: Optional[int] = 200):
```
bendichter:
```suggestion
def _subsample_data(data: Union[np.ndarray, h5py.Dataset], nelems: Optional[int] = 200):
```

the whole point here is to cache h5py dataset values, right?

nwbinspector/checks/tables.py
```diff
@@ -79,7 +80,11 @@ def check_time_intervals_stop_after_start(time_intervals: TimeIntervals, nelems:
         very long so you don't need to load the entire array into memory. Use None to
         load the entire arrays.
     """
-    if np.any(np.asarray(time_intervals["stop_time"][:nelems]) - np.asarray(time_intervals["start_time"][:nelems]) < 0):
+    if np.any(
+        np.asarray(_subsample_data(data=time_intervals["stop_time"], nelems=nelems))
+        - np.asarray(_subsample_data(data=time_intervals["start_time"], nelems=nelems))
```
```suggestion
        _subsample_data(data=time_intervals["stop_time"], nelems=nelems)
        - _subsample_data(data=time_intervals["start_time"], nelems=nelems)
```

do we really need the np.array wrappers?

nwbinspector/utils.py
```diff
@@ -52,9 +64,12 @@ def check_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
     return len(uniq_diff_ts) == 1
 
 
-def is_ascending_series(series: np.ndarray, nelems=None):
+def is_ascending_series(series: Union[VectorData, ArrayLike], nelems=None):
     """General purpose function for determining if a series is monotonic increasing."""
-    return np.all(np.diff(series[:nelems]) > 0)
+    if isinstance(series, VectorData):
```
bendichter: I'd rather test for h5py.Dataset.

nwbinspector/utils.py
bendichter: I also think I'd prefer to take a tuple of slice objects, so we can cache data more generally

CodyCBakerPhD: Done

nwbinspector/utils.py
CodyCBakerPhD: Done

nwbinspector/checks/tables.py
```diff
@@ -111,7 +111,7 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
             if np.asarray(column.data[0]).itemsize == 1:
                 continue  # already boolean, int8, or uint8
             try:
-                unique_values = np.unique(_subsample_data(data=column, nelems=nelems))
+                unique_values = np.unique(_cache_data_selection(data=column.data, selection=slice(nelems)))
```
bendichter, line 114:
```suggestion
                unique_values = np.unique(_cache_data_selection(data=column.data, selection=slice(0, nelems)))
```
should this be `slice(0, nelems)`?

nwbinspector/utils.py
```diff
@@ -2,18 +2,30 @@
...
 OptionalListOfStrings = Optional[List[str]]
 
 dict_regex = r"({.+:.+})"
+MAX_CACHE_ITEMS = 1000  # lru_cache default is 128 calls of matching input/output, but might need more to get use here
+
+
+@lru_cache(maxsize=MAX_CACHE_ITEMS)
+def _cache_data_selection(data: h5py.Dataset, selection: Union[slice, Tuple[slice]]) -> np.array:
+    """Extract the selection lazily from the data object for efficient caching (most beneficial during streaming)."""
+    return data[selection]
```
bendichter, line 31:
```suggestion
    data = np.asarray(data) if isinstance(data, list) else data
    return data[selection]
```

nwbinspector/utils.py
```diff
@@ -2,18 +2,31 @@
...
 PathType = TypeVar("PathType", str, Path)  # For types that can be either files or folders
 FilePathType = TypeVar("FilePathType", str, Path)
 OptionalListOfStrings = Optional[List[str]]
 
 dict_regex = r"({.+:.+})"
+MAX_CACHE_ITEMS = 1000  # lru_cache default is 128 calls of matching input/output, but might need more to get use here
+
+
+@lru_cache(maxsize=MAX_CACHE_ITEMS)
+def _cache_data_selection(data: h5py.Dataset, selection: Union[slice, Tuple[slice]]) -> np.array:
```
bendichter:
```suggestion
def _cache_data_selection(data: Union[h5py.Dataset, ArrayLike], selection: Union[slice, Tuple[slice]]) -> np.array:
```

CodyCBakerPhD:
OK finally got this working - some technical details arose related to the request to enable selections rather than just `nelems` - apparently, Python `slice`s are not hashable (a requirement of the LRU caching which is essentially just a global dictionary that `kwargs` get updated into): https://stackoverflow.com/questions/29980786/why-are-slice-objects-not-hashable-in-python

There is a simple workaround which is just to parse the representation of each slice in the tuple (which it can easily reconstruct afterward), which is hashable.

Also, other common return types from in-memory `VectorData.data`/`TimeSeries.data` such as lists are also not hashable and so the private helper function that decides whether or not to just slice the data directly or pass it through the `h5py.Dataset`-specific LRU cache (currently, the `h5py.Dataset` object is the only one that benefits from this caching, but when streaming that benefit should be worth it whenever the total quantity of data streamed is larger than the default HDF5 cache but the number of data access commands is smaller than the LRU cache size).

bendichter:
what finesse!

bendichter (approved):

PR #231, opened 2022-07-13
[New Check] Entire column of a table is not NaN

CodyCBakerPhD:
Along with #229, finishes #155

Whenever an entire column of a table is `NaN`, informs the user that there probably isn't a need to include it. Samples the `nelems` uniformly across the length of the column since it's quite possible to have a block of consecutive entries (even `nelems` many) that are purposefully and even informatively `NaN`.

Only giving it a `SUGGESTION` level of importance since this will fail for large columns of sparse non-`NaN`, and also could be a design choice of the user.

### DIFF
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -189,7 +189,7 @@ def check_col_not_nan(table: DynamicTable, nelems: int = 200):
     for column in table.columns:
         if not hasattr(column, "data") or isinstance(column, VectorIndex) or isinstance(column.data[0], str):
             continue
-        if not all(np.isnan(column[:nelems])):
+        if not all(np.isnan(column[:nelems]).flatten()):
             continue
 
         subindex_selection = np.unique(np.round(np.linspace(start=0, stop=column.shape[0] - 1, num=nelems)).astype(int))
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -148,9 +148,7 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
 def check_single_row(
-    table: DynamicTable,
-    exclude_types: Optional[list] = (Units,),
-    exclude_names: Optional[List[str]] = ("electrodes",),
+    table: DynamicTable, exclude_types: Optional[list] = (Units,), exclude_names: Optional[List[str]] = ("electrodes",),
 ):
     """
     Check if DynamicTable has only a single row; may be better represented by another data type.
@@ -191,6 +189,9 @@ def check_col_not_nan(table: DynamicTable, nelems: int = 200):
     for column in table.columns:
         if not hasattr(column, "data") or isinstance(column, VectorIndex) or isinstance(column.data[0], str):
             continue
+        if not all(np.isnan(column[:nelems])):
+            continue
+
         subindex_selection = np.unique(np.round(np.linspace(start=0, stop=column.shape[0] - 1, num=nelems)).astype(int))
         if all(np.isnan(column[subindex_selection])):
             yield InspectorMessage(
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -14,7 +14,6 @@ from ..utils import (
     is_ascending_series,
     is_dict_in_string,
     is_string_json_loadable,
-    safe_uniform_selection,
 )
 
 
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -199,7 +199,11 @@ def check_col_not_nan(table: DynamicTable, nelems: Optional[int] = 200):
         if nelems is not None and not all(np.isnan(column[:nelems]).flatten()):
             continue
 
-        if all(np.isnan(column[slice(0, None, np.ceil(length / nelems).astype(int) if nelems else None)])):
+        if all(
+            np.isnan(
+                column[slice(0, None, np.ceil(len(column.data) / nelems).astype(int) if nelems else None)]
+            ).flatten()
+        ):
             yield InspectorMessage(
                 message=f"Column {column.name} has all NaN values. Consider removing it from the table."
             )
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -200,8 +200,7 @@ def check_col_not_nan(table: DynamicTable, nelems: Optional[int] = 200):
         if nelems is not None and not all(np.isnan(column[:nelems]).flatten()):
             continue
 
-        subindex_selection = safe_uniform_selection(length=column.shape[0], nelems=nelems)
-        if all(np.isnan(column[subindex_selection])):
+        if all(np.isnan(column[slice(0, None, np.ceil(length / nelems).astype(int) if nelems else None)])):
             yield InspectorMessage(
                 message=f"Column {column.name} has all NaN values. Consider removing it from the table."
             )
--- a/nwbinspector/checks/tables.py
+++ b/nwbinspector/checks/tables.py
@@ -148,9 +148,7 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
 def check_single_row(
-    table: DynamicTable,
-    exclude_types: Optional[list] = (Units,),
-    exclude_names: Optional[List[str]] = ("electrodes",),
+    table: DynamicTable, exclude_types: Optional[list] = (Units,), exclude_names: Optional[List[str]] = ("electrodes",),
 ):
     """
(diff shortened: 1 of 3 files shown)


nwbinspector/checks/tables.py
```diff
@@ -183,3 +183,18 @@ def check_table_values_for_dict(table: DynamicTable, nelems: int = 200):
...
                 yield InspectorMessage(message=message)
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=DynamicTable)
+def check_col_not_nan(table: DynamicTable, nelems: int = 200):
+    """Check if all of the values in a single column of a table are NaN."""
+    for column in table.columns:
+        if not hasattr(column, "data") or isinstance(column, VectorIndex) or isinstance(column.data[0], str):
+            continue
+        subindex_selection = np.unique(np.round(np.linspace(start=0, stop=column.shape[0] - 1, num=nelems)).astype(int))
```
bendichter: what's going on here? Why not just slice with nelems?

nwbinspector/checks/tables.py
```diff
@@ -183,3 +183,18 @@ def check_table_values_for_dict(table: DynamicTable, nelems: int = 200):
...
+        if np.any(~np.isnan(column[subindex_selection])):
+            continue
+        else:
+            yield InspectorMessage(
+                message=f"Column {column.name} has all NaN values. Consider removing it from the table."
+            )
```
```suggestion
        if np.all(np.isnan(column[:nelems])):
            yield InspectorMessage(
                message=f"Column {column.name} has all NaN values. Consider removing it from the table."
            )
```

CodyCBakerPhD: It's written this way in case `np.any` has an early return condition once it detects a single non-`NaN` value

nwbinspector/checks/tables.py
CodyCBakerPhD: > Samples the nelems uniformly across the length of the column since it's quite possible to have a block of consecutive entries (even nelems many) that are purposefully and even informatively NaN.

Also see https://github.com/NeurodataWithoutBorders/nwbinspector/pull/229#issue-1302252779

bendichter: hmm ok. This has the downside of likely needing to access every chunk, so for chunked data at this point you are essentially reading the entire column

nwbinspector/checks/tables.py
bendichter: `np.all` can have the same early stopping

nwbinspector/checks/tables.py
CodyCBakerPhD: Would have to be a huge column (>125,000 rows of float elements) to have multiple chunks, though, but yes what you say is true - moreso for the TimeSeries one, but it's even more important to do it there since 200 otherwise consecutive elements of potentially high-frequency data isn't nearly enough to safely determine the logic.

nwbinspector/checks/tables.py
CodyCBakerPhD: Well, interesting results...

Indeed as you say, `all` does indeed have early stopping just like `any`...

```python
nelems = 1000000
array1 = [True for _ in range(nelems)]
array1 += [False]
array2 = [False]
array2 += [True for _ in range(nelems)]
start = time()
all(array1)
runtime = time() - start
print(f"Took {runtime}s!")
start = time()
all(array2)
runtime = time() - start
print(f"Took {runtime}s!")

----------------------------------

Took 0.006882190704345703s!
Took 3.719329833984375e-05s!
```

... however `numpy` does not appear to (likely because it's vectorized). Maybe that gives better performance for very large in-memory arrays, but for sparse inputs the early return from Python built-ins might be nicer.

```python
nelems = 1000000
array1 = [True for _ in range(nelems)]
array1 += [False]
array2 = [False]
array2 += [True for _ in range(nelems)]
start = time()
np.all(array1)
runtime = time() - start
print(f"Took {runtime}s!")
start = time()
np.all(array2)
runtime = time() - start
print(f"Took {runtime}s!") 


----------------------------------

Took 0.05324196815490723s!
Took 0.05084705352783203s!
```

nwbinspector/checks/tables.py
CodyCBakerPhD: How about for the TimeSeries one, we _first_ check if the first `nelemes` consecutive elements are all `NaN`, and if they are then check it again using the uniform sampling over the entire series? I do understand that if streaming on a long TimeSeries, the uniform sampling might even pull `nelems` number of total chunks, but this would make it so that we only request that much data if there is reason to do so.

bendichter: sounds fine

nwbinspector/checks/tables.py
bendichter: By sparse do you mean short?

Either approach is fine with me, either using `all` or `np.all`.

CodyCBakerPhD: I guess 'sparsity' (% of the number of `False` items being passed into `all`) isn't as relevant - I just mean for any case where there is _at least_ one `False` element occurring somewhat early on in the input to `all` (as opposed to occurring later in the vector, as seen above), it will exit a few ~`ms` faster per check (which could add up to seconds when streaming over an entire dataset).

Anyway, I implemented the `all` logic here.

nwbinspector/checks/tables.py
CodyCBakerPhD: OK, added here - adding to the other PR now

bendichter:
The idea is if nelems = None then it reads all the data. Could you make that true for this usage mode too?

CodyCBakerPhD:
> The idea is if nelems = None then it reads all the data. Could you make that true for this usage mode too?

Done

nwbinspector/utils.py
```diff
@@ -57,6 +57,15 @@ def is_ascending_series(series: np.ndarray, nelems=None):
     return np.all(np.diff(series[:nelems]) > 0)
 
 
+def get_uniform_indexes(length: int, nelems: Optional[int] = 200):
+    """General purpose function for generating nelems indices from 0 to length-1 with approximately equal gaps."""
+    indexes = np.round(np.linspace(start=0, stop=length - 1, num=nelems)).astype(int)
+    if nelems is None or nelems >= length - 1:
+        return slice(0, None)
+    else:
+        return indexes
```
bendichter: fancy indexing is going to be expensive. Could you use slice indexing instead?

CodyCBakerPhD: Done

nwbinspector/utils.py
```diff
@@ -57,6 +57,13 @@ def is_ascending_series(series: np.ndarray, nelems=None):
     return np.all(np.diff(series[:nelems]) > 0)
 
 
+def safe_uniform_selection(length: int, nelems: Optional[int] = 200) -> slice:
+    """General purpose function for safely generating an evenly spaced slice ."""
+    if nelems is None:
+        return slice(0, None)
+    return slice(0, length, np.ceil(length / nelems).astype(int))
```
bendichter:
```suggestion
    return slice(0, None, np.ceil(length / nelems).astype(int) if nelems else None)
```

nwbinspector/checks/tables.py
```diff
@@ -188,3 +189,19 @@ def check_table_values_for_dict(table: DynamicTable, nelems: int = 200):
...
+def check_col_not_nan(table: DynamicTable, nelems: Optional[int] = 200):
+    """Check if all of the values in a single column of a table are NaN."""
+    for column in table.columns:
+        if not hasattr(column, "data") or isinstance(column, VectorIndex) or isinstance(column.data[0], str):
+            continue
+        if nelems is not None and not all(np.isnan(column[:nelems]).flatten()):
+            continue
+
+        subindex_selection = safe_uniform_selection(length=column.shape[0], nelems=nelems)
+        if all(np.isnan(column[subindex_selection])):
```
```suggestion
        if all(np.isnan(column[slice(0, None, np.ceil(length / nelems).astype(int) if nelems else None)])):
```

nwbinspector/utils.py
```diff
@@ -90,6 +90,11 @@ def is_ascending_series(series: Union[h5py.Dataset, ArrayLike], nelems=None):
         return np.all(np.diff(series[:nelems]) > 0)  # already in memory, no need to cache
 
 
+def safe_uniform_selection(length: int, nelems: Optional[int] = 200) -> slice:
+    """General purpose function for safely generating an evenly spaced slice ."""
+    return slice(0, None, np.ceil(length / nelems).astype(int) if nelems else None)
+
+
```
bendichter:
```suggestion
```

nwbinspector/checks/tables.py
```diff
@@ -14,6 +14,7 @@
     is_ascending_series,
     is_dict_in_string,
     is_string_json_loadable,
+    safe_uniform_selection,
```
```suggestion
```

bendichter:
how would you feel about moving this one line into the check instead of making it its own function?

CodyCBakerPhD:
> how would you feel about moving this one line into the check instead of making it its own function?

Yeah that's fine since it's so parred down now, the only complicated bit is the logic for determining the 'by' of the slice but that should be readable enough as is - thanks for helping make that so much simpler!

bendichter (approved):

PR #232, opened 2022-07-13
[New Check] Check if identifiers are unique over a folder path

### LINKED ISSUE #157: [Add Check]: identifier field is unique across folder

An interesting test this would be; it is certainly a Best Practice that the `identifier` be unique to each individual NWBFile, but checks in the NWBInspector are designed to be run on single NWBFiles only. Will have to think about how to do this when calling the CLI on a folder or `inspect_all` on a folder path.

```shell
assert len(set(identifiers_from_nwbfiles_in_folder)) == len(list_of_nwbfiles)
```

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #157 

A very special one-of-a-kind check that isn't actually a function in the registry but rather a manual inspection triggered only when multiple files are detected from a folder path in `inspect_all` (or the analogous CLI call).

Extracts the identifiers of each file and checks to see if they are unique - if not, inform the user of which ones are in violation (iteratively over each non-unique identifier).

### DIFF
--- a/nwbinspector/nwbinspector.py
+++ b/nwbinspector/nwbinspector.py
@@ -155,7 +155,10 @@ def configure_checks(
 @click.argument("path")
 @click.option("--modules", help="Modules to import prior to reading the file(s).")
 @click.option(
-    "--report-file-path", default=None, help="Save path for the report file.", type=click.Path(writable=True),
+    "--report-file-path",
+    default=None,
+    help="Save path for the report file.",
+    type=click.Path(writable=True),
 )
 @click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("--levels", help="Comma-separated names of InspectorMessage attributes to organize by.")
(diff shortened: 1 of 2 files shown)


nwbinspector/nwbinspector.py
```diff
@@ -370,6 +371,30 @@ def inspect_all(
...
+                        f"{natsorted([x.name for x in nwbfiles_with_identifier])}. "
+                        "The identifier of any NWBFile should be a completely unique value - "
+                        "we recommend using uuid4 to achieve this."
+                    ),
+                    importance=Importance.CRITICAL,
+                    check_function_name="check_unique_identifiers",
+                    object_type="NWBFile",
+                    object_name="root",
+                    location="/",
+                    file_path=str(nwbfiles_with_identifier[0].parent),
```
bendichter: can you use the `path` that is passed to `inspect_all` here?

CodyCBakerPhD: Yeah that should be possible. This check would never proceed to the message in the case where it's only a file path.

CodyCBakerPhD: Done

bendichter (approved):

PR #238, opened 2022-07-25
Add numpy >= 1.20 requirement

jwodder:
`nwbinspector/utils.py` imports `numpy.typing`, which was only introduced in numpy 1.20, yet the numpy version bounds in nwbinspector's dependencies permit installing earlier versions of numpy.  nwbinspector thus needs to declare a dependency on numpy >= 1.20 in order to ensure that a compatible version of numpy is present wherever nwbinspector is installed.


CodyCBakerPhD:
Done in #240 with slight alteration for dependencies on Python versions.

PR #242, opened 2022-07-27
Fix error with calling np.isnan in table column check

### LINKED ISSUE #241: [Bug]: Unclear error message from `check_col_not_nan` check
### What happened?

I ran NWBinspector on some files, and got an unclear error message from the `check_col_not_nan` check, which was raised as an error. From some detective work, we realized this is because in an Electrode Table I do have a column of all NaN (as a placeholder, I had filled the impendences with NaN). 

NWBinspector is right to flag this, but it seems to have trouble reporting this in a clear way, and prints out this long error stack, from which it is tricky to figure out which table / column raised the error. If possible, it would be nice if this error could be caught in a way to be more clear about what the issue is:

Screen Shot 2022-07-26 at 6 00 46 PM

### Steps to Reproduce

```shell
This happened from an NWB file with a column of NaNs. If you need a test file, I can make and add one.
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.8

### Usage

Command Line Interface

No

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #241

## Motivation

Recently discovered that the current state of `check_table_column_is_nan` has a bug when running on certain columns types that are not valid to be compared to `np.nan`. This provides more stringent filtering of columns based on their data types.

## How to test the behavior?

- [x] Should add a specific test for this, but wondering if it currently solves the specific problem @TomDonoghue originally had.

## Checklist


nwbinspector/checks/tables.py
```diff
@@ -194,7 +194,7 @@ def check_table_values_for_dict(table: DynamicTable, nelems: int = 200):
 def check_col_not_nan(table: DynamicTable, nelems: Optional[int] = 200):
     """Check if all of the values in a single column of a table are NaN."""
     for column in table.columns:
-        if not hasattr(column, "data") or isinstance(column, VectorIndex) or isinstance(column.data[0], str):
+        if not hasattr(column, "data") or isinstance(column, VectorIndex) or not isinstance(column[0], Real):
```
```suggestion
        if not hasattr(column, "data") or isinstance(column, VectorIndex) or column.dtype == np.dtype('float'):
```

CodyCBakerPhD: A couple things here - the original `isinstance(column[0], Real)` was allowing integer types for which `NaN` has no meaning (and if a list with np.nan is casted as a numpy array with integer dtype, it collapses to `0`), so removed that.

For the suggested `column.dtype` approach,

1. the `.dtype` attribute may not be set to the `VectorData` `data` object if the NWBFile is still in-memory (only if it is being read from disk). Minor workaround to that is just always cast the first element as an array to get the `dtype`

2. `that_dtype == np.dtype("float")` is a strict assertion to exactly `float64`, whereas any other `floatXYZ` type would have valid interpretation of `np.nan`

This is also an early skip condition, so logic for detection is negated.

`not np.issubdtype(np.array(column[0]).dtype, np.floating)` is the current approach I'm taking, which is valid for all numpy-supported float types (including `float128`).

CodyCBakerPhD:
@bendichter Anything else needed for this one?

nwbinspector/checks/tables.py
```diff
@@ -205,5 +209,5 @@ def check_col_not_nan(table: DynamicTable, nelems: Optional[int] = 200):
             ).flatten()
         ):
             yield InspectorMessage(
-                message=f"Column {column.name} has all NaN values. Consider removing it from the table."
+                message=f"Column '{column.name}' has all NaN values. Consider removing it from the table."
```
```suggestion
                message=f"Column '{column.name}' might have all NaN values. Consider removing it from the table."
```

bendichter: what do you think of this? Better yet, we could check for the length of the column. If the column is less than nelems, we can say with certainty that all values are NaNs and if not we can say all values **might** be NaNs

CodyCBakerPhD: Hmm yeah that's true - I can add something like that

CodyCBakerPhD: Added, plus a test for it

bendichter (approved):

PR #244, opened 2022-07-29
[MNT] - Update version number location

TomDonoghue:
## Motivation

This PR moves the version number into the module, and reads this from `setup.py`. 
This approach then allows for storing the version number into `__version__` with the module, as is standard. 

## How to test the behavior?

This should now work:
```
from nwbinspector import __version__
```

## Checklist


CodyCBakerPhD:
This one is actually by design to promote the more modern use of 

```python
from importlib.metadata import version as importlib_version

package_version = importlib_version(package_name)
```

which does not require a package to set that module attribute.

> This approach then allows for storing the version number into __version__ with the module, as is standard.

It's a 'quasi-standard' in that many older projects use it, but it was rejected by PEP for the same reasons above: https://peps.python.org/pep-0396/#pep-rejection for background

Also see the convenience function https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/utils.py#L128-L150 for Python 3.7, which we will probably support here until next summer (`importlib` is available only for 3.8+, unless it's in the 3.7 `__futures__` - I haven't checked).

TomDonoghue:
Ohhh! I guess I'm behind on this topic! I know it's always been a slightly messy system for how to copy around version numbers, but I did think that getting it into `__version__` was more of an official standard that it apparently is.

I'll have to get used to this hahaha - I find it a bit odd to not be able to read the version number from the module directly.

I guess this does leave Python versions < 3.8 in a bit of a lurch though - for situations without `importlib.metadata` available there's not really a supported standard for accessing the version numbers?

CodyCBakerPhD:
> I guess this does leave Python versions < 3.8 in a bit of a lurch though - for situations without importlib.metadata available there's not really a supported standard for accessing the version numbers?

The convenience function we use for 3.7 support (https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/utils.py#L128-L150) utilizes `packaging.resources` for this: https://github.com/pypa/packaging

CodyCBakerPhD:
> I guess this does leave Python versions < 3.8 in a bit of a lurch though - for situations without importlib.metadata available there's not really a supported standard for accessing the version numbers?

The convenience function (https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/nwbinspector/utils.py#L128-L150) utilizes `pkg_resources` which ships with `setuptools` to achieve this

bendichter:
Is there a downside to supporting both? I know a lot of python users still rely on `import __version__` most of the time (myself included)

CodyCBakerPhD:
> Is there a downside to supporting both? I know a lot of python users still rely on `import __version__` most of the time (myself included)

No, no downside. Just a matter of practice to not require package creators and maintainers to add such an attribute, as it is practice for package users to not have to rely on it's optional existence. Just creates more work for everyone instead of using a built-in tool (now `importlib`) for retrieving it in all situations.

I would however want to avoid having a single file of a single line just for this. @TomDonoghue isn't it possible to still set the `verion="current.version.string"` inside the `setup.py`, and do one of either

```python
from ..setup import version as __version__
```

or just a special absolute at the top of the `__init__` as

```python
from nwbinspector.version import version as __version__
```

bendichter:
We could set the string in the root `__init__.py`. This is [what sklearn does](https://github.com/scikit-learn/scikit-learn/blob/ec6efc57fac36820919184ffb5fae70a7e871b62/sklearn/__init__.py#L42)

bendichter:
@CodyCBakerPhD it looks like codecov isn't working. Do you know why that might be?

CodyCBakerPhD:
> @CodyCBakerPhD it looks like codecov isn't working. Do you know why that might be?

Specific to this PR, it's because it's from an external fork and therefore doesn't have access to the repo secrets (one of which is codecov upload token). Basically that same thing Julia pointed out on NeuroConv for the testing suite.

Two options:

a) can enable the secrets to be shared by altering the trigger condition like we experimented with on Julias PR; this does however leave open certain vulnerabilities where someone could open a PR including some code additions to the testing suite that steals the secrets,

or b) can make codecov upload optional and skip it when running on a fork (only run and upload coverage when tested directly on the repo)

bendichter:
b sounds good to me

CodyCBakerPhD:
@bendichter Well, IDK actually. I triggered a re-run and it works perfectly fine as-is so maybe it doesn't actually need any correction after all. Strange, though, since the secret is indeed used in upload step

TomDonoghue:
For declaring the version number - I agree that supporting both is a nice idea, perhaps at least until minimum supported version are >= 3.8 (such that the new approach is guaranteed to be available). 

In terms of the suggestions above:
```
from ..setup import version as __version__
```
I feel like I've never seen the module files reading from the setup file - and I tried it out of curiosity. As I sorta expected, it fails ("ValueError: attempted relative import beyond top-level package"), based on, I assume, that once the module is installed, it doesn't have the setup file available / accessible one directory up. 

or just a special absolute at the top of the `__init__` as
```
from nwbinspector.version import version as __version__
```
^Maybe I'm mis-understanding this one - this would still require a `nwbinspector/version.py` which defines `version`, which gets linked into the `__init__`, right? Assuming so, this should work, but I'm not really clear if / how it's different from this PR. 

For the `__init__` idea: my understanding of the reason having the single line version declaration file is common is that it allows for parsing this line as plain text (as done in the suggested updated in `__init__`), since, in general, you can't import the module that's being installed in the `setup.py` file. In the `sklearn` example, the have to do some funky stuff in their __init__ (https://github.com/scikit-learn/scikit-learn/blob/ec6efc57fac36820919184ffb5fae70a7e871b62/sklearn/__init__.py#L68) for this to actually work - and this isn't a strategy I would generally vote for.

If the goal is to define `__version__`, while actually defining the version number in `setup.py` and not adding a new file, then in the special case of nwbinspector which has the `get_package_version` function, this works:
```
from .utils import get_package_version
__version__ = str(get_package_version('nwbinspector'))
```
^A quirk though is that this reads an installed version number, and so if someone does something like cloning and importing from the local directory (reading locally, without installing), then this will throw an error.

Generally, in terms of how best to support both, I'm not really clear on a solid way of defining version in setup and linking it into `__version__` (other than trying to read and link an installed version number, as in the last example). 

In terms of what to choose here - I think it depends a bit where one wants to prioritize writing the version number. FWIW, from my knowledge, I would tend to think that this PR is the cleanest way that I know that support both ways of reading the version number - though it does so by adding a new / single-line file.

bendichter:
@TomDonoghue thanks for pointing out the funky stuff in the sklearn `__init__`. I didn't realy that would be needed to get that strategy to work. I also think adding a new `__version__` file would be the cleanest way to support both methods.

CodyCBakerPhD (approved):

PR #247, opened 2022-08-03
Update dandi.inspector_config.yaml

bendichter:
## Motivation

## How to test the behavior?
```
```

## Checklist


CodyCBakerPhD:
Oh wow, is it finally time to elevate these to be DANDI upload requirements? 😲

Note the test here: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/tests/test_check_configuration.py#L68-L81

will have to be updated accordingly.

nwbinspector/internal_configs/dandi.inspector_config.yaml
```diff
@@ -1,8 +1,8 @@
 CRITICAL:  # All the fields under CRITICAL will be required for dandi validate to pass
   - check_subject_exists
   - check_subject_id_exists
+  - check_subject_sex
+  - check_subject_species
```
CodyCBakerPhD, line 5: All the current discussions have focused on requiring `age` and range bounds for it.

Are we sure we want to simultaneously make sec/species required at the same time as well, or take it one at a time? Just wanted to check.

bendichter: Well to me species makes a lot more sense. Experimenters may not know age, but they at least should know species!

bendichter: So I would say yes, let's make that required.

bendichter:
The way we have it now, `check_subject_proper_age_range` is not included in the dandi config. Do we want to add it?

CodyCBakerPhD:
@bendichter Good catch: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/255

bendichter:
Let's go ahead and merge this. I'd rather be ahead of DANDI's requirements than behind them

CodyCBakerPhD:
There is one possible change we might need to make to DANDI validation before going forward with this, which is to require (at time of performing `dandi validate`) the user to have the latest PyPI release of the NWB Inspector installed.

Otherwise, someone can simply install an earlier version of the Inspector and pass the new validation requirements without issue.

Some pseudo-code

```python
import json
import urllib2
from packaging import version
from nwbinspector.utils import get_package_version

def check_current_version():
    url = "https://pypi.org/pypi/nwbinspector/json"
    data = json.load(urllib2.urlopen(urllib2.Request(url)))
    versions = data["releases"].keys()
    #  get max_version from the `versions` list
    return get_package_version(package_name="nwbinspector") == max_version
```

bendichter:
Yeah that makes sense

CodyCBakerPhD:
@bendichter OK do you think this deserves to be more on the DANDI side of things like in this dev branch: https://github.com/dandi/dandi-cli/compare/master...catalystneuro:dandi-cli:force_usage_of_latest_nwb_inspector_release

OR we should make this a check function (or just a general-purpose assertion for all core inspection functions) for the NWBInspector in and of itself (which could warn people to update ahead of attempting `dandi validate` if they are following the user guide instructions and using the NWB Inspector early in the process)? [EDIT: obviously together with a simple lower bound version bump on the DANDI side, which is simpler for them to review]

CodyCBakerPhD:
@bendichter How about this - in NWB Inspector, we always check if user is using most recent version of the NWB Inspector and include it in the report/message stream if not.

Then on DANDI, bump the minimal version of the NWB Inspector to the version that enforces that?

CodyCBakerPhD (approved):

CodyCBakerPhD:
@bendichter OK since that last part is reliant on the DANDI side (see https://github.com/dandi/dandi-cli/pull/1108) we can go ahead and merge this when you're ready

CodyCBakerPhD (approved):

PR #248, opened 2022-08-03
add check and tests for if an optional value is set as an empty string

### LINKED ISSUE #246: [Add Check]: Missing values should be `None`, not empty string

Sometimes users are missing optional values, such as the NWBFile.experimenter, and they use an empty string as a placeholder instead of just omitting this value by setting it to `None`. `None` is the better approach. We should add this to the best practices and add checks for it.

_No response_

No.

### Code of Conduct

bendichter:
## Motivation

fix #246 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/nwbinspector/checks/nwb_containers.py
+++ b/nwbinspector/checks/nwb_containers.py
@@ -73,5 +73,5 @@ def check_empty_string_for_optional_attribute(nwb_container: NWBContainer):
     for field in fields:
         yield InspectorMessage(
             message=f'The attribute "{field}" is optional and you have supplied an empty string. Improve my omitting '
-            f"this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)"
+            "this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)"
         )
--- a/nwbinspector/checks/nwb_containers.py
+++ b/nwbinspector/checks/nwb_containers.py
@@ -67,8 +67,8 @@ def check_empty_string_for_optional_attribute(nwb_container: NWBContainer):
     ----------
     nwb_container: NWBContainer
     """
-    docval = type(nwb_container).__init__.__docval__["args"]
-    optional_attrs = [_d["name"] for _d in docval if "default" in _d and _d["default"] is None]
+    docval_args = type(nwb_container).__init__.__docval__["args"]
+    optional_attrs = [arg["name"] for arg in docval_args if "default" in arg and arg["default"] is None]
     fields = [attr for attr in optional_attrs if getattr(nwb_container, attr) == ""]
     for field in fields:
         yield InspectorMessage(
--- a/nwbinspector/checks/nwb_containers.py
+++ b/nwbinspector/checks/nwb_containers.py
@@ -57,7 +57,7 @@ def check_small_dataset_compression(
             )
 
 
-@ register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBContainer)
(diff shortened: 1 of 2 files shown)


nwbinspector/checks/nwb_containers.py
```diff
@@ -73,5 +73,5 @@ def check_empty_string_for_optional_attribute(nwb_container: NWBContainer):
     for field in fields:
         yield InspectorMessage(
             message=f'The attribute "{field}" is optional and you have supplied an empty string. Improve my omitting '
-                    f'this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)'
+            f"this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)"
```
CodyCBakerPhD: Minor; This level doesn't have to be an fstring to concatenate with the one above it

nwbinspector/checks/nwb_containers.py
```diff
@@ -73,5 +73,5 @@ def check_empty_string_for_optional_attribute(nwb_container: NWBContainer):
     for field in fields:
```
CodyCBakerPhD, line 73: The logic for `optional_attrs` is a bit to unpack - is there a better variable name than `_d` for explaining what we're iterating over?

And just to confirm, `"args"` is all we need from the `__docval__` attached to the `__init__` of some class? No `"kwargs"` alternatives?

And the logic for detecting optionality is the presence of a `"default"` value, and if there is one we check that it's set to `None` instead of some pre-defined value? Just curious, how does the `None` logic equate to a file written in MATLAB? Does the attribute simply not show up attached to the instance of the class when read back in through PyNWB?

bendichter: > The logic for optional_attrs is a bit to unpack - is there a better variable name than _d for explaining what we're iterating over?

ok, I renamed `_d` to `arg`. 

> And just to confirm, "args" is all we need from the __docval__ attached to the __init__ of some class? No "kwargs" alternatives?

docval just has everything in "args". There is no "kwargs" field.

> how does the None logic equate to a file written in MATLAB 

In MatNWB, all attributes are written with key,value pairs, e.g. 

```matlab
NwbFile("session_start_time", now(), "session_description", "this is my description")
```

To omit an attribute, you just don't provide the key, value pair. Omitted values are read back as `[]`.

nwbinspector/checks/nwb_containers.py
```diff
@@ -67,11 +67,11 @@ def check_empty_string_for_optional_attribute(nwb_container: NWBContainer):
...
-    docval = type(nwb_container).__init__.__docval__["args"]
-    optional_attrs = [_d["name"] for _d in docval if "default" in _d and _d["default"] is None]
+    docval_args = type(nwb_container).__init__.__docval__["args"]
+    optional_attrs = [arg["name"] for arg in docval_args if "default" in arg and arg["default"] is None]
     fields = [attr for attr in optional_attrs if getattr(nwb_container, attr) == ""]
     for field in fields:
         yield InspectorMessage(
             message=f'The attribute "{field}" is optional and you have supplied an empty string. Improve my omitting '
-                    f'this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)'
+            f"this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)"
```
bendichter:
```suggestion
            "this attribute (in MatNWB or PyNWB) or entering as None (in PyNWB)"
```

nwbinspector/checks/nwb_containers.py
CodyCBakerPhD: Oky doky, sounds good then

CodyCBakerPhD (approved):

PR #252, opened 2022-08-16
propose changes to best practices

### LINKED ISSUE #209: [Add Check]: age Duration range and Best Practice

From discussion the other day, need to formalize a Best Practice for using the `age` of `Subject` as a free-form text field to account for some additional properties.

We already recommend ISO 8601 when exact age is known. If only a range is available, this uncertainty should be indicated.

I think what we landed on was using interval notation, *e.g.*, `age="(P4W, P6W)"` to indicate subject age could have been as low as 4 weeks and as high as 6 weeks old, but exact age unknown.

There was also the reference time of 'birth' or 'stage' of development, which can be added on using a separation character such as '-', for the form `age="P4D - <reference point>"`, *.e.g.* `age="P4D - Conception"` for *in utero* mouse studies, or `age="P1D - Adulthood"` for *Drosophila melanogaster* studies, with global default `<reference point>` being `"birth"`.

_No response_

Yes.

### Code of Conduct

bendichter:
## Motivation

fix #209

### DIFF
--- a/docs/best_practices/nwbfile_metadata.rst
+++ b/docs/best_practices/nwbfile_metadata.rst
@@ -225,7 +225,7 @@ format. For instance indicating an age of 90 days would be 'P90D'. It is not nec
 ``date_of_birth``, but at least one of them is required by the DANDI Archive and recommended in general.
 
 If the precise age is unknown, an age range can be given by "[lower bound]/[upper bound]" e.g. "P10D/P20D" would mean
-that the age is in between 10 and 20 days. If only the lower or upper bound is known, "**" can be used to indicate a
+that the age is in between 10 and 20 days. If only the lower bound is known, then including only the slash after that lower bound can be used to indicate a
 missing bound. For instance, "P90Y/" would indicate that the age is 90 years or older.
 
 Check function: :py:meth:`~nwbinspector.checks.nwbfile_metadata.check_subject_age`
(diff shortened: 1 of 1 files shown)


docs/best_practices/nwbfile_metadata.rst
```diff
@@ -222,7 +222,11 @@ Age
 
 The ``age`` of a :nwb-schema:ref:`sec-Subject` should use the :wikipedia:`ISO 8601 Duration <ISO_8601#Durations>`
 format. For instance indicating an age of 90 days would be 'P90D'. It is not necessary to include both ``age`` and
-``date_of_birth``, but at least one of them is recommended.
+``date_of_birth``, but at least one of them is required by the DANDI Archive and recommended in general.
+
+If the precise age is unknown, an age range can be given by "[lower bound]/[upper bound]" e.g. "P10D/P20D" would mean
+that the age is in between 10 and 20 days. If only the lower or upper bound is known, "**" can be used to indicate a
```
CodyCBakerPhD:
```suggestion
that the age is in between 10 and 20 days. If only the lower bound is known, then including only the slash after that lower bound can be used to indicate a
```

With the 'or' there interpreting as the usual inclusive form, I do not like the open possibility of having the age field be `**/**`, unless that is intended to have some interaction with the upcoming `age.reference` field?

Otherwise, I think it's fair to make the lower bound required in our manually defined syntax here, with only the upper bound being optional with the `**`.

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -222,7 +222,11 @@ Age
...
+missing bound. For instance, "P90Y/**" would indicate that the age is 90 years or older.
```
CodyCBakerPhD:
```suggestion
missing bound. For instance, "P90Y/" would indicate that the age is 90 years or older.
```

PR #261, opened 2022-09-06
Add environmental control over network-dependent tests

### LINKED ISSUE #260: [Feature]: option to exclude tests which require network connectivity

I get a lot of errors of the form:

```
E   OSError: Unable to open file (curl cannot perform request)
```

When trying to run the test suite of versions `0.4.{3..12}` this being the full build log for the latter: https://ppb.chymera.eu/ec7618.log

This could be done by either having an option which just turns network-reliant tests off (e.g. depending on an environment variable [as seen in DANDI](https://github.com/dandi/dandi-cli/blob/17e0b04263ed48b3c2f5946b20daded49bef0dfa/dandi/tests/skip.py#L131)), or by releasing static test data which can be fetched explicitly instead of implicitly (which is the reason why tests often are and generally should be network-sandboxed). This latter option is used by ITK (not the additional release links):

decohost_2022-09-06T15:05:49

The band-aid solution is [to simply disable the affected tests manually `pytest ... --ignore tests/test_inspector.py --ignore tests/unit_tests`](https://ppb.chymera.eu/1b2995.log), but this is likely unsustainable as things change in future releases, so it would be cool to have something more reliable.

No.

### Code of Conduct

CodyCBakerPhD:
fix #260

@TheChymera, does this meet your request?

### DIFF
--- a/docs/developer_guide.rst
+++ b/docs/developer_guide.rst
@@ -35,3 +35,17 @@ Then, all that is needed for this to be automatically included when you run the 
 the modules flag ``-m`` or ``--modules`` along with the name of your module that contains the custom check. If using
 the library instead, you need only import the ``available_checks`` global variable from your own submodules, or
 otherwise import your check functions after importing the ``nwbinspector`` in your ``__init__.py``.
+
+
+Disable Tests That Require Network Connection
+---------------------------------------------
+
+Some of the tests in the suite require internet connectivity both to and from the DANDI archive S3 bucket.
+If this is failing for some reason, you can explicitly control all related tests by setting the environment variable
+``DANDI_TESTS_NONETWORK`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
+a linux system, run
+
+.. code-block::
+    export DANDI_TESTS_NONETWORK=1
+    
+in your environment before running ``pytest``.
--- a/docs/developer_guide.rst
+++ b/docs/developer_guide.rst
@@ -42,10 +42,10 @@ Disable Tests That Require Network Connection
 
 Some of the tests in the suite require internet connectivity both to and from the DANDI archive S3 bucket.
 If this is failing for some reason, you can explicitly control all related tests by setting the environment variable
-``DANDI_TESTS_NONETWORK`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
+``NWBI_SKIP_NETWORK_TESTS`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
 a linux system, run
 
 .. code-block::
-    export DANDI_TESTS_NONETWORK=1
+    export NWBI_SKIP_NETWORK_TESTS=1
 
 in your environment before running ``pytest``.
--- a/docs/developer_guide.rst
+++ b/docs/developer_guide.rst
@@ -47,5 +47,5 @@ a linux system, run
 
 .. code-block::
     export DANDI_TESTS_NONETWORK=1
-    
+
 in your environment before running ``pytest``.
--- a/tests/test_inspector.py
+++ b/tests/test_inspector.py
@@ -5,7 +5,6 @@ from tempfile import mkdtemp
 from pathlib import Path
 from unittest import TestCase
 from datetime import datetime
-from distutils.util import strtobool
 
 import numpy as np
 from pynwb import NWBFile, NWBHDF5IO, TimeSeries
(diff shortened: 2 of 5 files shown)


docs/developer_guide.rst
```diff
@@ -35,3 +35,17 @@ Then, all that is needed for this to be automatically included when you run the
...
+Disable Tests That Require Network Connection
+---------------------------------------------
+
+Some of the tests in the suite require internet connectivity both to and from the DANDI archive S3 bucket.
+If this is failing for some reason, you can explicitly control all related tests by setting the environment variable
+``DANDI_TESTS_NONETWORK`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
+a linux system, run
+
+.. code-block::
+    export DANDI_TESTS_NONETWORK=1
```
oruebel: I think something along the lines ``DISABLE_NETWORK_TESTS`` would be more intuitive and also make it clear that this should apply to all tests that require network access. To avoid possible interactions with other libraries and for consistency if you need to potentially add other environment variables in the future, it would be useful to pick a consistent prefix (e.g., ``NI``) for NWB Inspector environment variables, e.g,  ``NI_DISABLE_NETWORK_TESTS``

In terms of naming of the variables, it looks like the term ``NETWORK`` and ``STREAMING`` may be used interchangeably, i.e., ``DANDI_TEST_NONETWORK`` seems to refer to ``DISABLE_STREAMING_TESTS``. Just for consistency I would pick one term throughout.

tests/test_inspector.py
```diff
@@ -27,19 +28,22 @@
...
+        with NWBHDF5IO(
+            path="https://dandiarchive.s3.amazonaws.com/blobs/11e/c89/11ec8933-1456-4942-922b-94e5878bb991",
+            mode="r",
+            load_namespaces=True,
+            driver="ros3",
+        ) as io:
+            nwbfile = io.read()
+        HAVE_ROS3 = True
+    except ValueError:  # ValueError: h5py was built without ROS3 support, can't use ros3 driver
+        HAVE_ROS3 = False
```
oruebel: Rather than reading a file here, you could just check whether internet access is available and whether the ros3 driver is installed. Here is how PyNWB does this in the setup for the ROS3 tests:

```
    def setUp(self):
        # Skip ROS3 tests if internet is not available or the ROS3 driver is not installed
        try:
            urllib.request.urlopen('https://dandiarchive.s3.amazonaws.com/ros3test.nwb', timeout=1)
        except urllib.request.URLError:
            self.skipTest("Internet access to DANDI failed. Skipping all Ros3 streaming tests.")
        if 'ros3' not in h5py.registered_drivers():
            self.skipTest("ROS3 driver not installed. Skipping all Ros3 streaming tests.")
```

See https://github.com/NeurodataWithoutBorders/pynwb/blob/92ed81bcdc7f5243f08a72fd5ea10c0d5a0c5b90/tests/integration/ros3/test_ros3.py#L13-L20

tests/test_inspector.py
```diff
@@ -27,19 +28,22 @@
...
 HAVE_DANDI = is_module_installed("dandi")
+DISABLE_STREAMING_TESTS = NO_NETWORK or not HAVE_ROS3 or not HAVE_DANDI
```
oruebel: As far as I can tell only the ``DISABLE_STREAMING_TESTS`` variable is actually used by the tests. Instead of defining all the extra variables at the top of the module, I would suggest putting all the code to check whether streaming tests should be disabled into a function, e.g., ``check_streaming_tests_enabled()``  that you then call to set ``DISABLE_STREAMING_TESTS = check_streaming_tests_enabled()``. In this way you can also avoid duplicating the code below.

docs/developer_guide.rst
CodyCBakerPhD: I was thinking to save people the time/effort from having to specify multiple environment variables, but yes avoiding interactions is generally preferred, I'll change that.

`NI` might be too generic; *i.e.* conflicts with anything `^NeuroI(.+)$`. Maybe `NWBI`? Nothing else we plan on adding for NWB that begins with "I" is there?

tests/test_inspector.py
CodyCBakerPhD: Ahh, that's handy. Thanks!

tests/unit_tests/test_time_series.py
```diff
@@ -15,20 +16,25 @@
...
+            mode="r",
+            load_namespaces=True,
+            driver="ros3",
+        ) as io:
+            nwbfile = io.read()
+        HAVE_ROS3 = True
+    except ValueError:  # ValueError: h5py was built without ROS3 support, can't use ros3 driver
+        HAVE_ROS3 = False
+HAVE_DANDI = is_module_installed("dandi")
+DISABLE_STREAMING_TESTS = NO_NETWORK or not HAVE_ROS3 or not HAVE_DANDI
```
oruebel: This is duplicate code with ``test_inspector.py``. It may be worth to create a module for test utility functions either in the tests itself or the source directory of the inspector. In this way you could just import ``DISABLE_STREAMING_TESTS`` from the utility module without having to replicate the code.

oruebel (changes requested):
@CodyCBakerPhD the approach in general looks fine. The comments I added are mainly to help clean the code up a bit (e.g, to avoid code duplication)

docs/developer_guide.rst
oruebel: ``NWBI`` seems fine. I would say "first come first serve" in this case. If we ever add anything else that may want to use ``NWBI`` as a prefix then we'll have to deal with that then.

tests/unit_tests/test_time_series.py
CodyCBakerPhD: Fixed

tests/test_inspector.py
CodyCBakerPhD: Sure, generalized it into a general purpose tool for S3 detection (could be usable elsewhere...) and a testing helper.

docs/developer_guide.rst
oruebel:
```suggestion
``DANDI_TESTS_NONETWORK`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
a linux system, run

.. code-block::
    export NWBI_SKIP_NETWORK_TESTS=1
```

docs/developer_guide.rst
```diff
@@ -35,3 +35,17 @@ Then, all that is needed for this to be automatically included when you run the
...
 the library instead, you need only import the ``available_checks`` global variable from your own submodules, or
 otherwise import your check functions after importing the ``nwbinspector`` in your ``__init__.py``.
+
+
+Disable Tests That Require Network Connection
+---------------------------------------------
+
+Some of the tests in the suite require internet connectivity both to and from the DANDI archive S3 bucket.
+If this is failing for some reason, you can explicitly control all related tests by setting the environment variable
+``DANDI_TESTS_NONETWORK`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
```
oruebel:
```suggestion
``NWBI_SKIP_NETWORK_TESTS`` to some value able to be parsed by ``distutils.util.str2tool``. For example, to disable them on
```

docs/developer_guide.rst
CodyCBakerPhD: Oops. Fixing documentation now, totally forgot

oruebel (approved):
Looks good to me

PR #262, opened 2022-09-15
[Bug]: Fix for reading files written using extensions

CodyCBakerPhD:
Small but important bug fix for the identifier check added in #157 - forgot to include `load_namespaces=True` in the `NWBHDF5IO` pre-call that applies before deploying the full inspection on each file to collect all the folder-level identifiers.


oruebel (approved):

CodyCBakerPhD:
For provenance, bug discovered and communicated by @h-mayorquin on Slack. Also related to a possible mis-documentation/communication of the `--modules` flag, but that's a separate issue we'll try to improve as well.

Issue #263, opened 2022-09-16
[Bug]: Philosophy for interactions with PyNWB versions


CodyCBakerPhD:
### What happened?

Pertaining to https://github.com/conda-forge/nwbinspector-feedstock/pull/8, if someone runs the NWB Inspector tests (or the NWB Inspector itself on a file that has the relevant neurodata types) in an environment that has an older PyNWB version (for example, `2.0.0` when `2.1.0` is the latest) certain errors can occur, such as `IntracellularElectrodes` not having a `cell_id` field.

There are two solutions:

(a) require latest PyNWB when running the Inspector. Note that this does not imply that people are required to use the latest PyNWB to create those files - they could be and probably are encouraged to install and run the Inspector in an independent virtual/conda environment and I'd even be happy to compile some Docker images if that's an even more convenient way to go about it. Point being that it is the act of opening that file with the latest PyNWB version that we implicitly base the Inspector behavior off of.

(b) skip NWB Inspector checks that in some way or another depend on certain version ranges of PyNWB. Probably easiest short-term solution, but this could get complicated to track relative to schema changes and whatnot going into the far future.

As an aside I should probably have a gallery test suite that spans multiple past PyNWB versions.

### Operating System

Linux

### Python Executable

Python

### Python Version

3.9

### Usage

Library (Python code)

### Were you streaming with ROS3?

No

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/neurodatawithoutborders/nwbinspector/issues)?
- [X] To the best of your ability, have you ensured this is a bug within the code that checks the NWBFile, rather than a bug in the NWBFile reader (e.g., [PyNWB](https://github.com/NeurodataWithoutBorders/pynwb) or [MatNWB](https://github.com/NeurodataWithoutBorders/matnwb))?

CodyCBakerPhD:
From @bendichter in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/458#issuecomment-2093591598

> It looks like this is (edit: were) failing because there is a workflow that explicitly tests older versions of pynwb and this fails in that case since the feature is only supported in 2.7.0+. What should we do here? Here are some options:
>
[...]

PR #264, opened 2022-09-20
Fix attribute retrieval from version-dependent icephys

CodyCBakerPhD:
After discussing we're just going to follow KISS principles and add safe attribute access to things like this instead of getting bogged down by version specifics.

Should fix the core issue of #263 and the corresponding conda-forge issue.


CodyCBakerPhD:
@bendichter @oruebel @rly If someone could review when any of you get a chance

nwbinspector/checks/icephys.py
```diff
@@ -6,5 +6,5 @@
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=IntracellularElectrode)
 def check_intracellular_electrode_cell_id_exists(intracellular_electrode: IntracellularElectrode):
     """Check if the IntracellularElectrode contains a cell_id."""
-    if intracellular_electrode.cell_id is None:
+    if getattr(intracellular_electrode, "cell_id", "") is None:  # Will only be None with PyNWB>=2.1.0
```
bendichter: I understand that this would work, but maybe a better option is to be a bit more explicit, using `get_package_version(name="pynwb") < Version("2.1.0")` within the check function itself.

CodyCBakerPhD: I thought about that but that is precisely the precedence that leads us down a rabbit hole in the future to having to figure out which parts of which tests depend on which version of PyNWB is being used at runtime.

So it's a lot easier on us to set the precedence of just safely checking if an attribute is accessible and do an early return if not, independent of the reason that field might not be accessible.

Also, apologies - I had thought the last commit included an adjustment I made swapping it from `getattr` to `hasattr` but it didn't get pushed, my bad.

.github/workflows/version_gallery.yml
```diff
@@ -0,0 +1,32 @@
+name: Past PyNWB Version
```
bendichter, line 1: is there a reason you are setting this up as a cron job instead of just adding it to the PR triggered workflow?

nwbinspector/checks/icephys.py
bendichter: I don't really understand why you think this approach is better. It feels like this other is more explicit and accomplishes the same thing but I suppose it doesn't matter than much so we can do it this way if you really prefer it

bendichter:
Right now you are detecting the version of PyNWB. Wouldn't we instead want to check the version of the nwb schema used in the NWB file?

.github/workflows/version_gallery.yml
CodyCBakerPhD: Multiplicity - If I include it in the main workflow we'd have (number of platforms) x (number of Python versions) x (number of PyNWB versions), which gets a bit out of hand.

This instead constrains to `ubuntu-latest` + Python 3.9 and then goes back as far in the version history as desired.

CodyCBakerPhD:
> Right now you are detecting the version of PyNWB. Wouldn't we instead want to check the version of the nwb schema used in the NWB file?

The _testing suite_ detects PyNWB version to determine which type of in-memory object to create and determine the expected testing behavior. That's only important for the devs and the CI.

The check itself, what users will actually call, merely checks for existence of the attribute, which is the simplest approach we can take for this kind of thing. This is especially robust to any future changes to the NWB schema that might change the data model for that object or even minor adjustments to the name of an attribute.

CodyCBakerPhD:
> Wouldn't we instead want to check the version of the nwb schema used in the NWB file?

This is also true - are you able to provide some minimal code for retrieving this value?

bendichter (approved):

CodyCBakerPhD:
Leaving here for provenance: this PR is a simple hotfix to allow various downstream releases to go forward.

After some discussion, we need a better long-term solution for handling and communicating NWB schema version-dependent checks. The open issue #220 and PR #225 should be a good alternative for implementing this since it's analogous.

PR #273, opened 2022-09-23
Fixing all sphinx build warnings

CodyCBakerPhD:
As a first step towards a more robust documentation it should be noted that intersphinx links aren't testing alongside 'external links' and instead show up only as warnings when you try to perform a sphinx build. There are also a number of broken references within our own sections, that show up as warnings on sphinx build.

My goal to detect these going forward is to make readthedocs fail the build on warning to draw our attention to any broken links as they occur.

This PR also fixes a bunch of other things throwing warnings, and updates to latest sphinx version.


docs/best_practices/ecephys.rst
```diff
@@ -12,15 +12,15 @@ Electrodes
 Location
 ~~~~~~~~
 
-The ``location`` column of the :nwb-schema:ref:`sec-ElectrodeTable` should reflect your best estimate of the recorded
+The ``location`` column of the ``ElectrodeTable`` should reflect your best estimate of the recorded
```
CodyCBakerPhD, line 15: Big problem I'm seeing now is the old links to various 'special' tables like Electrodes and Trials have since changed or are no longer available? I've tried everything I can think of but it acts like there isn't a label for https://nwb-schema.readthedocs.io/en/latest/format.html?highlight=electrode#groups-general-extracellular-ephys-electrodes

docs/best_practices/ecephys.rst
```diff
@@ -54,22 +54,21 @@ Relative Coordinates
...
 
-
 .. _best_practice_negative_spike_times:
 
 Negative Spike Times
 ~~~~~~~~~~~~~~~~~~~~
 
 All spike times should be greater than zero. Being less than zero implies the spikes are either trial-aligned (and
-should therefore be aligned to the ``timestamps_reference_time`` of the :nwb-schema:ref:`sec-NWBFile`) or the
+should therefore be aligned to the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`) or the
```
CodyCBakerPhD, line 71: Newest version of sphinx uses slightly different syntax for intersphinx references

docs/conf_extlinks.py
```diff
@@ -1,27 +1,28 @@
 # Use this for mapping to external links
 extlinks = {
-    "pynwb-docs": ("https://pynwb.readthedocs.io/en/stable/", "%s"),
-    "matnwb-src": ("https://github.com/NeurodataWithoutBorders/matnwb", "%s"),
+    "pynwb-docs": ("https://pynwb.readthedocs.io/en/stable/%s", "%s"),
+    "matnwb-src": ("https://github.com/NeurodataWithoutBorders/matnwb/%s", "%s"),
```
CodyCBakerPhD, line 4: sphinx has warned about this for a while but now it's official - kind of annoying, but even hard-coded links require a `%s` symbol in the base URL

docs/best_practices/todo.rst
```diff
@@ -0,0 +1,116 @@
+:orphan:
+
+This is an isolated page of the documentation containing all current section references autogenerated by the check to best practice linkage of the conf.py - this page is temporary and is not to be included in any index
+
+When all of these are done, this page will be removed.
```
CodyCBakerPhD, line 5: Explanatory description for this file. Otherwise the missing references rightly throw a warnings which is now an error.

Going through these now, but some actually do represent BP sections that are currently missing and are better to add in subsequent PRs.

docs/conf.py
```diff
@@ -64,7 +63,7 @@
 
 
 def add_refs_to_docstrings(app, what, name, obj, options, lines):
-    if what == "function" and obj.__name__.startswith("check_"):
+    if what == "function" and obj.__name__.startswith("check_") and "Best Practice: " not in obj.__doc__:
```
CodyCBakerPhD, line 66: I thought I had added this a while back, but I guess not. In the case where a manual override of the BP <-> check function linkage is required due to name mismatch (or commonly, if there are multiple checks for a single best practice) this can be overridden directly in the check function docstring.

bendichter (approved):

PR #276, opened 2022-09-30
Adjust orange highlight for in-line code

CodyCBakerPhD:
Instead of that awkward orange text for in-line code


CodyCBakerPhD:
Always thought the orange code text looked a bit odd in the default RTD format since such code is rarely colored that way by most Python IDE coloring schemes. This maintains the nice box around it but allows us to choose a different color for the highlight.

Demo:

image

is now

image

bendichter (approved):

PR #277, opened 2022-10-02
Add isort to pre-commit

CodyCBakerPhD:
Trying out an extension of the pre-commit bot to handle automated import ordering

EDIT: working pretty well IMO. One less thing to have have to worry about since automation will handle it for us. Minor detail, the check registry in the main `__init__.py` has to be skipped since the order of that one global import matters.


CodyCBakerPhD:
Too many conflicts

PR #278, opened 2022-10-02
Add pydocstyle to pre-commit

CodyCBakerPhD:
curious how well a pydocstyle pre-commit does

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,4 +19,4 @@ repos:
                 docs/|
                 tests/|
                 pyproject.toml
-            )$
+            )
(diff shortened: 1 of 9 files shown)


CodyCBakerPhD:
OK it's basically like just running pydocstyle from CLI, but at least it aggregates with all the other pre-commit stuff so it doesn't have to be added as an entirely separate workflow.

Some of these, like missing docstring warnings on check functions, are actually very important because the sphinx automodule won't even detect (and also won't report a warning) a function if it doesn't have some kind of docstring.

CodyCBakerPhD:
Will deal with this later with isort

PR #279, opened 2022-10-02
Add direct parameter linkage for keyword arguments

CodyCBakerPhD:
Testing if I can get direct linkage between `:py:attr:` and an API doc for the method that uses that keyword argument.

Demonstrated on https://nwbinspector--279.org.readthedocs.build/en/279/user_guide/library_usage.html#inspectormesssage-objects

EDIT: `:py:attr:` does not work, but the extension for param links works like a charm. The in-line command is :paramref: and will lead directly to any `Parameters` section from our NumPy formatted (and Napolean parsed) docstrings.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,6 +18,7 @@ extensions = [
     "sphinx.ext.intersphinx",
     "sphinx.ext.extlinks",
     "sphinx_copybutton",
+    "sphinx_paramlinks",
 ]
 templates_path = ["_templates"]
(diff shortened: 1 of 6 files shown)


CodyCBakerPhD:
The https://pypi.org/project/sphinx-paramlinks/ works like a charm (and even plays nice with napolean)

CodyCBakerPhD:
Doesn't seem to work anymore, will relist later in time

PR #281, opened 2022-10-03
[Best Practices] Add check for `TimeSeries` unit follows `CMIXF-12` convention

### LINKED ISSUE #208: [Add Check]: units are in SI values

There is a best practice section currently on units of measurement (https://nwbinspector.readthedocs.io/en/dev/best_practices/time_series.html#units-of-measurement) for the `units` field of any `TimeSeries`.

@weiglszonja Recently found some examples worth bringing up regarding this.

(a) where they use the British spellingof the words, *e.g.*, 'metres' instead of 'meters'. We should probably decide and specify in the Best Practices whether to use American (which is what we normally suggest) or British spelling, since this can have an impact on machine readability down the line.

(b) with respect to abbreviations, we could also just enforce people to use standard SI abbreviations, *i.e*, 'meters' -> 'm'

(c) we also need to decide the official BP way of specifying division quantities, such as 'meters per second' or in combination with (b), 'm/s', which is what we've typically used ourselves.

Also it would be cool if we had a check that ensures the unit field follows SI units, but not sure what the best way to do that would be other than hard coding all the different SI base types separated from prefixes ('meters', 'second', 'volts', 'amperes', etc.)

_No response_

Yes.

### Code of Conduct

weiglszonja:
## Motivation

WIP
Proposing a new check for `TimeSeries` that checks if unit formatting complies with the CMIXF-12 convention.
#208 

## How to test the behavior?
```
```

## Checklist


docs/best_practices/best_practices_index.rst
```diff
@@ -29,3 +29,4 @@ Authors: Oliver Ruebel, Andrew Tritt, Ryan Ly, Cody Baker and Ben Dichter
    image_series
    simulated_data
    extensions
+   units
```
CodyCBakerPhD, line 32: Can go ahead and include this in the `ecephys` section

weiglszonja:
Double check what can and cannot be changed for units

nwbinspector/checks/time_series.py
```diff
@@ -92,3 +93,18 @@ def check_resolution(time_series: TimeSeries):
         return InspectorMessage(
             message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
         )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_unit_formatting(time_series: TimeSeries):
+    """Check the unit value of a TimeSeries that it complies with CMIXF-12
+    convention for formatting the units."""
```
CodyCBakerPhD:
```suggestion
    """
    Check the unit value of a TimeSeries that it complies with CMIXF-12 convention for formatting the units.
    
    Best Practice: :ref:`best_practice_units_of_measurement`
    """
```

nwbinspector/checks/time_series.py
```diff
@@ -92,3 +93,18 @@ def check_resolution(time_series: TimeSeries):
...
+
+    # Early return for arbitrary units that are unknown or unavailable.
+    if time_series.unit == "a.u.":
+        return
+
+    try:
+        parse(text="1"+time_series.unit, debug=False)
+    except ValueError:
+        return InspectorMessage(message=f"'unit' should adhere to CMIXF-12 format instead of {time_series.unit}.")
```
CodyCBakerPhD:
```suggestion
        return InspectorMessage(message=f"The 'unit' should adhere to CMIXF-12 format instead of '{time_series.unit}'.")
```

nwbinspector/checks/time_series.py
```diff
@@ -99,15 +99,17 @@ def check_resolution(time_series: TimeSeries):
...
+
     Best Practice: :ref:`best_practice_units_of_measurement`
     """
 
     # Early return for arbitrary units that are unknown or unavailable.
     if time_series.unit == "a.u.":
         return
 
+    lexer = CMIXFLexer()
+    tokens = lexer.tokenize(time_series.unit)
```
weiglszonja, line 111: `lexer.tokenize()` a `generator` object, when a bad character is found it immediately raises an error. e.g.

```
tokens = lexer.tokenize("ta")
next(tokens) 
```
would return the recognized token value for unit t
`Token(type='UNITC', value='t', lineno=1, index=0)`

next(tokens)
would return:
`{ValueError}Line 1: Bad character 'a'`

docs/best_practices/time_series.rst
```diff
@@ -37,8 +37,75 @@ measurement for that data, using the appropriate type from the
...
+     - Arbitrary Units (unknown)
+   * - :py:class:`~pynwb.image.ImageMaskSeries`
+     - "a.u."
+     - Arbitrary Units (unknown)
+   * - :py:class:`~pynwb.image.OpticalSeries`
+     - "a.u."
+     - Arbitrary Units (unknown)
+   * - :py:class:`~pynwb.retinotopy.AxisMap`
+     - "o"
+     - Degrees
```
weiglszonja, line 116: This is not `TimeSeries` but I thought could be good to include for example for representing degrees?

CodyCBakerPhD: CompassDirection is more typical for that as an example (be sure to include radians as well)

bendichter:
can we hold off on this? I don't know if we want to use CMIXF

h-mayorquin:
I am closing this. Let's open an issue to see whether this makes sense or not if someone wants to push it. Maybe at the nwb-schema level. Note that #647 implemented a simpler subset of this for the subject weight. Maybe we want to move slower here.

PR #284, opened 2022-10-03
Best Practice for not writing empty tables

CodyCBakerPhD:
Add best practice for not writing empty tables


docs/best_practices/tables.rst
```diff
@@ -6,6 +6,15 @@ degree of flexibility.
 
 
 
+.. _best_practice_empty_table:
+
+Empty Tables
+~~~~~~~~~~~~
+
+An empty :ref:`hdmf-schema:sec-dynamictable` is one that does not have any rows or columns added to it. When adding tables to an NWB file, do not write empty tables.
```
```suggestion
An empty :ref:`hdmf-schema:sec-dynamictable` is one that does not have any rows. When adding tables to an NWB file, do not write empty tables.
```

PR #285, opened 2022-10-03
Best Practice for table values (not serialized or all-NaN)

CodyCBakerPhD:
Best practice for table values, and not-NaN columns


docs/best_practices/tables.rst
```diff
@@ -5,6 +5,16 @@ The :ref:`hdmf-schema:sec-dynamictable` data type stores tabular data. It also a
 degree of flexibility.
 
 
+.. _best_practice_table_values_for_dict:
+
+Table Values
+~~~~~~~~~~~~
+
+String-valued table entries should not contain :wikipedia:`serialized representations <Serialization>`, including :wikipedia:`JSON <JSON>`. Instead, these values should be unpacked and used as additional rows or columns of that table. In the event that this might produce a :wikipedia:`jagged array <Jagged_array>`, padding with ``NaN`` values in combination with compression is preferable for user comprehension, unless this would cause conflict with :ref:`best_practice_col_not_nan` in which case that column should be removed.
```
```suggestion
String-valued table entries should not contain :wikipedia:`JSON <JSON>`. Instead, these values should be unpacked and used as additional columns of that table.
```

CodyCBakerPhD:
```suggestion
String-valued table entries should not contain :wikipedia:`JSON <JSON>`. Instead, these values should be unpacked and used as additional columns of that table.
```

PR #286, opened 2022-10-03
Best Practice for observation intervals

CodyCBakerPhD:
Best practice for observation intervals.

### DIFF
--- a/docs/best_practices/ecephys.rst
+++ b/docs/best_practices/ecephys.rst
@@ -72,3 +72,14 @@ should therefore be aligned to the ``timestamps_reference_time`` of the :ref:`nw
 ``timestamps_reference_time`` itself is not set to the earliest recording time during the session.
 
 Check function: :py:meth:`~nwbinspector.checks.ecephys.check_negative_spike_times`
+
+
+
+.. _best_practice_spike_times_not_in_unobserved_interval:
+
+Observation Intervals
+~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


docs/best_practices/ecephys.rst
```diff
@@ -72,3 +72,14 @@ should therefore be aligned to the ``timestamps_reference_time`` of the :ref:`nw
...
 Check function: :py:meth:`~nwbinspector.checks.ecephys.check_negative_spike_times`
+
+
+
+.. _best_practice_spike_times_not_in_unobserved_interval:
+
+Observation Intervals
+~~~~~~~~~~~~~~~~~~~~~
+
+The ``obs_intervals`` field of the :ref:`nwb-schema:sec-units-src` table is used to indicate periods of time where the underlying electrical signal(s) moved outside the region of view. Therefore, there should not be any identified spike times for units matched to those electrical signal(s) occuring outside of the defined ``obs_intervals``.
```
bendichter:
```suggestion
The ``obs_intervals`` field of the :ref:`nwb-schema:sec-units-src` table is used to indicate periods of time where the underlying electrical signal(s) were not observed. This can happen if the recording site moves away from the unit, or if the recording is stopped. Since the channel is not observed, it is not determinable whether a spike occurred during this time. Therefore, there should not be any identified spike times for units matched to those electrical signal(s) occurring outside of the defined ``obs_intervals``. If this variable is not set, it is assumed that all time is observed.
```

bendichter (approved):

PR #296, opened 2022-10-31
Debug lazy dataset class types and timestamps loading

CodyCBakerPhD:
@weiglszonja and I ran into some problems running the inspector on our ophys files and I tracked the issue down to a couple of inappropriate data access calls in two checks.

First problem is there's a special dataset type in HDMF called the `HDMFDataset` (https://github.com/hdmf-dev/hdmf/blob/2a42e4194853b856fe8b13a121d326b3d3d91de6/src/hdmf/query.py#L96) that apparently behaves exactly to a `h5py.Dataset` and is fully lazy (mostly b/c it seems to map API commands to an underlying `h5py.Dataset` object) but the check of `isinstance(some_dataset_object, h5py.Dataset)` fails since it's not explicitly contained in the MRO.


CodyCBakerPhD:
Not really sure how to go about adding tests for this though. Could generate a 'big' example HDMFDataset and check memory usage over time to ensure it's actually lazy?

src/nwbinspector/checks/tables.py
```diff
@@ -113,6 +113,8 @@ def check_column_binary_capability(table: DynamicTable, nelems: int = 200):
             if np.asarray(column.data[0]).itemsize == 1:
                 continue  # already boolean, int8, or uint8
             try:
+                print(column.name)
+                print(column.data.shape)
```
CodyCBakerPhD:
```suggestion
```

src/nwbinspector/nwbinspector.py
```diff
@@ -554,6 +554,7 @@ def run_checks(nwbfile: pynwb.NWBFile, checks: list):
     checks : list
     """
     for check_function in checks:
+        print(check_function)
```
CodyCBakerPhD:
```suggestion
```

src/nwbinspector/utils.py
```diff
@@ -35,7 +36,10 @@ def _cache_data_selection(data: Union[h5py.Dataset, ArrayLike], selection: Union
     """Extract the selection lazily from the data object for efficient caching (most beneficial during streaming)."""
     if isinstance(data, np.memmap):  # Technically np.memmap should be able to support this type of behavior as well
         return data[selection]  # But they aren't natively hashable either...
-    if not isinstance(data, h5py.Dataset):  # No need to attempt to cache if already an in-memory object
+    print(type(data))
+    print(isinstance(data, h5py.Dataset))
+    print(isinstance(data, H5Dataset))
```
CodyCBakerPhD:
```suggestion
```

CodyCBakerPhD:
@bendichter OK this is good for review now - just ran it on a 150 GB file without any issue (previously, it attempted to load in a very large table to memory, which massively slowed it down and could, if the table is large enough, overwhelm the RAM)

bendichter (approved):
LGTM! Thanks for pushing this through!

PR #301, opened 2022-11-09
Add false positive skip for ImageSeries check on TwoPhotonSeries 

CodyCBakerPhD:
## Motivation

Discovered a false positive on a recent run - I had a big `TwoPhotonSeries` and it triggered `check_image_series_data_size` (the one that tries to dissuade users from writing natural videos directly to the file instead of through external mode) since it is technically an `ImageSeries` child.

This adds the skip and a test for the skip.


CodyCBakerPhD:
Looking to include this in next Inspector release if possible

src/nwbinspector/checks/image_series.py
```diff
@@ -56,6 +57,11 @@ def check_image_series_data_size(image_series: ImageSeries, gb_lower_bound: floa
 
     Best Practice: :ref:`best_practice_use_external_mode`
     """
+    # False positive case; TwoPhotonSeries are a subclass of ImageSeries, but it is very common and perfectly fine
+    # to write lots of data using one without an external file
+    if isinstance(image_series, TwoPhotonSeries):
+        return
+
     data = image_series.data
     data_size_gb = data.size * data.dtype.itemsize / 1e9
```
bendichter: I don't think this accounts for compression. With compression the actual dataset may be way less than 20 GB. Is there a convenient way to check this?

CodyCBakerPhD: Yes, but that would be a separate improvement to the check. This PR is simply about skipping a neurodata type that ought to be excluded (logic of check is relative only to natural videos, not imaging `.tif` stacks)

CodyCBakerPhD: Done in #311

bendichter (approved):

Issue #216, opened 2022-07-01
[Bug]: names not working


bendichter:
### What happened?

```
MacBook-Pro-5:dev bendichter$ nwbinspector --config dandi /Users/bendichter/Downloads/09_35_09.nwb
                                                                                                                            

**************************************************
NWBInspector Report Summary

Timestamp: 2022-07-01 09:15:52.790617-04:00
Platform: macOS-10.16-x86_64-i386-64bit
NWBInspector version: 0.4.6

Found 8 issues over 1 files:
       8 - BEST_PRACTICE_SUGGESTION
**************************************************


0  BEST_PRACTICE_SUGGESTION
===========================

0.0  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Description is missing.

0.1  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Description ('no description') is a placeholder.

0.2  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: data is not compressed. Consider enabling compression when writing a dataset.

0.3  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Experimenter is missing.

0.4  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Experiment description is missing.

0.5  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Metadata /general/institution is missing.

0.6  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Metadata /general/keywords is missing.

0.7  /Users/bendichter/Downloads/09_35_09.nwb: None - 'None' object with name 'None'
       Message: Subject age, '24-48 hrs', does not follow ISO 8601 duration format, e.g. 'P2Y' for 2 years or 'P23W' for 23 weeks.
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.9

### Usage

Command Line Interface

### Were you streaming with ROS3?

No

### Package Versions

attrs==21.4.0
certifi @ file:///private/var/folders/sy/f16zz6x50xz3113nwtb9bvq00000gp/T/abs_83242e7e-f82d-4a71-8ef2-9d71d212d249gu_wxmeq/croots/recipe/certifi_1655968827803/work/certifi
[...]

PR #303, opened 2022-11-09
[Bug fix]: hotfix configuration importance

### LINKED ISSUE #302: [Bug]: nwbinspector with dandi flag doesn't contain dandi validation errors
### What happened?

I have an error from `dandi validator`, but no critical file from `nwbinspector` (file from the [user guide](https://github.com/dandi/example-notebooks/blob/master/dandi/DANDI%20User%20Guide%2C%20Part%20I.ipynb))

follow-up to https://github.com/dandi/dandi-cli/issues/1031

### Steps to Reproduce

```shell
Was not able to upload files from the [user guide](https://github.com/dandi/example-notebooks/blob/master/dandi/DANDI%20User%20Guide%2C%20Part%20I.ipynb), so run validators that gave two different things:


(dandi_pip) dhcp-10-31-85-60:205769 dorota$ dandi validate
2022-11-09 15:30:04,586 [   ERROR] sub-001/sub-001_ecephys.nwb: 1 error(s)
2022-11-09 15:30:04,586 [   ERROR]   Subject.sex is missing.
Summary: Validation errors in 1 out of 1 files
2022-11-09 15:30:04,586 [    INFO] Logs saved in /Users/dorota/Library/Logs/dandi-cli/20221109203003Z-76498.log
(dandi_pip) dhcp-10-31-85-60:205769 dorota$ nwbinspector . --config dandi
                                                                                                                                                   

**************************************************
NWBInspector Report Summary

Timestamp: 2022-11-09 15:30:11.718071-05:00
Platform: macOS-12.0.1-arm64-arm-64bit
NWBInspector version: 0.4.17

Found 7 issues over 1 files:
       1 - BEST_PRACTICE_VIOLATION
       6 - BEST_PRACTICE_SUGGESTION
**************************************************


0  BEST_PRACTICE_VIOLATION
==========================

0.0  sub-001/sub-001_ecephys.nwb: check_regular_timestamps - 'ElectricalSeries' object at location '/acquisition/test_ephys_data'
       Message: TimeSeries appears to have a constant sampling rate. Consider specifying starting_time=0.0 and rate=0.1 instead of timestamps.

1  BEST_PRACTICE_SUGGESTION
===========================

1.1  sub-001/sub-001_ecephys.nwb: check_description - 'Subject' object at location '/general/subject'
       Message: Description is missing.

1.2  sub-001/sub-001_ecephys.nwb: check_description - 'Device' object at location '/general/devices/trodes_rig123'
       Message: Description is missing.

1.3  sub-001/sub-001_ecephys.nwb: check_experimenter_form - 'NWBFile' object at location '/'
       Message: The name of experimenter 'Dr. Bilbo Baggins' does not match any of the accepted DANDI forms: 'LastName, Firstname', 'LastName, FirstName MiddleInitial.' or 'LastName, FirstName, MiddleName'.

1.4  sub-001/sub-001_ecephys.nwb: check_keywords - 'NWBFile' object at location '/'
       Message: Metadata /general/keywords is missing.

1.5  sub-001/sub-001_ecephys.nwb: check_subject_sex - 'Subject' object at location '/general/subject'
       Message: Subject.sex is missing.

1.6  sub-001/sub-001_ecephys.nwb: check_col_not_nan - 'DynamicTable' object with name 'electrodes'
       Message: Column 'imp' has all NaN values. Consider removing it from the table.
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Python

### Python Version

3.8

### Usage

Command Line Interface

No

### Package Versions

you will have the same issue when you try to run in dandi hub so it is not platform specific

### Code of Conduct

CodyCBakerPhD:
fix #302 

@djarecka and/or @yarikoptic (remember the `--config dandi`) If you can give this branch a try when you get a chance to ensure it works as expected

## Motivation

From the previous attempted fix of https://github.com/NeurodataWithoutBorders/nwbinspector/pull/218 from the state of https://github.com/NeurodataWithoutBorders/nwbinspector/pull/96 in response to #216, there was a subtle issue in which registered checks would indeed have their copied attributes correctly modified by the specified configuration; alas, the complexity of the nested wrappers in the registration (https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/register_checks.py#L121-L137) actually caused the non-configured importance level to backslide to the default value by wrapping only the original check (with original attributes) and refusing my attempts thus far to update these nested fields in the wrapper dynamically.

Regrettably, I had introduced tests for exactly this in #218 however I mistakenly set the ground truth values to the defaults instead of the expected `dandi` configuration. I've appended comments to the importance values to mentally ensure the test is now correct at a glance.

This PR introduces a hotfix to the problem which will at least restore the correct behavior, along with some other minor syntax updates I caught along the way.

Current output for me with this hotfix matches the expectation of

```shell
> nwbinspector some_file.nwb --config dandi


**************************************************
NWBInspector Report Summary

Timestamp: 2022-11-09 18:28:12.760474-05:00
Platform: Windows-10-10.0.19045-SP0
NWBInspector version: 0.4.17

Found 4 issues over 1 files:
       1 - CRITICAL
       3 - BEST_PRACTICE_SUGGESTION
**************************************************


0  CRITICAL
===========

0.0  some_file.nwb: check_subject_exists - 'NWBFile' object at location '/'
       Message: Subject is missing.

1  BEST_PRACTICE_SUGGESTION
===========================
...
```

I'll try a bit more to get a more proper fix, or otherwise rethink this approach to be less annoyingly magical (and less complex) to avoid things like this going forward.

### DIFF
--- a/src/nwbinspector/nwbinspector.py
+++ b/src/nwbinspector/nwbinspector.py
@@ -1,12 +1,10 @@
 """Primary functions for inspecting NWBFiles."""
-import os
 import re
 import importlib
 import traceback
 import json
 import jsonschema
 from pathlib import Path
-from collections.abc import Iterable
 from enum import Enum
 from typing import Union, Optional, List
 from concurrent.futures import ProcessPoolExecutor, as_completed
@@ -70,7 +68,7 @@ def _copy_function(function):
     return copied_function
 
 
-def copy_check(function):
+def copy_check(check):
     """
     Copy a check function so that internal attributes can be adjusted without changing the original function.
 
@@ -82,11 +80,9 @@ def copy_check(function):
     Taken from
     https://stackoverflow.com/questions/6527633/how-can-i-make-a-deepcopy-of-a-function-in-python/30714299#30714299
     """
-    if getattr(function, "__wrapped__", False):
-        check_function = function.__wrapped__
-    copied_function = _copy_function(function)
-    copied_function.__wrapped__ = _copy_function(check_function)
-    return copied_function
+    copied_check = _copy_function(function=check)
+    copied_check.__wrapped__ = _copy_function(function=check.__wrapped__)
+    return copied_check
 
 
 def load_config(filepath_or_keyword: PathType) -> dict:
@@ -150,13 +146,19 @@ def configure_checks(
         checks_out = []
         ignore = ignore or []
         for check in checks:
-            mapped_check = copy_check(check)
+            mapped_check = copy_check(check=check)
             for importance_name, func_names in config.items():
                 if check.__name__ in func_names:
                     if importance_name == "SKIP":
                         ignore.append(check.__name__)
                         continue
                     mapped_check.importance = Importance[importance_name]
+                    # Output wrappers are apparently parsed at time of wrapping not of time of output return...
+                    # Attempting to re-wrap the copied function if the importance level is being adjusted...
+                    # From https://github.com/NeurodataWithoutBorders/nwbinspector/issues/302
+                    # new_check_wrapper = _copy_function(function=mapped_check.__wrapped__)
+                    # new_check_wrapper.importance = Importance[importance_name]
+                    # mapped_check.__wrapped__ = new_check_wrapper
             checks_out.append(mapped_check)
     else:
         checks_out = checks
@@ -178,11 +180,11 @@ def configure_checks(
     help="Save path for the report file.",
     type=click.Path(writable=True),
 )
-@click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("--levels", help="Comma-separated names of InspectorMessage attributes to organize by.")
 @click.option(
     "--reverse", help="Comma-separated booleans corresponding to reversing the order for each value of 'levels'."
 )
+@click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("--ignore", help="Comma-separated names of checks to skip.")
 @click.option("--select", help="Comma-separated names of checks to run.")
 @click.option(
@@ -284,12 +286,10 @@ def inspect_all_cli(
         with open(file=json_file_path, mode="w") as fp:
             json_report = dict(header=get_report_header(), messages=messages)
             json.dump(obj=json_report, fp=fp, cls=InspectorOutputJSONEncoder)
-            print(f"{os.linesep*2}Report saved to {str(Path(json_file_path).absolute())}!{os.linesep}")
     formatted_messages = format_messages(messages=messages, levels=levels, reverse=reverse, detailed=detailed)
     print_to_console(formatted_messages=formatted_messages)
     if report_file_path is not None:
         save_report(report_file_path=report_file_path, formatted_messages=formatted_messages, overwrite=overwrite)
-        print(f"{os.linesep*2}Report saved to {str(Path(report_file_path).absolute())}!{os.linesep}")
 
 
 def inspect_all(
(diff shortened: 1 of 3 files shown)


src/nwbinspector/nwbinspector.py
```diff
@@ -82,11 +81,9 @@ def copy_check(function):
...
     https://stackoverflow.com/questions/6527633/how-can-i-make-a-deepcopy-of-a-function-in-python/30714299#30714299
     """
-    if getattr(function, "__wrapped__", False):
-        check_function = function.__wrapped__
-    copied_function = _copy_function(function)
-    copied_function.__wrapped__ = _copy_function(check_function)
-    return copied_function
+    copied_check = _copy_function(function=check)
+    copied_check.__wrapped__ = _copy_function(function=check.__wrapped__)
+    return copied_check
```
CodyCBakerPhD, line 86: Minor syntax change, the function never applied to something not wrapped

src/nwbinspector/nwbinspector.py
```diff
@@ -150,13 +147,19 @@ def configure_checks(
...
                     if importance_name == "SKIP":
                         ignore.append(check.__name__)
                         continue
                     mapped_check.importance = Importance[importance_name]
+                    # Output wrappers are apparently parsed at time of wrapping not of time of output return...
+                    # Attempting to re-wrap the copied function if the importance level is being adjusted...
+                    # From https://github.com/NeurodataWithoutBorders/nwbinspector/issues/302
+                    # new_check_wrapper = _copy_function(function=mapped_check.__wrapped__)
+                    # new_check_wrapper.importance = Importance[importance_name]
+                    # mapped_check.__wrapped__ = new_check_wrapper
```
CodyCBakerPhD, line 162: Leaving this in place for the follow-up where I'll either get it working or remove it in lieu of something simpler

src/nwbinspector/nwbinspector.py
```diff
@@ -178,11 +181,11 @@ def configure_checks(
     help="Save path for the report file.",
     type=click.Path(writable=True),
 )
-@click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
 @click.option("--levels", help="Comma-separated names of InspectorMessage attributes to organize by.")
 @click.option(
     "--reverse", help="Comma-separated booleans corresponding to reversing the order for each value of 'levels'."
 )
+@click.option("--overwrite", help="Overwrite an existing report file at the location.", is_flag=True)
```
CodyCBakerPhD, line 188: Noticed these were out of order compared to their keyword argument ordering. Doesn't change anything since they're matched by keyword, but still an aesthetic/readability thing

src/nwbinspector/register_checks.py
```diff
@@ -120,20 +120,20 @@ def register_check_and_auto_parse(check_function):
...
             if args:
                 obj = args[0]
             else:
                 obj = kwargs[list(kwargs)[0]]
             output = check_function(*args, **kwargs)
-            if isinstance(output, Iterable):
+            auto_parsed_result = None
+            if isinstance(output, InspectorMessage):
+                auto_parsed_result = auto_parse(check_function=check_function, obj=obj, result=output)
+            elif output is not None:
```
CodyCBakerPhD, line 131: Now this is an interesting thing I just discovered. With both `typing.Iterable` and `collections.abc.Iterable`, the claim `isinstance(list, Iterable)` fails. Not a huge issue since the `output` here is always a raw generator anyway (which appears to work fine) but it sends the wrong impression that we are truly checking if the returned value can be iterated, which we are apparently not.

I played and read around with detecting iterability and the best solution appear to really just have a helper function like

```python
def is_iterable(obj):
    try:
        iter(obj)
        return True
    except TypeError:
        return False
```

though it's likely that in most of our cases, a less obtrusive `hasattr(obj, "__iter__")` should also suffice...

Anyway that all got a bit messy so I decided the simplest case is actually to reverse the order of logic.

Test first if the output is an `InspectorMessage`, if not then check if it's `None`, if not then you know it has to be the iterable of messages.

djarecka:
@CodyCBakerPhD - yes, i get a `CRITICAL` issue now!

oruebel (approved):
Looks good to me

PR #306, opened 2022-11-21
Add downstream DANDI CI

CodyCBakerPhD:
## Motivation

From https://github.com/dandi/dandi-cli/pull/1162 and discussion at SFN with @yarikoptic, adding a workflow to the Inspector to make sure we don't accidently break DANDI validation whenever we make changes here.


CodyCBakerPhD:
@TheChymera OK as of https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/3526132039/jobs/5913701521 I have fixed the issue with the iterator error (unintended result of hotfix to #303, surprised it didn't raise any errors here; added some tests to detect that in the future) that should allow you to downgrade all the way to `BEST_PRACTICE_SUGGESTION` with relative safety. As you can see there are a number of extra messages to add to the expected results (including one 'error' that is actually just a `DeprecationWarning`, hadn't seen that one before but probably fine to expect that behavior).

One thing I will caution when building tests for this though - we've been hesitant to add a full 'integration' test that runs the entire check registry against a single file because then when we need to write a new check function for something, it multiplies the amount of effort involved. Our usual process is to write only unit tests for a check and limit the integration tests to a small handful of established checks from each importance level.

Most important to you and the rest of the DANDI team, it also means that whenever we add a new check function over here, if it would trigger on any of your fixtures in your testing suite, that means those tests (if made against an exact content of returned messages) will fail simply because we added a new level of ground truth. So there's also the extra effort of updating the testing suite across repositories in that case.

I'm not really sure what the best way to about resolving this is; for our testing suite, as I said we limit the number of checks used in integration tests on a single file (via `checks=some_list_of_checks`) but there is no injection point for this argument in the `dandi validate` usage (it always uses the entire registry by default; with configuration, of course).

Maybe the best approach for your testing suite would be a lower-bound assertion (that the returned test results contain *some* of the expected ground truth results, but may contain more) instead of exact equality?

TheChymera:
> Most important to you and the rest of the DANDI team, it also means that whenever we add a new check function over here, if it would trigger on any of your fixtures in your testing suite, that means those tests (if made against an exact content of returned messages) will fail simply because we added a new level of ground truth. So there's also the extra effort of updating the testing suite across repositories in that case.

We don't really check for the full string, just the name of the error, e.g. https://github.com/dandi/dandi-cli/blob/master/dandi/cli/tests/test_cmd_validate.py#L53 Can that be expected to be moderately stable in the long term?

TheChymera:
Another thing you mentioned at some other point, was that there will be deprecation warnings for falling behind the newest release... is that true? if so could that be disabled? since there are NWBI releases already, it makes sense for them to be usable in the longer term with update at the leisure of the user, rather than always chasing the newest one.

CodyCBakerPhD:
> We don't really check for the full string, just the name of the error, e.g. https://github.com/dandi/dandi-cli/blob/master/dandi/cli/tests/test_cmd_validate.py#L53 Can that be expected to be moderately stable in the long term?

Yes, that's safe and scalable!


> Another thing you mentioned at some other point, was that there will be deprecation warnings for falling behind the newest release... is that true? if so could that be disabled? since there are NWBI releases already, it makes sense for them to be usable in the longer term with update at the leisure of the user, rather than always chasing the newest one.

`dandi validate` and `dandi upload` (most important to the latter) already require the latest version of the Inspector to be used whenever an active internet connection is available: https://github.com/dandi/dandi-cli/blob/master/dandi/files/bases.py#L694-L708

This is very important because otherwise someone could forcibly downgrade their NWB Inspector version in order to bypass whatever the most up-to-date DANDI metadata requirements are. Also, Yarik didn't want me having to update the lower bound of the Inspector in the DANDI requirements every time that's needed (which would also need a specific DANDI release just for that).

CodyCBakerPhD:
> As you can see there are a number of extra messages to add to the expected results (including one 'error' that is actually just a DeprecationWarning, hadn't seen that one before but probably fine to expect that behavior).

Or if this is what you mean, it's a relatively harmless warning from PyNWB that only makes sense in certain cases (probably irrelevant to the test files), so I'm happy to add a filter to it for the time being. Though I'm not entirely sure why it's getting raised in that particular check. If it's a problem (which, I guess it would technically prevent upload) I can look into more detail if it's reproducible

CodyCBakerPhD:
@oruebel Another follow-up of #303 fixing an issue caused from that re-ordering of the `if` statement related to detection of iterable-or-not `output` as well as a bunch of tests for detecting it in the future

Also added a CI for downstream testing against DANDI to make sure we don't accidentally break their CI again

TheChymera:
> This is very important because otherwise someone could forcibly downgrade their NWB Inspector version in order to bypass whatever the most up-to-date DANDI metadata requirements are.

Well, that's a logger warning, not a validation error (and a paper-thin defense against bypassing requirements :D ). I thought NWBI itself will complain if it's not the newest version, I know it did that in the past, but then the behaviour went away and I was very happy :3 Is NWBI trying to continue to do this, or did I misunderstand that?

CodyCBakerPhD:
> and a paper-thin defense against bypassing requirements :D

You mean because people can force upload without validation anyway? That's highly discouraged


> I thought NWBI itself will complain if it's not the newest version

We never had anything like that on our side - we had discussed it around the same time as the DANDI-level thing was added but decided people can run whatever version of the inspector they want directly, but for dandi purposes they need latest.

The DANDI-CLI also requires recent versions to perform operations relative to the server version at time of upload, otherwise, again, an entire myriad of issues from the past might re-occur just due to an outdated client version, no?

bendichter (approved):

PR #315, opened 2022-11-30
fix get_data_shape for unbounded dataset

bendichter:
* create a custom get_data_shape that does not return maxshape
* write test that has an unbounded dataset

### DIFF
--- a/tests/unit_tests/test_time_series.py
+++ b/tests/unit_tests/test_time_series.py
@@ -3,6 +3,8 @@ import pynwb
 import pytest
 from packaging import version
 
+import h5py
+
 from nwbinspector import (
     InspectorMessage,
     Importance,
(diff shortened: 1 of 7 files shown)


tests/unit_tests/test_time_series.py
```diff
@@ -77,6 +79,25 @@ def test_check_data_orientation():
...
 
 
+def test_check_data_orientation_unbounded_maxshape(tmp_path):
+    filepath = tmp_path / "test.nwb"
+    with h5py.File(filepath, "w") as file:
+        data = file.create_dataset(
+            "data",
+            data=np.ones((10, 3)),
+            maxshape=(None,3),
+        )
```
CodyCBakerPhD: Recommend using proper `TestCase` `setUp`/`tearDown` structure for cleaning up the temporary file

bendichter: the tmp_path fixture should take care of that

CodyCBakerPhD:
Re-triggering tests after pre-commit bot, something strange happened

CodyCBakerPhD (approved):

PR #318, opened 2022-12-02
remove the optional module check for pynwb.base.images

bendichter:
## Motivation

This module check looks unnecessary


CodyCBakerPhD:
Note the CI passes because the tests related to use of those checks are also skipped in similar fashion: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/tests/unit_tests/test_images.py#L9-L15

I think if that test skip was disabled, the Inspector would run into an import issue on any environment that uses PyNWB 2.0.0 (which we test against) - can you try disabling the test skip to verify that?

PR #321, opened 2022-12-04
Added special skip condition for check_timestamps_match

CodyCBakerPhD:
## Motivation

As per discussion.

What is the name of the file on Globus to test? Upload seems to have only finished right now for `1109699524.spikes.nwb` and that only has a bunch of `BEST_PRACTICE_SUGGESTION`'s.

Anyway I tried my best to replicate the issue and new expected behavior in the test.

@bendichter If it's easier than waiting for globus, can you try running this branch on the file you have and see if it clears up the isssue?

### DIFF
--- a/tests/unit_tests/test_time_series.py
+++ b/tests/unit_tests/test_time_series.py
@@ -138,7 +138,7 @@ def test_check_timestamps_match_first_dimension_special_skip(tmp_path):
         pynwb.image.IndexSeries(
             name="IndexSeries",
             unit="n.a.",
-            data=[0,1],
+            data=[0, 1],
             indexed_timeseries=image_series,
             timestamps=np.arange(0, num_images, 0.1),
         )
--- a/tests/unit_tests/test_time_series.py
+++ b/tests/unit_tests/test_time_series.py
@@ -1,9 +1,9 @@
-import numpy as np
+import h5py
 import pynwb
 import pytest
+import numpy as np
 from packaging import version
 
-import h5py
 
 from nwbinspector import (
     InspectorMessage,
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@bendichter OK upload finished of the `763949859.nwb` file, tested against this branch and the issue went away

 [issue_file_with_dandi_config.txt](https://github.com/NeurodataWithoutBorders/nwbinspector/files/10149474/issue_file_with_dandi_config.txt)

tests/unit_tests/test_time_series.py
```diff
@@ -98,7 +99,62 @@ def test_check_data_orientation_unbounded_maxshape(tmp_path):
...
+        unit="n.a.",
+        data=np.empty(shape=(num_images, image_width, image_height, num_channels), dtype=dtype),
+        timestamps=[],
+    )
+    nwbfile.add_acquisition(image_series)
+    nwbfile.add_acquisition(
+        pynwb.image.IndexSeries(
+            name="IndexSeries",
+            unit="n.a.",
+            data=np.empty(shape=num_images),
```
bendichter: This generates floats, but the data should be `np.int32`. Also, I think `np.empty` is cool but I'm not sure it's the best thing to use here. In the future I'd like to build a check that the values of `data` are within the range of the length of the target, which would break this. Instead, we could just have it be `[0, 1]` have have the `timestamps` also have a length of 2. Having the timestamps depend on the number of images also doesn't make much sense here.

```suggestion
            data=[0,1],
```

tests/unit_tests/test_time_series.py
```diff
@@ -98,7 +99,62 @@ def test_check_data_orientation_unbounded_maxshape(tmp_path):
...
+            indexed_timeseries=image_series,
+            timestamps=np.arange(0, num_images, 0.1),
```
bendichter:
```suggestion
            timestamps=[0.5, 0.6],
```

bendichter (approved):

PR #322, opened 2022-12-04
add check that IndexSeries is not pointed at a TimeSeries

### LINKED ISSUE #317: [Add Check]: IndexSeries should reference Images, not TimeSeries (which is deprecated)

See title.

_No response_

Maybe.

### Code of Conduct

bendichter:
## Motivation

fix #317


tests/unit_tests/test_images.py
```diff
@@ -1,10 +1,15 @@
 import pytest
 
 import numpy as np
-from pynwb.image import GrayscaleImage
+from pynwb.testing.mock.base import mock_TimeSeries
```
CodyCBakerPhD: As much as I like the idea of using mocks throughout our tests, we'll have to give it a while longer until our minimal testing version is the version of PyNWB that introduced them

src/nwbinspector/checks/images.py
```diff
@@ -29,3 +30,11 @@ def check_order_of_images_len(images: Images):
                 message=f"Length of order_of_images ({len(images.order_of_images)}) does not match the number of "
                 f"images ({len(images.images)})."
             )
+
+    @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=IndexSeries)
+    def check_index_series_points_to_image(index_series: IndexSeries):
+        if index_series.indexed_timeseries is not None:
+            return InspectorMessage(
+                message="Pointing an IndexSeries to a TimeSeries is is deprecated. Please point to an Images container "
```
CodyCBakerPhD:
```suggestion
                message="Pointing an IndexSeries to a TimeSeries is deprecated. Please point to an Images container "
```

tests/unit_tests/test_images.py
CodyCBakerPhD: Or I suppose in this special case, since we're already skipping if the `Images` data type isn't available, you could get away with doing the mock import inside the test and skip the test itself if not using PyNWB v2.2.0

tests/unit_tests/test_images.py
```diff
@@ -67,3 +72,55 @@ def test_pass_check_order_of_images_len():
...
+        indexed_timeseries=time_series,
+        unit="N/A",
+        timestamps=[0.1, 0.2, 0.3, 0.4],
+    )
+
+    assert check_index_series_points_to_image(idx_series) == InspectorMessage(
+        object_name="stimuli",
+        importance=Importance.BEST_PRACTICE_VIOLATION,
+        object_type="IndexSeries",
+        message="Pointing an IndexSeries to a TimeSeries is is deprecated. Please point to an Images container "
```
CodyCBakerPhD:
```suggestion
        message="Pointing an IndexSeries to a TimeSeries is deprecated. Please point to an Images container "
```

src/nwbinspector/checks/images.py
CodyCBakerPhD: Technically the schema says it 'will be' deprecated. https://github.com/NeurodataWithoutBorders/nwb-schema/blob/dev/core/nwb.image.yaml#L215-L216

Should we mirror that language or act as if it already has been?

CodyCBakerPhD (approved):

PR #325, opened 2022-12-07
Add inspector check for negative timestamps

### LINKED ISSUE #323: [Add Check]: check for negative time

Check first timestamp of TimeSeries
Check for `start_time` of `TimeIntervals` table

_No response_

Maybe.

### Code of Conduct

weiglszonja:
## Motivation

Add a Best Practice and an Inspector check to make sure the timestamp of the first sample in a `TimeSeries` is not negative.  When `starting_time` or the first value in the `timestamps` vector of a `TimeSeries` is negative, the Inspector will output a Best Practice Violation message.

### DIFF
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -107,23 +107,6 @@ Check function: :py:meth:`~nwbinspector.checks.time_series.check_regular_timesta
 
 
 
-
-.. _best_practice_timestamp_of_the_first_sample:
-
-Timestamp of the first sample
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-The timestamp of the first sample of a :ref:`nwb-schema:sec-TimeSeries` defined by
-:ref:`nwb-schema:TimeSeries.starting_time <sec-TimeSeries>` or the first value in
-:ref:`nwb-schema:TimeSeries.timestamps <sec-TimeSeries>` should be `0.0` or any positive value.
-If you have multiple modalities with different timestamps, the earliest timestamp should
-correspond to `0.0` and every other time reference should be shifted by this timestamp.
-
-Check function: :py:meth:`~nwbinspector.checks.time_series.check_timestamp_of_the_first_sample_is_not_negative`
-
-
-
-
 .. _best_practice_chunk_data:
 
 Chunk Data
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -107,6 +107,23 @@ Check function: :py:meth:`~nwbinspector.checks.time_series.check_regular_timesta
 
 
 
+
+.. _best_practice_timestamp_of_the_first_sample:
+
+Timestamp of the first sample
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The timestamp of the first sample of a :ref:`nwb-schema:sec-TimeSeries` defined by
+:ref:`nwb-schema:TimeSeries.starting_time <sec-TimeSeries>` or the first value in
+:ref:`nwb-schema:TimeSeries.timestamps <sec-TimeSeries>` should be `0.0` or any positive value.
+If you have multiple modalities with different timestamps, the earliest timestamp should
+correspond to `0.0` and every other time reference should be shifted by this timestamp.
+
+Check function: :py:meth:`~nwbinspector.checks.time_series.check_timestamp_of_the_first_sample_is_not_negative`
+
+
+
+
 .. _best_practice_chunk_data:
 
 Chunk Data
--- a/src/nwbinspector/checks/time_series.py
+++ b/src/nwbinspector/checks/time_series.py
@@ -99,7 +99,7 @@ def check_timestamp_of_the_first_sample_is_not_negative(time_series: TimeSeries)
     first_timestamp = time_series.starting_time if time_series.starting_time is not None else time_series.timestamps[0]
     if first_timestamp < 0:
         return InspectorMessage(
-            message=f"{time_series.name} timestamp of the first sample should not be negative value."
+            message=f"Timestamps should not be negative."
             " It is recommended to align the `session_start_time` or `timestamps_reference_time` to be the earliest time value that occurs in the data, and shift all other signals accordingly."
         )
 
--- a/src/nwbinspector/checks/time_series.py
+++ b/src/nwbinspector/checks/time_series.py
@@ -99,7 +99,7 @@ def check_timestamp_of_the_first_sample_is_not_negative(time_series: TimeSeries)
     first_timestamp = time_series.starting_time if time_series.starting_time is not None else time_series.timestamps[0]
     if first_timestamp < 0:
         return InspectorMessage(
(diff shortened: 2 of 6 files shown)


docs/best_practices/time_series.rst
```diff
@@ -107,6 +107,23 @@ Check function: :py:meth:`~nwbinspector.checks.time_series.check_regular_timesta
...
+
+.. _best_practice_timestamp_of_the_first_sample:
+
+Timestamp of the first sample
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The timestamp of the first sample of a :ref:`nwb-schema:sec-TimeSeries` defined by
+:ref:`nwb-schema:TimeSeries.starting_time <sec-TimeSeries>` or the first value in
+:ref:`nwb-schema:TimeSeries.timestamps <sec-TimeSeries>` should be `0.0` or any positive value.
+If you have multiple modalities with different timestamps, the earliest timestamp should
```
CodyCBakerPhD: I would say that this section can be condensed to a single line addition to https://github.com/catalystneuro/nwbinspector/blob/add_check_for_timestamps_are_not_negative/docs/best_practices/nwbfile_metadata.rst#global-date-and-time-reference

along the lines of "The `timestamps_reference_time` should also be the earliest timestamp in the file, giving all other time references a positive value relative to that. There should be no time references which are negative."

CodyCBakerPhD:
Can you also include the check function for any `_times` columns on any `DynamicTable` here so they merge together?

src/nwbinspector/checks/time_series.py
```diff
@@ -88,6 +88,21 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
...
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_timestamp_of_the_first_sample_is_not_negative(time_series: TimeSeries):
+    """
+    Check that the timestamp of the first sample is not negative.
+
+    Best Practice: :ref:`best_practice_timestamp_of_the_first_sample`
+    """
+
+    first_timestamp = time_series.starting_time if time_series.starting_time is not None else time_series.timestamps[0]
```
CodyCBakerPhD, line 99: Just mentally checking, is it 100% guaranteed among all the `TimeSeries` children that they have either `starting_time` or `timestamps` set to `None`? Are there any other special time references, including and not limited to `starting_frame` in `ImageSeries` (with multiple external files)? Is it also true that `starting_time` is always set `None` as an attribute and is never missing (i.e., `hasattr(time_series, "starting_time") == False`?

CodyCBakerPhD: Recommend utilizing `pynwb.testing.mock` to test it out on various subtypes just to be sure (doesn't need to be in testing suite, just want to be confident we're not missing something)

src/nwbinspector/checks/time_series.py
```diff
@@ -88,6 +88,21 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
...
+    if first_timestamp < 0:
+        return InspectorMessage(
+            message=f"{time_series.name} timestamp of the first sample should not be negative value."
```
CodyCBakerPhD:
```suggestion
            message=f"{time_series.name} timestamp of the first sample should not be negative value. It is recommended to align the `session_start_time` or `timestamps_reference_time` to be the earliest time value that occurs in the data, and shift all other signals accordingly."
```

src/nwbinspector/checks/time_series.py
```diff
@@ -88,6 +88,21 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
         return InspectorMessage(f"{time_series.name} timestamps are not ascending.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
```
CodyCBakerPhD: I'm unsure if this is serious enough to merit a `VIOLATION`. It doesn't cause any serious issues within the schema (as long as the time reference are, in truth, synchronized to be relative to the session start time) or external tools (AFAIK). It's mostly a flavor and 'expectation' thing to avoid confusion for a first time user upon seeing a negative value for time (has to think about what that means; if always positive, it's pretty clear).

I'd say maybe just a `SUGGESTION` is fine

src/nwbinspector/checks/time_series.py
bendichter: I'm pretty sure this holds across subtypes

src/nwbinspector/checks/time_series.py
```diff
@@ -93,13 +93,14 @@ def check_timestamp_of_the_first_sample_is_not_negative(time_series: TimeSeries)
...
 
-    Best Practice: :ref:`best_practice_timestamp_of_the_first_sample`
+    Best Practice: :ref:`best_practice_global_time_reference`
     """
 
     first_timestamp = time_series.starting_time if time_series.starting_time is not None else time_series.timestamps[0]
     if first_timestamp < 0:
         return InspectorMessage(
-            message=f"{time_series.name} timestamp of the first sample should not be negative value. It is recommended to align the `session_start_time` or `timestamps_reference_time` to be the earliest time value that occurs in the data, and shift all other signals accordingly."
+            message=f"{time_series.name} timestamp of the first sample should not be negative value."
```
bendichter:
```suggestion
            message=f"Timestamps should not be negative."
```

bendichter: No need to include the name of the container.

src/nwbinspector/checks/time_series.py
weiglszonja: also: [FeatureExtraction](https://pynwb.readthedocs.io/en/stable/pynwb.ecephys.html#pynwb.ecephys.FeatureExtraction), [EventDetection](https://pynwb.readthedocs.io/en/stable/pynwb.ecephys.html#pynwb.ecephys.EventDetection) both have a `times` attribute.

src/nwbinspector/checks/time_series.py
weiglszonja: Changed it to **BEST_PRACTICE_SUGGESTION**

src/nwbinspector/checks/time_series.py
CodyCBakerPhD: Oh, good find! Perhaps a separate check for those objects too then?

CodyCBakerPhD: Though perhaps in a separate PR; if those data types are deemed important enough to need it. I've never actually seen them used before.

This PR is already enough for the major cases

CodyCBakerPhD (approved):

CodyCBakerPhD:
Looks great to me! Thanks for the hard work @weiglszonja

CodyCBakerPhD:
PS: I'll wait to merge until I can retrigger and pass the dev gallery, ran into some kind of cloud issue with conda

src/nwbinspector/checks/time_series.py
weiglszonja: me neither, just checked `pynwb` API documentation for time attributes.

PR #330, opened 2022-12-16
Bump Numpy version

CodyCBakerPhD:
## Motivation

From https://github.com/dandi/dandi-cli/pull/1143#issuecomment-1355512856, trying to update requirements for `numpy`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Upcoming
 
+# v0.4.24
+
+* Loosened upper bound of numpy version. [PR # 330](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/330)
+
+
 # v0.4.23
 
 ### New Checks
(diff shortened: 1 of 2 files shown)


bendichter:
I'm fine with loosening our restriction on numpy, but I think the best way to address that issue is to test for Python 3.11.

CodyCBakerPhD:
Does PyNWB do that yet?

CodyCBakerPhD:
Also this is really just pushing the buck up to PyNWB anyway: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/setup.py#L24 

Just making it so we don't have to track them exactly here every time they bump the upper bound

CodyCBakerPhD:
Similar to https://github.com/catalystneuro/neuroconv/pull/234, awaiting word that `h5py` supports 3.11 (last check as of 2 weeks ago says no builds available: https://github.com/catalystneuro/neuroconv/actions/runs/3569661856/jobs/5999836476)

bendichter:
It looks like we should start with hdmf, which does not yet test against 3.11

https://github.com/hdmf-dev/hdmf/blob/dev/.github/workflows/run_all_tests.yml

bendichter (approved):

PR #338, opened 2023-02-02
add extension best practices

bendichter:
## Motivation

Integrate best practices from this PR: https://github.com/NeurodataWithoutBorders/nwb-overview/pull/72


docs/best_practices/extensions.rst
```diff
@@ -45,3 +133,67 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+            target_type: TimeSeries
+            reftype: object
+
+
+Best practices for object names
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Names for groups, datasets, attributes, or links should typically:
+
+* **Use lowercase letters only**
```
CodyCBakerPhD, line 182: Do we also enforce `snake_case` for these?

docs/best_practices/extensions.rst
```diff
@@ -45,3 +133,67 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+* **Use ``_`` instead of `` `` to separate parts in names**. E.g., use the name
```
CodyCBakerPhD: Ah, I see

This plus the above are essentially `snake_case`

docs/best_practices/extensions.rst
```diff
@@ -45,3 +133,67 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+  ``starting_time`` instead of ``starting time``
+* **Explicit**. E.g., avoid the use of ambiguous abbreviation in names.
+
+
+Best practices for naming ``neurodata_types``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For defining new types via ``neurodata_type_def`` use:
+
+* **Use camelcase:**  notation, i.e., names of types should NOT include spaces,
```
CodyCBakerPhD, line 193: Well, since CamelCase is specifically mentioned by name here, might as well mention for object names that they use snake case

CodyCBakerPhD:
ReadTheDocs does not like the `tabs` stuff

```
/home/docs/checkouts/readthedocs.org/user_builds/nwbinspector/checkouts/338/docs/best_practices/extensions.rst:26: ERROR: Unknown directive type "tabs".
```

from https://readthedocs.org/api/v2/build/19360519.txt

docs/best_practices/extensions.rst
bendichter, line 193:
```suggestion
* **Use CamelCase:**  notation, i.e., names of types should NOT include spaces,
```

docs/best_practices/extensions.rst
```diff
@@ -16,15 +16,104 @@ If an extension is required, tutorials for the process may be found through the
...
 When possible, use existing types when creating extensions either by creating new ``neurodata_types`` that inherit from
 existing ones, or by creating ``neurodata_types`` that contain existing ones. Building on existing types facilitates the
 reuse of existing functionality and interpretation of the data. If a community extension already exists that has a
-similar scope, it is preferable to use that extension rather than creating a new one.
+similar scope, it is preferable to use that extension rather than creating a new one. For example:
+
+* Extend ``TimeSeries`` for storing timeseries data. NWB provides main types of ``TimeSeries``
+  and you should identify the most specific type of ``TimeSeries`` relevant for your use case
+  (e.g., extend ``ElectricalSeries`` to define a new kind of electrical recording).
+* Extend ``DynamicTable`` to store tabular data/
```
CodyCBakerPhD:
```suggestion
* Extend ``DynamicTable`` to store tabular data.
```

docs/best_practices/extensions.rst
```diff
@@ -16,15 +16,104 @@ If an extension is required, tutorials for the process may be found through the
...
 
 When possible, use existing types when creating extensions either by creating new ``neurodata_types`` that inherit from
 existing ones, or by creating ``neurodata_types`` that contain existing ones. Building on existing types facilitates the
 reuse of existing functionality and interpretation of the data. If a community extension already exists that has a
-similar scope, it is preferable to use that extension rather than creating a new one.
+similar scope, it is preferable to use that extension rather than creating a new one. For example:
+
+* Extend ``TimeSeries`` for storing timeseries data. NWB provides main types of ``TimeSeries``
+  and you should identify the most specific type of ``TimeSeries`` relevant for your use case
+  (e.g., extend ``ElectricalSeries`` to define a new kind of electrical recording).
```
CodyCBakerPhD: Should these `TimeSeries`/`ElectricalSeries` be :py:class: intersphinx references?

CodyCBakerPhD: Ditto for below on `TimeIntervals` / `DynamicTable`

docs/best_practices/extensions.rst
```diff
@@ -45,3 +134,68 @@ anybody who receives the data also receives the necessary data to interpret it.
 .. note::
     In :pynwb-docs:`PyNWB <>`, the extension is cached automatically. This can be specified explicitly with
     ``io.write(filepath, cache_spec=True)``
+
+
+Use Attributes for small metadata related to a particular data object (Group or Dataset)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Attributes should be used mainly to store small metadata (usually less than 64 Kbytes) that
+is associated with a particular Group or Dataset. Typical uses of attributes are, e.g., to
```
CodyCBakerPhD: Just reading this, it's hard to grasp 'how big' something is to be less than 64 KB

Since most metadata are strings, maybe it would be useful to mention a relative scale to that?

A standard Python string under 64 KB would correspond to ~1300 characters, from 

```python
import sys

sys.getsizeof("a")
> 50

64_000 / 50
> 1280
```

Google also says the average number of characters per word in English is 4.7, so that's ~272 words per string

CodyCBakerPhD: Mostly relevant to fields like `description`, or sometimes `reference_frame`

docs/best_practices/extensions.rst
```diff
@@ -45,3 +134,68 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+define the ``unit`` of measurement of a dataset or to store a short ``description`` of
+a group or dataset. For larger data, datasets should be used instead.
+
+In practice, the main difference is that in PyNWB and MatNWB all Attributes are read into memory when reading the
```
CodyCBakerPhD:
```suggestion
In practice, the main difference is that in PyNWB and MatNWB all attributes are read into memory when reading the
```

'attributes' isn't capitalized mid-sentence in preceding paragraph

docs/best_practices/extensions.rst
bendichter: I like the idea of relating this to characters. Do you know how big characters are within HDF5? Is it the same as Python strings in memory?

CodyCBakerPhD: Hmm I'll have to investigate that a bit deeper, but most likely not the same as Python... I seem to remember us having some issues related to their byte string types here on the Inspector, but maybe that was for fields that are technically set as datasets, not attributes

Even in Python, if you apply a specific encoding it can change the size quite a bit

```python
sys.getsizeof("a".encode("utf-8"))
> 34
```
 which would put the max closer to ~400 words

CodyCBakerPhD: Well, from https://docs.h5py.org/en/stable/strings.html#storing-strings and https://docs.h5py.org/en/stable/special.html#h5py.string_dtype and manually inspecting some of these attributes with HDFView

It seems HDF5 stores with UTF-8 encoding, so about ~400 words on average from above

oruebel: > ~400 words

words or characters?

docs/best_practices/extensions.rst
oruebel: I would suggest intersphinx to the nwb-schema docs since this is for extensions

docs/best_practices/extensions.rst
CodyCBakerPhD: > words or characters?

words using Google's estimate of average characters per word in English

CodyCBakerPhD: Or just shy of ~1900 characters if that's easier to communicate

oruebel:
Some additional suggestions  inspired by BIDS https://docs.google.com/document/d/1pWmEEY-1-WuwBPNy5tDAxVJYQ9Een4hZJM06tQZg8X4/edit#heading=h.f581wbvw1fil

**Limit flexibility: Consider data reuse and tool developers**

One of the aims of NWB is to make reusing data easier. This means that when proposing an extension you need to put yourself in the shoes of someone who will receive an NWB dataset and attempt to analyze it. Additionally, consider developers that will try to write tools that take NWB datasets as inputs. It’s worth assessing how much additional code different ways of approaching your extension will lead to.

**Get the community involved**
Try to reach out to colleagues working with the type of data you are trying to add support for. The more eyes you will get on your extension the better it will get. 

**Strive for backward compatible changes**
NWB is already incorporated in many tools - proposing a change that will make already released NWB datasets non-compliant will cause a lot of confusion and will lead to significant cost to update codes.

docs/best_practices/extensions.rst
```diff
@@ -45,3 +134,68 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+* **Use camelcase:**  notation, i.e., names of types should NOT include spaces,
+  always start with an uppercase letter, and use a single capitalized letter to
+  separate parts of the name. E.g,. ``neurodata_type_def: LaserMeasurement``
+* **Use the postfix ``Series`` when extending a ``TimeSeries`` type.** E.g., when
+  creating a new ``TimeSeries`` for laser measurements then add ``Series`` to
+  the type name, e.g,. ``neurodata_type_def: LaserMeasurementSeries``
+* **Use the postfix ``Table`` when extending a ``DynamicTable`` type.** e.g.,
+  ``neurodata_type_def: LaserSettingsTable``
+* **Explicit**. E.g., avoid the use of ambiguous abbreviation in names.
+
```
oruebel, line 202:
```suggestion

Limit flexibility: Consider data reuse and tool developers
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

One of the aims of NWB is to make reusing data easier. This means that when proposing an extension you need to put yourself in the shoes of someone who will receive an NWB dataset and attempt to analyze it. Additionally, consider developers that will try to write tools that take NWB datasets as inputs. It’s worth assessing how much additional code different ways of approaching your extension will lead to.

```

docs/best_practices/extensions.rst
```diff
@@ -16,15 +16,104 @@ If an extension is required, tutorials for the process may be found through the
...
 existing ones, or by creating ``neurodata_types`` that contain existing ones. Building on existing types facilitates the
 reuse of existing functionality and interpretation of the data. If a community extension already exists that has a
-similar scope, it is preferable to use that extension rather than creating a new one.
+similar scope, it is preferable to use that extension rather than creating a new one. For example:
+
+* Extend ``TimeSeries`` for storing timeseries data. NWB provides main types of ``TimeSeries``
+  and you should identify the most specific type of ``TimeSeries`` relevant for your use case
+  (e.g., extend ``ElectricalSeries`` to define a new kind of electrical recording).
+* Extend ``DynamicTable`` to store tabular data.
+* Extend ``TimeIntervals`` to store specific annotations of intervals in time.
```
oruebel:
```suggestion
* Extend ``TimeIntervals`` to store specific annotations of intervals in time.

Strive for backward compatible changes
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

NWB is already incorporated in many tools - proposing a change that will make already released NWB datasets non-compliant will cause a lot of confusion and will lead to significant cost to update codes.

```

docs/best_practices/extensions.rst
```diff
@@ -45,3 +134,68 @@ anybody who receives the data also receives the necessary data to interpret it.
...
+  the type name, e.g,. ``neurodata_type_def: LaserMeasurementSeries``
+* **Use the postfix ``Table`` when extending a ``DynamicTable`` type.** e.g.,
+  ``neurodata_type_def: LaserSettingsTable``
+* **Explicit**. E.g., avoid the use of ambiguous abbreviation in names.
+
+Use the ``ndx-template`` to create new extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By using the :nwb_extension_git:`ndx-template` to create new extensions helps ensure
+that extensions can be easily shared and reused and published via the :ndx-catalog:`NDX Catalog <>`.
```
oruebel, line 214:
```suggestion
that extensions can be easily shared and reused and published via the :ndx-catalog:`NDX Catalog <>`.


Get the community involved
~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Try to reach out to colleagues working with the type of data you are trying to add support for. The more eyes you will get on your extension the better it will get.
```

docs/best_practices/extensions.rst
stephprince: Added a note to the text that this is approximately 1900 characters for a string attribute.

stephprince (approved):
Was going through open PRs and I think this is good to merge - made some minor changes to address the remaining open discussions.

@oruebel @bendichter do you have anything else to add or is this ok to merge?

PR #362, opened 2023-04-17
add docs page for checks by importance

### LINKED ISSUE #361: [Documentation]: show checks by importance
### What would you like changed or added to the documentation and why?

It would be useful for the DANDI handbook to be able to link to a page in the NWB Inspector docs that list checks by importance. We should have headings for each importance level and then a list of each check with that importance. @CodyCBakerPhD can you think of a way to automate that?

You can get this organization with the following code:

```python
from collections import defaultdict

dd = defaultdict(list)

for x in available_checks:
    dd[x.importance].append(x)
```

Is there a way of automatically generating a markdown or rst file from this?

### Do you have any interest in helping write or edit the documentation?

No.

### Code of Conduct

- [X] Have you ensured this change was not already [requested](https://github.com/neurodatawithoutborders/nwbinspector/issues)?

bendichter:
fix #361

### DIFF
--- a/docs/gen_checks_by_importance.py
+++ b/docs/gen_checks_by_importance.py
@@ -1,0 +1,31 @@
+import os
+from nwbinspector import available_checks
+
+from collections import defaultdict
+
+
+def gen_checks_by_importance():
+    dd = defaultdict(list)
+
+    for x in available_checks:
+        dd[x.importance.name.replace("_", " ")].append(x)
+
+    generate_checks_rst_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'checks_by_importance.rst')
+    with open(generate_checks_rst_file, 'w') as f:
+        f.write('''Checks by Importance
+======================
+
+This section lists the available checks organized by their importance level.
+
+''')
+
+        for importance_level, checks in dd.items():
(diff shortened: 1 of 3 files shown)


bendichter:
I tried adding the docstrings of each check, but the 
```
Parameters
-----------
```
section was causes problems with .rst parsing so I removed them for now. I think it would be pretty straightforward to do some text manipulation to remove that part, but we can do that later.

docs/gen_checks_by_importance.py
```diff
@@ -0,0 +1,31 @@
...
+
+This section lists the available checks organized by their importance level.
+
+''')
+
+        for importance_level, checks in dd.items():
+            f.write(f'''{importance_level}
+{'-' * (len(f'{importance_level}') + 1)}
+
+''')
```
CodyCBakerPhD, line 26: ?

bendichter: This is the title of each section

CodyCBakerPhD: Ah I see now - pre-commit must not be running on this, black usually as a different way of breaking the parentheses around triple quotes

CodyCBakerPhD (approved):

PR #363, opened 2023-04-18
inspect_nwb() accepts nwb file object

luiztauffer:
## Motivation

This should be useful for when a nwbfile object is already loaded. 
An use case example is auto inspection before uploading to dandi after a sorting+conversion workflow.

### DIFF
--- a/src/nwbinspector/nwbinspector.py
+++ b/src/nwbinspector/nwbinspector.py
@@ -515,7 +515,7 @@ def inspect_nwb(
         This sets a hard bound on the number of times to attempt to retry the collection of messages.
         Defaults to 10 (corresponds to 102.4s maximum delay on final attempt).
     """
-    if nwbfile_path is None and nwbfile_object is None:
+    if nwbfile_path is None and nwbfile_object is None or (nwbfile_path and nwbfile_object):
         raise ValueError("Either nwbfile_path or nwbfile_object must be specified.")
     importance_threshold = (
         Importance[importance_threshold] if isinstance(importance_threshold, str) else importance_threshold
--- a/src/nwbinspector/nwbinspector.py
+++ b/src/nwbinspector/nwbinspector.py
@@ -462,7 +462,8 @@ def _pickle_inspect_nwb(
 
 
 def inspect_nwb(
-    nwbfile_path: FilePathType,
+    nwbfile_path: Optional[FilePathType] = None,
+    nwbfile_object: Optional[pynwb.NWBFile] = None,
     checks: list = available_checks,
     config: dict = None,
     ignore: OptionalListOfStrings = None,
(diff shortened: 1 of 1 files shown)


src/nwbinspector/nwbinspector.py
```diff
@@ -512,6 +515,8 @@ def inspect_nwb(
         This sets a hard bound on the number of times to attempt to retry the collection of messages.
         Defaults to 10 (corresponds to 102.4s maximum delay on final attempt).
     """
+    if nwbfile_path is None and nwbfile_object is None:
+        raise ValueError("Either nwbfile_path or nwbfile_object must be specified.")
```
bendichter:
```suggestion
    if nwbfile_path is None and nwbfile_object is None or (nwbfile_path and nwbfile_object):
        raise ValueError("Either nwbfile_path or nwbfile_object must be specified.")
```

CodyCBakerPhD:
If the NWBFIle is already loaded in memory, you can just call `run_checks` on it: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L49-L50

That was the intended design, too, to reduce the complexity of any one function in the workflow.

`inspect_all_cli` (CLI wrapper of API)

-> `inspect_all` (deploys `inspect_nwb` over a folder`)

-> `inspect_nwb` (deploys checks on a file on disk, handles I/O)

-> `run_checks` (actually runs the checks on any open I/O)

Can you try giving that approach a try and make any needed improvements to `run_checks` instead of making `inspect_nwb` do more than it's intended to do?

src/nwbinspector/nwbinspector.py
```diff
@@ -523,31 +528,36 @@ def inspect_nwb(
...
 
-    with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True, driver=driver) as io:
-        if not skip_validate:
-            validation_errors = pynwb.validate(io=io)
-            for validation_error in validation_errors:
+    if nwbfile_object is not None:
+        for inspector_message in run_checks(nwbfile=nwbfile_object, checks=checks):
+            inspector_message.file_path = nwbfile_path
+            yield inspector_message
+    else:
```
CodyCBakerPhD, line 535: Can you try adjusting your usage of the Inspector to essentially just do these 3-4 lines instead of increasing the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of a particular function?

CodyCBakerPhD (changes requested):

luiztauffer:
All I want is a function to easily run the inspection on an open nwb file object, like this:

```
easy_to_inspect_function(nwbfile)
```

this will be immediately useful to our automatic conversion and publishing workflows.

`run_checks` still requires a list of checks as input, and that’s why it sounds like `inspect_nwb` should be the one to handle it, because the logic for forming the list of check objects is already built there. 

I don’t think we should push the complexity of dealing with this to the applications either, when it is already dealt with by inspect_nwb.

Sounds more reasonable to me to add a very simple logic for dealing with NWBFile objects to inspect_nwb instead.

luiztauffer:
closing it as it was fixed by https://github.com/NeurodataWithoutBorders/nwbinspector/pull/364

PR #367, opened 2023-04-19
add email notification of daily failures

CodyCBakerPhD:
The recent failures in the dailies would have been caught sooner if I had been getting my usual emails - but apparently the built-in notification feature on GitHub only applies to the person who last modified the action file in question, which was not myself

Setting up an actual email notification also allows me to add other individuals to the email list that might be interested if it starts failing

@oruebel @rly @bendichter let me know if any of you would like to be on that email list


bendichter (approved):

bendichter:
I'm fine with not being on the list for this

PR #368, opened 2023-04-20
Try to remove pre-commit inclusion rule

CodyCBakerPhD:
Curious what this will do; noticed DANDI doesn't specify it either

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,4 +9,3 @@ repos:
     rev: 23.3.0
     hooks:
     -   id: black
-        exclude: ^docs/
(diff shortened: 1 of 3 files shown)


docs/gen_checks_by_importance.py
CodyCBakerPhD, line 1: Proof that pre-commit is skipping any `.py` files under the `docs/` heading, but running `black` on it directly produces changes

.pre-commit-config.yaml
```diff
@@ -9,4 +9,3 @@ repos:
     rev: 23.3.0
     hooks:
     -   id: black
-        exclude: ^docs/
```
CodyCBakerPhD, line 12: Ahhh I'm guessing this was the problem after all

CodyCBakerPhD:
Replaced by #369

PR #374, opened 2023-06-01
Fix `is_ascending_series` to not fail with monotonic (but non-strictly monotonic)

### LINKED ISSUE #373: Should the check for ascending series be monotonic or strictly monotonic as it is?
The docstring of the utility function claims that checks for monotonic but it actually checks for strictly monotonic:

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/801d8ec97ff238b13e8a6c30a8b8f0b419b00a93/src/nwbinspector/utils.py#L90-L95

This is flagging the following data as a best practice violation:

```
0:
{'start_time': 1.0, 'stop_time': 100.0, 'label': 'WAKEstate'}
1:
{'start_time': 1.0, 'stop_time': 50.0, 'label': 'WAKEnontheta'}
2:
{'start_time': 1.0, 'stop_time': 1.0, 'label': 'WAKEtheta_ThDt'}
3:
{'start_time': 1.0, 'stop_time': 3.0, 'label': 'QWake_ThDt'}
4:
{'start_time': 1.0, 'stop_time': 3.0, 'label': 'QWake_noRipples_ThDt'}
5:
{'start_time': 3.0, 'stop_time': 4.0, 'label': 'WAKEtheta_ThDt'}
6:
{'start_time': 5.0, 'stop_time': 13.0, 'label': 'QWake_ThDt'}
7:
{'start_time': 5.0, 'stop_time': 13.0, 'label': 'QWake_noRipples_ThDt'}
8:
{'start_time': 14.0, 'stop_time': 15.0, 'label': 'WAKEtheta_ThDt'}
9:
{'start_time': 16.0, 'stop_time': 24.0, 'label': 'QWake_ThDt'}
```

h-mayorquin:
Fix #373

Added a test and did the change.

### DIFF
--- a/src/nwbinspector/utils.py
+++ b/src/nwbinspector/utils.py
@@ -90,9 +90,11 @@ def is_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
 def is_ascending_series(series: Union[h5py.Dataset, ArrayLike], nelems: Optional[int] = None):
     """General purpose function for determining if a series is monotonic increasing."""
     if isinstance(series, h5py.Dataset):
-        return np.all(np.diff(_cache_data_selection(data=series, selection=slice(nelems))) > 0)
+        differences = np.diff(_cache_data_selection(data=series, selection=slice(nelems)))
     else:
-        return np.all(np.diff(series[:nelems]) > 0)  # already in memory, no need to cache
+        differences = np.diff(series[:nelems])
+    
+    return np.all(differences >= 0)  # already in memory, no need to cache
 
 
 def is_dict_in_string(string: str):
--- a/src/nwbinspector/utils.py
+++ b/src/nwbinspector/utils.py
@@ -93,7 +93,7 @@ def is_ascending_series(series: Union[h5py.Dataset, ArrayLike], nelems: Optional
         differences = np.diff(_cache_data_selection(data=series, selection=slice(nelems)))
     else:
         differences = np.diff(series[:nelems])
-    
+
     return np.all(differences >= 0)  # already in memory, no need to cache
 
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,6 +10,7 @@ from nwbinspector.utils import (
     is_dict_in_string,
     get_package_version,
     calculate_number_of_cpu,
+    is_ascending_series,
 )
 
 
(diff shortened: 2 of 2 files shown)


src/nwbinspector/utils.py
```diff
@@ -90,9 +90,11 @@ def is_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
 def is_ascending_series(series: Union[h5py.Dataset, ArrayLike], nelems: Optional[int] = None):
     """General purpose function for determining if a series is monotonic increasing."""
     if isinstance(series, h5py.Dataset):
-        return np.all(np.diff(_cache_data_selection(data=series, selection=slice(nelems))) > 0)
+        differences = np.diff(_cache_data_selection(data=series, selection=slice(nelems)))
```
bendichter, line 93:
```suggestion
        differences = np.diff(_cache_data_selection(data=series, selection=slice(nelems)) >= 0)
```

src/nwbinspector/utils.py
```diff
@@ -90,9 +90,11 @@ def is_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
...
     else:
-        return np.all(np.diff(series[:nelems]) > 0)  # already in memory, no need to cache
+        differences = np.diff(series[:nelems])
```
bendichter, line 95:
```suggestion
        differences = np.diff(series[:nelems] >= 0)
```

src/nwbinspector/utils.py
```diff
@@ -90,9 +90,11 @@ def is_regular_series(series: np.ndarray, tolerance_decimals: int = 9):
...
+
+    return np.all(differences >= 0)  # already in memory, no need to cache
```
bendichter, line 97: `>= 0` is needed because a negative number resolves to `True`

src/nwbinspector/utils.py
h-mayorquin: I am confused. this suggestion applies np.diff to the booleans. The previous line was applying np.diff to the series and then checking for all of them positive.

src/nwbinspector/utils.py
h-mayorquin: I don't understand, what is the test that would fail:

```
def test_is_ascending_series():
    assert is_ascending_series(series=[1, 1, 1])
    assert is_ascending_series(series=[1, 2, 3])
    assert not is_ascending_series(series=[1, 2, 1])
```

I think I am missing something.

bendichter: oh sorry, I didn't read this all the way through. I see now you have moved the `>=` test later

bendichter (approved):

PR #379, opened 2023-06-25
move from ros3 to fsspec

### LINKED ISSUE #293: [Feature]: update from ros3 to fsspec

fsspec has many advantages over ros3 as a streaming strategy. We should change of our streaming functionality within nwbwidgets to use fsspec instead of ros3

No.

### Code of Conduct

bendichter:
also created a context manager `read_nwb`, which will probably migrate to PyNWB eventually

## Motivation

fix #293 

I'd like to build zarr support into the new read_nwb context manager, so I'd like to get this merged before looking at supporting the zarr backend

### DIFF
--- a/.github/workflows/streaming-tests.yml
+++ b/.github/workflows/streaming-tests.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.7", "3.10"]
+        python-version: ["3.8", "3.11"]
     steps:
       - uses: s-weigand/setup-conda@v1
         with:
@@ -26,10 +26,6 @@ jobs:
         run: |
           pip install -e .
           pip install dandi
-      - name: Uninstall h5py
-        run: pip uninstall -y h5py
-      - name: Install ROS3
-        run: conda install -c conda-forge h5py
 
       - name: Run pytest on stream tests
         run: pytest tests/ -rsx tests/streaming_tests.py
--- a/src/nwbinspector/nwbinspector.py
+++ b/src/nwbinspector/nwbinspector.py
@@ -32,7 +32,7 @@ from .inspector_tools import (
 )
 from .register_checks import InspectorMessage, Importance
 from .tools import get_s3_urls_and_dandi_paths
-from .utils import FilePathType, PathType, OptionalListOfStrings, robust_s3_read, calculate_number_of_cpu
+from .utils import FilePathType, PathType, calculate_number_of_cpu
 from nwbinspector import __version__
 
 INTERNAL_CONFIGS = dict(dandi=Path(__file__).parent / "internal_configs" / "dandi.inspector_config.yaml")
@@ -322,10 +322,10 @@ def read_nwb(nwbfile_path, stream: bool = False):
 
 def inspect_all(
     path: PathType,
-    modules: OptionalListOfStrings = None,
+    modules: Optional[List[str]] = None,
     config: Optional[dict] = None,
-    ignore: OptionalListOfStrings = None,
-    select: OptionalListOfStrings = None,
+    ignore: Optional[List[str]] = None,
+    select: Optional[List[str]] = None,
     importance_threshold: Union[str, Importance] = Importance.BEST_PRACTICE_SUGGESTION,
     n_jobs: int = 1,
     skip_validate: bool = False,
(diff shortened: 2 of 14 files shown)


CodyCBakerPhD:
Have you been able to run this branch successfully on your local device? I.e., 

```
nwbinspector 000350 --stream
```

as well as

```
nwbinspector 000350 --stream --n-jobs 4
```

(which should then run about ~4 times faster)

The CI reports certain issues, and when I try this I also encounter a couple of problems when I try it on the Hub (especially the second one - any time I've ever tried parallelizing `fsspec` streaming something goes weird and it just indefinitely stalls)

If this does run for you, then could you also report a speed comparison between this branch and current `dev` (which would still use ros3 as the baseline)

Also, is there any significant downside to keeping ROS3 as an option for the environments where it is already configured? Given that we haven't run into any more problems with it in the past ~6 months; `fsspec` could just be an alternate mode for people that don't / can't do the special `h5py` installation through conda

bendichter:
I didn't anticipate these issues with parallelizing. I'll keep working on this and see if I can iron this out, with a speed comparison. In my experience fsspec is faster and more reliable than ros3, but I agree IRS worth a test. I'm not aware of any use case that would require ros3 instead of fsspec, so I'd rather drop support for it unless we can come up with a compelling reason to keep it.

CodyCBakerPhD:
It might just require a refactor of the parallelization step to use `joblib` instead of `concurrent.futures.ProcessPoolExecutor`

See examples: https://gist.github.com/CodyCBakerPhD/b9d7d10ce1572dafcbd4646251781234

The `ThreadPoolExecutor` is possibly an easier option but last time I ran speed tests using it it bottlenecked pretty quickly and didn't scale as well as multiprocessing

bendichter:
@CodyCBakerPhD What do you think of this? I am adding back ros3 as an available "method" of `open_nwb`.

CodyCBakerPhD:
Sequential stream is working now, and is about 33% faster than ros3

The command

```
nwbinspector 000350 --stream --n-jobs 8 [or anything >=2]
```

Still stalls out as per my demonstration in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/379#issuecomment-1606525398

I'd recommend either switching to use `joblib` for parallelization (which apparently doesn't have whatever mysterious issue `fsspec` has with `ProcessPoolExecutor`)
OR an automatic switch in the CLI logic to use `ros3` as default stream mode when `n_jobs != 1`
OR raise an informative error if someone tries `fsspec` + `n_jobs != 1` telling them that combination only works for `ros3`

.github/workflows/streaming-tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.7", "3.10"]
+        python-version: ["3.8", "3.11"]
```
CodyCBakerPhD: This will have to get into `dev` first - probably easier to do it in a separate fast-track PR

src/nwbinspector/nwbinspector.py
```diff
@@ -297,18 +301,35 @@ def inspect_all_cli(
         print(f"{os.linesep*2}Report saved to {str(Path(report_file_path).absolute())}!{os.linesep}")
 
 
+@contextlib.contextmanager
+def open_nwb(nwbfile_path: str, mode: str = "r", method: Literal["local", "fsspec", "ros3"] = "local"):
```
bendichter, line 305:
```suggestion
def read_nwb(nwbfile_path: str, mode: str = "r", method: Literal["local", "fsspec", "ros3"] = "local"):
```

bendichter, line 305:
```suggestion
def open_nwb(nwbfile_path: str, mode: str = "r", method: Literal["local", "fsspec", "ros3"] = "local", return_io=False):
```

.github/workflows/streaming-tests.yml
bendichter: yup

CodyCBakerPhD: fixed in #380

CodyCBakerPhD:
Update: as mentioned in meeting, `joblib` won't work for us either because it breaks the generator behavior of the inspection functions

Martin mentioned a possible fix for `fsspec` so I'll try that out and see where it leads and then we can decide on whether to keep `ros3` around just for parallelization - I do agree that we should try to move away from it however, largely due to the upcoming Zarr ssupport

As I play around more with the context, as well as reflections on issue https://github.com/NeurodataWithoutBorders/pynwb/issues/1711 and PR https://github.com/NeurodataWithoutBorders/nwbwidgets/pull/295, I think we might want to make the introduction of the `read_nwbfile` context function an entirely separate PR of its own with extensive tests to ensure it behaves as we expect - we have already seen users attempt to define their own such helper functions with mixed success, so we really need to ensure that the version of it we support is done the proper way

CodyCBakerPhD:
replaced by #490

PR #386, opened 2023-07-10
[Read Helpers IIa] Add read_nwbfile with TestCase tests

CodyCBakerPhD:
Introducing the latest version of the `read_nwbfile` function as a stand-alone PR with its own unit tests; the integration of it into the full NWB Inspector workflow and subsequent integration testing +/- debugging will occur in a follow-up

I also reorganized the `tools` sub-module here to separate the division of each helper function (some of which may be able to be deprecated soon), which I will also splinter in a smaller PR (which should be merged first, or I'll just rebase this one...) to make this PR specific to only the introduction of the `read_nwbfile` function and its tests

This PR serves as an first example of the convenience functions that will and should eventually make their way up to PyNWB once confirmed to operate as intended to the best of our knowledge

Depends on [HDMF #882](https://github.com/hdmf-dev/hdmf/pull/882)


CodyCBakerPhD:
Was having too much trouble getting the files to stick around and cleanup, deferring to #388 (which has its own CI issues but the tests operate on the files locally at least)

PR #387, opened 2023-07-10
[Read Helpers I] Reorganize tools sub-module

CodyCBakerPhD:
pre-PR to #386  to reorganize the tools sub-module according to each type of tool (testing-specific, NWB-specific, DANDI-specific)

No import structure has changed aside from secondary imports that aren't encouraged anyway (like `from nwbinspector.tools import NWBFile`)


src/nwbinspector/tools/testing.py
```diff
@@ -0,0 +1,54 @@
...
+    return True, None
+
+
+class TemporaryFolderTestCase(HDMFTestCase):
+    @classmethod
+    def setUpClass(cls):
+        """
+        The `tmpdir` pytest fixture does not immediately clean itself up after running the test suite.
+
+        Instead it caches some number of total testing runs under a temporary folder.
```
CodyCBakerPhD: Now, technically pytest will eventually clean up 'older' runs that it has cached - so it's mostly just personal preference that I think the testing suite should immediately and automatically clean itself up after each run for tests that write things to disk

You can verify this yourself by injecting something like

```python
def test_tmpdir(tmpdir):
    from pathlib import Path
    print(Path(tmpdir))  # To see save path on your system

    import numpy as np
    np.memmap(filename=Path(tmpdir)/"test.dat", shape=(10,2), dtype="int16", mode="w+")
    assert False # to force the print above to display in log
```

into a `test_<something>.py` file and running `pytest` directly on it several times, then watching what happens to the parent folder on the system

.github/workflows/update-testing-files.yml
```diff
@@ -25,6 +25,10 @@ jobs:
 
       - name: Install DANDI
         run: pip install -e .[dandi]
+      - name: Install older PyNWB version to generate example invalid files
+        run: |
+          pip uninstall pynwb
+          pip install pynwb==2.1.0
```
CodyCBakerPhD, line 31: There was a bug with this older workflow, which only triggers when changes are made to the testing submodule

In hindsight it might make more sense to use a docker container for this type of approach going into the future (whenever we want to make invalid NWB files or NWB files violating best practices that modern versions of PyNWB prohibits)

CodyCBakerPhD: (the workflow will continue failing on this PR since the deployment is pinned to the version on `main` - I could pin it to this branch but then I'd have to go back and fix it after merge so it's w/e you prefer)

CodyCBakerPhD:
Moving back to draft, having some troubles in testing #386 with tmpdir persistence so going back and forth between which method might be best cross-platform

src/nwbinspector/testing.py
```diff
@@ -118,3 +125,49 @@ def generate_image_series_testing_files():  # pragma: no cover
...
+        """
+        cls.temporary_folder = mkdtemp()
+
+    @classmethod
+    def tearDownClass(cls):
+        """Note that this cleanup is only attempted and not guaranteed to succeed on stuck I/O; mostly on Windows."""
+        try:
+            rmtree(cls.temporary_folder)
+        except PermissionError:  # pragma: no cover
+            warn(f"Unable to clean up the temporary folder {cls.temporary_folder}!")
```
CodyCBakerPhD:
```suggestion
```

CodyCBakerPhD: Just moving it as a potential addition specific to #386

src/nwbinspector/tools/testing.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        cls.temporary_folder = mkdtemp()
+
+    @classmethod
+    def tearDownClass(cls):
+        """Note that this cleanup is only attempted and not guaranteed to succeed on stuck I/O; mostly on Windows."""
+        try:
+            rmtree(cls.temporary_folder)
+        except PermissionError:  # pragma: no cover
+            warn(f"Unable to clean up the temporary folder {cls.temporary_folder}!")
```
CodyCBakerPhD:
```suggestion
```

CodyCBakerPhD: Moving to potential addition in #386

bendichter (approved):

PR #388, opened 2023-07-10
[Read Helpers IIb] Add read_nwbfile with Pytest fixtures

CodyCBakerPhD:
Analog of #386 but using pytest fixtures instead of `unittest.TestCase`


CodyCBakerPhD:
The CI has trouble untangling the HDMF versioning, but manual local dev installation of each package allows the tests to run

Getting some actual legit errors now though, beginning debugging now

CodyCBakerPhD:
Update: the `test_<backend>_explicit_closure` tests are passing for both backends now

However the other tests related to deletion are legitimately failing locally, so https://github.com/hdmf-dev/hdmf/pull/882 might need some more investigation

bendichter:
Notes from convo:

*add tests for fsspec
   * https and s3
   * hdf5 and zarr



* add tests for ros3 (just for hdf5)

src/nwbinspector/testing.py
```diff
@@ -150,3 +148,34 @@ def check_streaming_enabled() -> Tuple[bool, Optional[str]]:
...
+    try:
+        io._file["acquisition"]["TimeSeries"]["data"][:2]
+        assert False, "The test to close the HDF5 I/O failed!"  # The line above should throw a ValueError
+    except ValueError as exception:
+        assert str(exception) in [
+            "Invalid location identifier (invalid location identifier)",
+            "Not a location (invalid object ID)",
+        ]
+    except KeyError as exception:  # Error type may depend on h5py version and/or installation source
+        assert str(exception) == "'Unable to synchronously open object (invalid identifier type to function)'"
```
bendichter:
```suggestion
```

src/nwbinspector/testing.py
```diff
@@ -150,3 +148,34 @@ def check_streaming_enabled() -> Tuple[bool, Optional[str]]:
     if "ros3" not in h5py.registered_drivers():
         return False, "ROS3 driver not installed."
     return True, None
+
+
+# HDF5 assertion styles
+def check_hdf5_io_open(io: HDMFIO):
+    """For HDF5, the file is 'open' if we can access data from slicing one of its `h5py.Dataset` objects."""
+    assert io._file["acquisition"]["TimeSeries"]["data"][:2] is not None
```
bendichter:
```suggestion
# HDF5 assertion styles
def check_hdf5_io_open(io: HDF5IO):
    """You can check if an h5py.File object is open or not by using the file object's .id attribute, which is invalid when the file is closed."""
    return io._file.id.valid
```

CodyCBakerPhD:
@bendichter All tests green now

.github/workflows/streaming-tests.yml
```diff
@@ -26,6 +26,10 @@ jobs:
         run: |
           pip install -e .
           pip install dandi
+      - name: Install HDMF-zarr # temporary
+        run: pip install hdmf-zarr
+      - name: Install latest HDMF # temporary
+        run: pip install -U hdmf
```
bendichter, line 32: It looks like it might be better to switch to a requirements-tests.txt file

CodyCBakerPhD: This is temporary b/c of hdmf-zarr's dependencies - it's normally not needed at all for that workflow

CodyCBakerPhD: Also the goal here is not to make such things directly be base requirements of nwbinspector (instead relying on `pynwb` to handle it); this is only temporary stuff for running the tests that will be ported to PyNWB after thorough evaluation using the NWB Inspector

bendichter (approved):

PR #390, opened 2023-07-17
trying to fix error in test file updates

CodyCBakerPhD:
Noticed in https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/5579433909/jobs/10194979342?pr=388 from https://github.com/NeurodataWithoutBorders/nwbinspector/pull/388

that this particular workflow was failing due to awaiting a Yes/No confirmation in the CI, seeing if this fixes it

### DIFF
--- a/.github/workflows/update-testing-files.yml
+++ b/.github/workflows/update-testing-files.yml
@@ -27,7 +27,7 @@ jobs:
         run: pip install -e .[dandi]
       - name: Install older PyNWB version to generate example invalid files
         run: |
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Bah, now it's failing due to an incompatibility between latest DANDI and much older PyNWB version: https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/5591921621/jobs/10223671392

Will need to think on different process for this, suppressing the workflow for the time being

PR #393, opened 2023-07-25
Should check_subject_exists be CRITICAL? (i.e., should it block DANDI upload)

CodyCBakerPhD:
## Motivation

Noticed when looking at the report from someone at the User Days, this check should maybe have a higher importance since dandi upload is actually impossible if a Subject is not in the file

### DIFF
--- a/src/nwbinspector/checks/nwbfile_metadata.py
+++ b/src/nwbinspector/checks/nwbfile_metadata.py
@@ -104,7 +104,7 @@ def check_keywords(nwbfile: NWBFile):
         return InspectorMessage(message="Metadata /general/keywords is missing.")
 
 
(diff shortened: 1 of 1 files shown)


bendichter:
This check is critical in the DANDI config: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/internal_configs/dandi.inspector_config.yaml

CodyCBakerPhD:
Right; the remark from the user was that they needed to know ahead of time to set their config to DANDI level to get informed about this

Otherwise, 'SUGGESTION' level messages are usually viewed as 'safe to ignore' but they remarked that this likely could have led them to spend a fair amount of time writing a big NWB file only to learn that it needed a subject added to it to be valid for DANDI upload

Maybe bumping it up to a VIOLATION could help emphasize the importance?

PR #404, opened 2023-09-18
[New Check]: TimeSeries rate is zero

### LINKED ISSUE #389: [Add Check]: rate of TimeSeries is not zero

Apparently it's possible for TimeSeries to have a rate set to `0.0`, though that makes no sense when the series has more than one frame (technically there is an edge case for ImageSeries and its variants where it can be interpreted to mean a static ophys image)

_No response_

Yes.

### Code of Conduct

alessandratrapani:
Related to Issue #389 
Add a check function to check if the sampling rate is set to 0.0Hz

- [x]  add check function
- [x]  add unit tests
- [x]  add check function documentation
- [x]  update CHANGELOG.md

### DIFF
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -165,7 +165,6 @@ Check function: :py::meth:`~nwbinspector.checks.time_series.check_resolution`
 Zero Rate
 ~~~~~~~~~
 
-If ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, thus first dimension that represent time is
-greater than one, ``rate`` should not be 0.0.
+If the ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, and according to :ref:`best_practice_data_orientation` this axis ought to be time, then the ``rate`` field should not be ``0.0``.
 
 Check function: :py::meth:`~nwbinspector.checks.time_series.check_rate_is_not_zero`
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -157,3 +157,15 @@ Unknown Resolution
 If the ``resolution`` of a :ref:`nwb-schema:sec-TimeSeries` is unknown, use ``-1.0`` or ``NaN`` to indicate this.
 
 Check function: :py::meth:`~nwbinspector.checks.time_series.check_resolution`
+
+
+
+.. _best_practice_non_zero_rate:
+
+Zero Rate
+~~~~~~~~~~~~~~~~~~
+
+If ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, thus first dimension that represent time is
+greater than one, ``rate`` should not be 0.0.
+
+Check function: :py::meth:`~nwbinspector.checks.time_series.check_rate_is_not_zero`
--- a/docs/best_practices/time_series.rst
+++ b/docs/best_practices/time_series.rst
@@ -163,7 +163,7 @@ Check function: :py::meth:`~nwbinspector.checks.time_series.check_resolution`
 .. _best_practice_non_zero_rate:
 
 Zero Rate
-~~~~~~~~~~~~~~~~~~
+~~~~~~~~~
 
 If ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, thus first dimension that represent time is
 greater than one, ``rate`` should not be 0.0.
--- a/src/nwbinspector/checks/time_series.py
+++ b/src/nwbinspector/checks/time_series.py
@@ -133,9 +133,10 @@ def check_resolution(time_series: TimeSeries):
 
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
 def check_rate_is_not_zero(time_series: TimeSeries):
-    if time_series.data is None and time_series.timestamps is None:
+    if time_series.data is None:
         return
-    data_shape = get_data_shape(time_series.data) 
-    timestamps_shape = get_data_shape(time_series.timestamps)
-    if time_series.rate == 0.0 and (data_shape[0] > 1 or timestamps_shape[0] > 1):
-        return InspectorMessage(f"{time_series.name} has a sampling rate value of 0.0Hz but the series has more than one frame.")
+    data_shape = get_data_shape(time_series.data)
+    if time_series.rate == 0.0 and data_shape[0] > 1:
+        return InspectorMessage(
+            f"{time_series.name} has a sampling rate value of 0.0Hz but the series has more than one frame."
+        )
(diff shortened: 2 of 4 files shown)


src/nwbinspector/checks/time_series.py
```diff
@@ -129,3 +129,14 @@ def check_resolution(time_series: TimeSeries):
...
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_rate_is_not_zero(time_series: TimeSeries):
+    if time_series.data is not None:
+        data_shape = get_data_shape(time_series.data) 
+    if time_series.timestamps is not None:
+        timestamps_shape = get_data_shape(time_series.timestamps)
+    if time_series.rate == 0.0:
+        if (data_shape[0] > 1 or timestamps_shape[0] > 1):
```
CodyCBakerPhD: A good first opportunity here to introduce you to [Cyclomatic Complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) and our approaches to keeping it as low as possible

Python has short-circuit logic for conditional statements, so that if an earlier one fails, it never proceeds to the next ones after

```python
    if time_series.rate == 0.0 and (data_shape[0] > 1 or timestamps_shape[0] > 1):
```

voila, the code is now nested by one less level. We try to keep code as flat as possible within reason (sometimes it's unavoidable of course...)

A lot of other inspirations we use for code style relate to [PEP 20](https://peps.python.org/pep-0020/), so I'd give that a quick look as well

src/nwbinspector/checks/time_series.py
```diff
@@ -129,3 +129,14 @@ def check_resolution(time_series: TimeSeries):
         return InspectorMessage(
             message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
         )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_rate_is_not_zero(time_series: TimeSeries):
+    if time_series.data is not None:
+        data_shape = get_data_shape(time_series.data)
```
CodyCBakerPhD: Another suggestion related to the other complexity one; this is more of a style, but it is one that does help readability when you get into the habit

Early exit conditions are easier to understand than later structural returns

What I mean by that, I'll just show here

```python
if time_series.data is None:
    return
data_shape = get_data_shape(time_series.data)   # no need for an else statement either
```

This also helps with a minor issue where `data_shape` wouldn't even be a defined variable when/if the program hits lines 141 below

alessandratrapani: Ok, thanks! I am including your correction.

src/nwbinspector/checks/time_series.py
alessandratrapani: Thank you for this correction as well, I will keep this in mind

CodyCBakerPhD:
A couple things here:

(i) need unit tests

(ii) It occurs to me that for all of these PRs, you will have to run `pre-commit` manually since the bot does not work when the contributor is technically a GitHub organization (weird edge case with permissions that no one seems to be able to figure out - external forks are fine whenever they are an individual rather than an organization)

It's very simple though

```
pip install pre-commit
pre-commit install
pre-commit run -a
```

from a terminal while inside your clone of the CN fork of the repo

tests/unit_tests/test_time_series.py
```diff
@@ -292,6 +293,28 @@ def test_check_unknown_resolution_pass():
...
+
+def test_check_rate_is_not_zero_single_frame_pass():
+    time_series = pynwb.TimeSeries(name="test", unit="test_units", data=[1], rate=0.0)
+    assert check_rate_is_not_zero(time_series) is None
+
+
+def test_check_rate_is_not_zero_timestamp_is_none_fail():
+    time_series = pynwb.TimeSeries(name="test", unit="test_units", data=[1, 2, 3], rate=0.0)
+    assert check_rate_is_not_zero(time_series) == InspectorMessage(
+        message=f"{time_series.name} has a sampling rate value of 0.0Hz but the series has more than one frame.",
```
CodyCBakerPhD: Minor suggestion

Lately we've been in the habit of making tests as explicit as possible to help identify potential issues quicker - a part of that means avoiding f-strings where possible / not inconvenient

```suggestion
    time_series = pynwb.TimeSeries(name="TimeSeriesTestt", unit="n.a.", data=[1, 2, 3], rate=0.0)
    assert check_rate_is_not_zero(time_series) == InspectorMessage(
        message="TimeSeriesTest has a sampling rate value of 0.0Hz but the series has more than one frame.",
```

src/nwbinspector/checks/time_series.py
```diff
@@ -129,3 +129,14 @@ def check_resolution(time_series: TimeSeries):
         return InspectorMessage(
             message=f"'resolution' should use -1.0 or NaN for unknown instead of {time_series.resolution}."
         )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_rate_is_not_zero(time_series: TimeSeries):
```
CodyCBakerPhD: Hm...

Big question is, should this be a VIOLATION or CRITICAL, since any time series with more than one data element on the first axis and a rate of zero is practically invalid?

alessandratrapani: I agree that maybe CRITICAL would be better, I will change it

tests/unit_tests/test_time_series.py
alessandratrapani: Ok! thanks for the correction

docs/best_practices/time_series.rst
```diff
@@ -157,3 +157,15 @@ Unknown Resolution
 If the ``resolution`` of a :ref:`nwb-schema:sec-TimeSeries` is unknown, use ``-1.0`` or ``NaN`` to indicate this.
 
 Check function: :py::meth:`~nwbinspector.checks.time_series.check_resolution`
+
+
+
+.. _best_practice_non_zero_rate:
+
+Zero Rate
+~~~~~~~~~~~~~~~~~~
```
CodyCBakerPhD:
```suggestion
Zero Rate
~~~~~~~~~
```

length of tildes should match length of title; sphinx build will warn about this but I don't think it causes an error yet

docs/best_practices/time_series.rst
```diff
@@ -157,3 +157,15 @@ Unknown Resolution
...
+
+If ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, thus first dimension that represent time is
+greater than one, ``rate`` should not be 0.0.
```
CodyCBakerPhD:
```suggestion
If the ``data`` field of :ref:`nwb-schema:sec-TimeSeries` has more than one frame, and according to :ref:`best_practice_data_orientation` this axis ought to be time, then the ``rate`` field should not be ``0.0``.
```

CodyCBakerPhD:
This looks good, just need a changelog entry (also if you approve #406 I can merge that in first so it can always be there to flag it on PRs)

CodyCBakerPhD (approved):

bendichter:
congratulations @alessandratrapani on your first contribution!

PR #405, opened 2023-09-18
[New Check]: Spike times are increasing

### LINKED ISSUE #394: [Add Check]: Check if spike times are increasing

@gviejo noticed in DANDI set `000003`, file `sub-YutaMouse41/sub-YutaMouse41_ses-YutaMouse41-150831_behavior+ecephys.nwb`, unit ID `1`

The spike times reset to zero about halfway through

Investigating if this is a problem with the conversion code or with the source data itself

But we should probably add a check for this in the Inspector to catch it going forward

_No response_

Yes.

### Code of Conduct

alessandratrapani:
Relate to Issue #394
Add check function for ascending spike times

- [x] add check function
- [x]  add unit tests
- [x]  add check function documentation
- [x]  update CHANGELOG.md

### DIFF
--- a/docs/best_practices/ecephys.rst
+++ b/docs/best_practices/ecephys.rst
@@ -83,3 +83,14 @@ Observation Intervals
 The ``obs_intervals`` field of the :ref:`nwb-schema:sec-units-src` table is used to indicate periods of time where the underlying electrical signal(s) were not observed. This can happen if the recording site moves away from the unit, or if the recording is stopped. Since the channel is not observed, it is not determinable whether a spike occurred during this time. Therefore, there should not be any identified spike times for units matched to those electrical signal(s) occurring outside of the defined ``obs_intervals``. If this variable is not set, it is assumed that all time is observed.
 
 Check function: :py:meth:`~nwbinspector.checks.ecephys.check_spike_times_not_in_unobserved_interval`
+
+
+
+.. _best_practice_ascending_spike_times:
+
+Ascending Spike Times
+~~~~~~~~~~~~~~~~~~~~~
+
+All spike times should be sorted in ascending order. Spike times may reset to zero about halfway through. This can happen if the spikes are not properly aligned with the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`.
+
+Check function: :py:meth:`~nwbinspector.checks.ecephys.check_ascending_spike_times`
--- a/docs/best_practices/ecephys.rst
+++ b/docs/best_practices/ecephys.rst
@@ -91,6 +91,6 @@ Check function: :py:meth:`~nwbinspector.checks.ecephys.check_spike_times_not_in_
 Ascending Spike Times
 ~~~~~~~~~~~~~~~~~~~~~
 
-All spike times should be sorted in ascending order. Spike times may reset to zero about halfway through. This can happen if the spikes are not properly aligned with the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`.
+All spike times should be sorted in ascending order. If they reset to zero, this can be a sign that spikes are not properly aligned with the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`.
 
 Check function: :py:meth:`~nwbinspector.checks.ecephys.check_ascending_spike_times`
--- a/src/nwbinspector/checks/ecephys.py
+++ b/src/nwbinspector/checks/ecephys.py
@@ -89,8 +89,4 @@ def check_ascending_spike_times(units_table: Units):
     for spike_times_per_unit in units_table["spike_times"]:
         differences_per_unit = np.diff(spike_times_per_unit)
     if np.all(differences_per_unit >= 0):
-        return InspectorMessage(
-            message=(
-                "This Units table contains spike times that are not ascending."
-            )
-        )
+        return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
--- a/src/nwbinspector/checks/ecephys.py
+++ b/src/nwbinspector/checks/ecephys.py
@@ -87,6 +87,5 @@ def check_ascending_spike_times(units_table: Units):
     if "spike_times" not in units_table:
         return
     for spike_times_per_unit in units_table["spike_times"]:
-        differences_per_unit = np.diff(spike_times_per_unit)
-    if np.all(differences_per_unit >= 0):
-        return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
+        if not np.all(spike_times_per_unit[:-1] <= spike_times_per_unit[1:]):
+            return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
--- a/src/nwbinspector/checks/ecephys.py
+++ b/src/nwbinspector/checks/ecephys.py
@@ -86,8 +86,9 @@ def check_ascending_spike_times(units_table: Units):
     """Check that the values in the timestamps array are strictly increasing."""
     if "spike_times" not in units_table:
         return
-    differences = np.diff(np.asarray(units_table["spike_times"].target.data[:]))
-    if np.all(differences >= 0):
+    for spike_times_per_unit in units_table["spike_times"]:
+        differences_per_unit = np.diff(spike_times_per_unit)
+    if np.all(differences_per_unit >= 0):
         return InspectorMessage(
             message=(
                 "This Units table contains spike times that are not ascending."
--- a/src/nwbinspector/checks/ecephys.py
+++ b/src/nwbinspector/checks/ecephys.py
@@ -79,3 +79,17 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
                     "observed intervals."
                 )
             )
+
+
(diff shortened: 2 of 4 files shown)


src/nwbinspector/checks/ecephys.py
```diff
@@ -79,3 +79,17 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
...
                 )
             )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_ascending_spike_times(units_table: Units):
+    """Check that the values in the timestamps array are strictly increasing."""
+    if "spike_times" not in units_table:
+        return
+    differences = np.diff(np.asarray(units_table["spike_times"].target.data[:]))
```
CodyCBakerPhD: I do believe for this one we will have to iterate over the units identified via the API convenience functions, _i.e._,

```python
for unit_index in range(len(units_table.id)):
    spike_times_per_unit = units_table.get_unit_spike_times(index=unit_index)
    differences_per_unit = np.diff(spike_times_per_unit)
```

Reason for this will be clear when you add tests of this function; creating a simple `Units` table with two units, each having a few 'good' (monotone increasing) spike times, then looking at the value of  `units_table["spike_times"].target.data[:]` will reveal quite a lot about how NWB structures the table

bendichter: This would be simpler:

```python
for spike_times_per_unit in units_table["spike_times"]:
    differences_per_unit = np.diff(spike_times_per_unit)
```

bendichter:
Thanks @alessandratrapani. We also need units tests

src/nwbinspector/checks/ecephys.py
```diff
@@ -86,10 +86,7 @@ def check_ascending_spike_times(units_table: Units):
...
-        return InspectorMessage(
-            message=(
-                "This Units table contains spike times that are not ascending."
-            )
-        )
\ No newline at end of file
+    for spike_times_per_unit in units_table["spike_times"]:
+        differences_per_unit = np.diff(spike_times_per_unit)
+    if np.all(differences_per_unit >= 0):
+        return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
```
bendichter: see more efficient approach [here](https://stackoverflow.com/a/47004507)

src/nwbinspector/checks/ecephys.py
```diff
@@ -79,3 +79,14 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
...
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_ascending_spike_times(units_table: Units):
+    """Check that the values in the timestamps array are strictly increasing."""
+    if "spike_times" not in units_table:
+        return
+    for spike_times_per_unit in units_table["spike_times"]:
+        differences_per_unit = np.diff(spike_times_per_unit)
+    if np.all(differences_per_unit >= 0):
+        return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
```
CodyCBakerPhD:
```suggestion
        if not np.all(spike_times_per_unit[:-1] <= spike_times_per_unit[1:]):
            return InspectorMessage(message=("This Units table contains spike times that are not ascending."))
```

A couple of suggestions here:

(i) from Ben's pointer, using the index shifting strategy instead (compare the array to a version of itself shifted by one index to the right)

(ii) indenting the `if` statement per unit inside the `for` loop

(iii) reversing the logic to say 'if not all entries are in sorted order' in line with (i); this should inherit an early return functionality where the first time it encounters a `False` in the array, it should exit

tests/unit_tests/test_ecephys.py
```diff
@@ -215,3 +216,24 @@ def test_check_spike_times_not_in_unobserved_interval_multiple_units():
...
+
+
+def test_check_ascending_spike_times_pass():
+    units_table = Units()
+    units_table.add_unit(spike_times=[0.0, 0.1])
+    units_table.add_unit(spike_times=[1.0, 2.0])
+    assert check_ascending_spike_times(units_table=units_table) is None
+
+
+def test_check_ascending_spike_times_fail():
```
CodyCBakerPhD: These checks look good - for coverage, can you add one more that the missing spike times is properly skipped (i.e., a simple test with a Units table with no units, checking that the function returns None and does not error out)

alessandratrapani: Sure! Thank you for the suggestion

bendichter:
@alessandratrapani can you change the status of this PR from Draft?

bendichter:
@alessandratrapani can you please add an entry [here](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/docs/best_practices/ecephys.rst) explaining this best practice in plain English?

tests/unit_tests/test_ecephys.py
```diff
@@ -215,3 +216,31 @@ def test_check_spike_times_not_in_unobserved_interval_multiple_units():
...
+    units_table = Units()
+    units_table.add_unit(spike_times=[0.0, 0.1])
+    units_table.add_unit(spike_times=[1.0, 2.0])
+    assert check_ascending_spike_times(units_table=units_table) is None
+
+
+def test_check_ascending_spike_times_w_missing_spike_times_pass():
+    units_table = Units()
+    units_table.add_unit(spike_times=[0.0, 0.1])
+    units_table.add_unit(spike_times=[])
```
CodyCBakerPhD, line 231:
```suggestion
def test_check_ascending_spike_times_missing_column_skip():
    units_table = Units()
```

The point of this test is to evoke the early return of https://github.com/NeurodataWithoutBorders/nwbinspector/pull/405/files#diff-dfe54828956c62fba202acb9497ca060f0448c8e0cd1b945cf31d0daaa53d769R87

which is for the case where we have an empty Units table

This would of course trigger a separate check function: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/checks/tables.py#L40

but the point is still that we don't want the logic of the check to reach https://github.com/NeurodataWithoutBorders/nwbinspector/pull/405/files#diff-dfe54828956c62fba202acb9497ca060f0448c8e0cd1b945cf31d0daaa53d769R89 and error out because the column doesn't exist

bendichter: @CodyCBakerPhD yeah but I think it would also not be a bad idea to include the existing test as well

CodyCBakerPhD: To check the safe range slicing of https://github.com/NeurodataWithoutBorders/nwbinspector/pull/405/files#diff-dfe54828956c62fba202acb9497ca060f0448c8e0cd1b945cf31d0daaa53d769R90, sure

Both would be fine to have in that case

alessandratrapani: Ok, keeping both

docs/best_practices/ecephys.rst
```diff
@@ -83,3 +83,14 @@ Observation Intervals
...
 Check function: :py:meth:`~nwbinspector.checks.ecephys.check_spike_times_not_in_unobserved_interval`
+
+
+
+.. _best_practice_ascending_spike_times:
+
+Ascending Spike Times
+~~~~~~~~~~~~~~~~~~~~~
+
+All spike times should be sorted in ascending order. Spike times may reset to zero about halfway through. This can happen if the spikes are not properly aligned with the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`.
```
bendichter:
```suggestion
All spike times should be sorted in ascending order. If they reset to zero, this can be a sign that spikes are not properly aligned with the ``timestamps_reference_time`` of the :ref:`nwb-schema:sec-NWBFile`.
```

Reworded to not include "may" which can be interpretted to mean that spike times resetting is permitted.

bendichter (approved):
lgtm!

bendichter:
@alessandratrapani , can you resolve the conflicts?

image

src/nwbinspector/checks/ecephys.py
```diff
@@ -79,3 +79,13 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
                     "observed intervals."
                 )
             )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_ascending_spike_times(units_table: Units):
```
CodyCBakerPhD, line 85: We are still testing the performance of this, since it's the first data heavy check to be added without options to control how much data is loaded

CodyCBakerPhD (changes requested):

src/nwbinspector/checks/ecephys.py
CodyCBakerPhD: @alessandratrapani Can you add a `nelems` option here to mirror other table tests?

bendichter: ping @alessandratrapani

bendichter:
@daphnedequatrebarbes could you take a look at this PR? It looks like it might be best to redo this from scratch rather than trying to fix the merge conflicts

PR #408, opened 2023-09-27
[New Check]: Add check electricalseries of dtype int16

### LINKED ISSUE #395: [Add Check]: ElectricalSeries of dtype int16 with conversion=1.0 and offset=0.0 is potentially wrong?

More of a question for either @alejoe91 or @samuelgarcia based on their experience

If you saw an ElectricalSeries that was of dtype `uint16` or `int16` and the conversion factor (spikeinterface 'gain_to_uV') was all `1.0` and offset `0.0`, this is most likely incorrect, right?

If so, how strongly incorrect is it?

@bendichter @rly @oruebel Would/Should it technically count as being invalid? Should it prevent DANDI upload?

_No response_

Yes.

### Code of Conduct

alessandratrapani:
Fix #395 

Add check on ElectricalSeries data type: if the conversion and offset are DEFAULT values and the data type is any integer ("uint16", "int16", etc). 

- [x] add check function
- [x] add unit tests
- [ ] add check function documentation in ecephys.rst 
- [x] update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Upcoming
 
+### New Checks
+
+* Add `check_electrical_series_dtype` [#395](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/395)
+
 # v0.4.30
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### New Checks
 
-* Add `check_electrical_series_dtype` [#395](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/395)
+* Added `check_electrical_series_conversion_factors` for ensuring non-default conversion/offset factors of any `ElectricalSeries` that has an integer data type. [#395](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/395)
 
 # v0.4.30
 
--- a/src/nwbinspector/checks/ecephys.py
+++ b/src/nwbinspector/checks/ecephys.py
@@ -82,7 +82,7 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
 
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=ElectricalSeries)
-def check_electrical_series_dtype(electrical_series: ElectricalSeries):
+def check_electrical_series_conversion_factors(electrical_series: ElectricalSeries):
     data = electrical_series.data
     if (
         np.issubdtype(data.dtype, np.integer)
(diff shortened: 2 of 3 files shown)


tests/unit_tests/test_ecephys.py
```diff
@@ -155,6 +156,76 @@ def test_trigger_check_electrical_series_reference_electrodes_table(self):
             == "electrodes does not  reference an electrodes table."
         )
 
+    def test_check_electrical_series_dtype_pass(self):
+        electrodes = self.nwbfile.create_electrode_table_region(region=[0, 1, 2, 3, 4], description="all")
+
+        electrical_series = ElectricalSeries(
+            name="elec_series",
+            description="desc",
+            data=np.ones((100, 5), dtype=float),
```
CodyCBakerPhD:
```suggestion
            data=np.ones((100, 5), dtype=np.dtype("float64")),
```

Specifying the native Python `float` type defaults to using `np.dtype("float64")`; I'd recommend writing this explicitly instead of letting numpy determine it implicitly

src/nwbinspector/checks/ecephys.py
```diff
@@ -79,3 +79,16 @@ def check_spike_times_not_in_unobserved_interval(units_table: Units, nunits: int
                     "observed intervals."
                 )
             )
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=ElectricalSeries)
+def check_electrical_series_dtype(electrical_series: ElectricalSeries):
```
CodyCBakerPhD:
```suggestion
def check_electrical_series_conversion_factors(electrical_series: ElectricalSeries):
```

Since we're not just testing the the dtype, suggesting a more descriptive name; if you have any better ideas for a name feel free to suggest those as well

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Upcoming
 
+### New Checks
+
+* Add `check_electrical_series_dtype` [#395](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/395)
```
CodyCBakerPhD:
```suggestion
* Added `check_electrical_series_conversion_factors` for ensuring non-default conversion/offset factors of any `ElectricalSeries` that has an integer data type. [#395](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/395)
```

src/nwbinspector/checks/ecephys.py
alessandratrapani: Thank you, makes more sense

CodyCBakerPhD:
Just occurred to me while inspecting some files on the GUIDE; we should also consider the `channel_conversion` array in this check

If the attribute exists, and the elements of it are not also `1.0`, then it's fine to have `conversion=1.0` and `offset=0.0` because the array of conversion factors is the intended way of specifying that information

PR #411, opened 2023-10-06
Clarify 'or' requirement in message for subject age + DOB

CodyCBakerPhD:
We discovered this during an alpha test and it caused some confusion

DANDI only requires one or the other; theoretically possible to have both, but the message implies it's required to have both, which is not true

### DIFF
--- a/src/nwbinspector/checks/nwbfile_metadata.py
+++ b/src/nwbinspector/checks/nwbfile_metadata.py
@@ -130,6 +130,42 @@ def check_doi_publications(nwbfile: NWBFile):
 
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=Subject)
+def check_subject_date_of_birth(subject: Subject):
+    """Check if the Subject age is in ISO 8601 or our extension of it for ranges."""
+    if subject.date_of_birth is None and subject.age is None:
+        return InspectorMessage(message="Subject is missing age or date_of_birth.")
+    elif subject.date_of_birth is None and subject.age is not None:
+        return
+    try:
+        datetime.fromisoformat(subject.date_of_birth)
+    except ValueError as exception:
+        assert str(exception) == f"Invalid isoformat string: '{subject.date_of_birth}'"
+        return InspectorMessage(
+            message=(
+                f"Subject date_of_birth, '{subject.date_of_birth}', does not follow ISO 8601 format, "
+                "e.g., '2016-12-13' for the date or '2016-12-13T21:20:37.593194+00:00' for the full datetime."
+            )
+        )
+
+    if "/" in subject.age:
+        subject_lower_age_bound, subject_upper_age_bound = subject.age.split("/")
+
+        if re.fullmatch(pattern=duration_regex, string=subject_lower_age_bound) and (
+            re.fullmatch(pattern=duration_regex, string=subject_upper_age_bound) or subject_upper_age_bound == ""
+        ):
+            return
+
+    return InspectorMessage(
+        message=(
+            f"Subject age, '{subject.age}', does not follow ISO 8601 duration format, e.g. 'P2Y' for 2 years "
+            "or 'P23W' for 23 weeks. You may also specify a range using a '/' separator, e.g., 'P1D/P3D' for an "
+            "age range somewhere from 1 to 3 days. If you cannot specify the upper bound of the range, "
+            "you may leave the right side blank, e.g., 'P90Y/' means 90 years old or older."
+        )
+    )
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=Subject)
 def check_subject_age(subject: Subject):
     """Check if the Subject age is in ISO 8601 or our extension of it for ranges."""
     if subject.age is None and subject.date_of_birth is None:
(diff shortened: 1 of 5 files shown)


garrettmflynn:
While a better direction for users of the GUIDE, it's technically incorrect—as the metadata _is_ missing age AND date of birth.

Would we be able to note `Subject requires an age or date of birth.` instead? 

Also, the reason it's confusing on the GUIDE is that only one of the fields (`age`) is checked AND that field provides information about both values. Would it be out of the question to expose the two checks separately—possibly combining them for an aggregate check that still behaves like this one?

CodyCBakerPhD:
> While a better direction for users of the GUIDE, it's technically incorrect—as the metadata is missing age AND date of birth.

I've modified the logic to reflect the cases

> Would we be able to note Subject requires an age or date of birth. instead?

That is what is says now, right?

> Would it be out of the question to expose the two checks separately

The check applies to the Subject object entirely, and via PyNWB the form of `date_of_birth` is already checked for format so no need for an extra check function

garrettmflynn:
> That is what is says now, right?

I guess when I read "Subject is missing age or date_of_birth.", I think _one is provided_ but the other is not. But this PR was raised because "Subject is missing age and date_of_birth." implies that both _are / could be_ required—though it is correctly describing the Subject metadata.

What both of these messages fail to accomplish is a clear indication of the user's course of action based on the requirements. So I was thinking a message that directly states "Subject requires an age or date of birth." is a better way to describe _what the user should do_ (provide one or the other) rather than merely describe _the status of the fields being evaluated_ by the Inspector.

src/nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -132,11 +132,10 @@ def check_doi_publications(nwbfile: NWBFile):
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=Subject)
 def check_subject_age(subject: Subject):
     """Check if the Subject age is in ISO 8601 or our extension of it for ranges."""
-    if subject.age is None:
-        if subject.date_of_birth is None:
-            return InspectorMessage(message="Subject is missing age and date_of_birth.")
-        else:
-            return
+    if subject.age is None and subject.date_of_birth is None:
+        return InspectorMessage(message="Subject is missing age or date_of_birth.")
```
CodyCBakerPhD:
```suggestion
        return InspectorMessage(message="Subject is missing age or date_of_birth. Please specify at least one.")
```

tests/unit_tests/test_nwbfile_metadata.py
```diff
@@ -336,7 +336,7 @@ def test_pass_check_subject_age_with_dob():
 def test_check_subject_age_missing():
     subject = Subject(subject_id="001")
     assert check_subject_age(subject) == InspectorMessage(
-        message="Subject is missing age and date_of_birth.",
+        message="Subject is missing age or date_of_birth.",
```
CodyCBakerPhD:
```suggestion
        message="Subject is missing age or date_of_birth. Please specify at least one.",
```

CodyCBakerPhD:
Except the word 'requirement' in the message is only true with the DANDI config, not in general (anyone can make a Subject object without anything more than an ID) - and making config-dependent error messages would be more complicated than simply correcting the wordage here

> a better way to describe what the user should do (provide one or the other) rather than merely describe the status of the fields being evaluated by the Inspector.

The phrasing across the Inspector is consistently the latter; it only reports what is missing, or what structures are written incorrectly; occasionally it will mention how to fix it, but for existence checks I'd it is pretty self-evident that if the Inspector says something is missing, the way to get it to dismiss the message is to make it not missing

In this special case there are two things that are each missing in order to trigger the message, and the incorrect message is to say that both are missing and therefore both need to be specified to make the message go away

Anyway, I added an extra clarification to the message without making it seem like a hard requirement

garrettmflynn:
Ah thank you for the clarification about the difference between _when_ these requirements hold.

With your clarification, though, I think it's clearer to keep the `and` over the `or` since that's the actual structure of the data when it comes in. But the extra sentence helps!

src/nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -132,11 +132,10 @@ def check_doi_publications(nwbfile: NWBFile):
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=Subject)
 def check_subject_age(subject: Subject):
     """Check if the Subject age is in ISO 8601 or our extension of it for ranges."""
-    if subject.age is None:
-        if subject.date_of_birth is None:
-            return InspectorMessage(message="Subject is missing age and date_of_birth.")
-        else:
-            return
+    if subject.age is None and subject.date_of_birth is None:
+        return InspectorMessage(message="Subject is missing age or date_of_birth. Please specify at least one.")
```
CodyCBakerPhD:
```suggestion
        return InspectorMessage(message="Subject is missing age and date_of_birth. Please specify at least one of these fields.")
```

tests/unit_tests/test_nwbfile_metadata.py
```diff
@@ -336,7 +336,7 @@ def test_pass_check_subject_age_with_dob():
 def test_check_subject_age_missing():
     subject = Subject(subject_id="001")
     assert check_subject_age(subject) == InspectorMessage(
-        message="Subject is missing age and date_of_birth.",
+        message="Subject is missing age or date_of_birth. Please specify at least one.",
```
CodyCBakerPhD:
```suggestion
        message="Subject is missing age and date_of_birth. Please specify at least one of these fields.",
```

CodyCBakerPhD:
@garrettmflynn OK, back to the old message plus an extra response on how to resolve

garrettmflynn (approved):
Looks good to me!

PR #425, opened 2023-12-19
Use paths instead of io in pynwb.validate call

rly:
Addresses issue in https://github.com/NeurodataWithoutBorders/pynwb/issues/1777

In short, `validate(paths=[nwbfile_path])` performs the expected behavior of validating against all cached namespaces in the file except for dependent ones and `validate(io=io)` does not. Changing `validate(io=io)` to perform the expected behavior would be a breaking change in pynwb. I think we should make that breaking change in pynwb as soon as possible, but given the urgency of the issue and that `validate(paths=[nwbfile_path])` does the expected behavior, I think we should also change nwbinspector to use `validate(paths=[nwbfile_path])` now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Added `check_rate_is_not_zero` for ensuring non-zero rate value of `TimeSeries` that has more than one frame. [#389](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/389)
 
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Ohhh I see - good idea! Thanks for the quick fix

PR #433, opened 2024-02-02
BF: do or to "" to ensure that we do not end up treating None as str

yarikoptic:
We started to get a test in dandi-cli to error out

```
E     +     '"/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/nwbinspector/nwbinspector.py", '
E     +     'line 632, in _intercept_in_vitro_protein\n'
E     +     '    and getattr(subject, "subject_id", "").startswith("protein")\n'
E     +     '        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n'
E     +     "AttributeError: 'NoneType' object has no attribute 'startswith'\n",
```

I didn't test yet but this I think should fix it

PR #434, opened 2024-02-02
Remove tox from dandi dev tests

CodyCBakerPhD:
## Motivation

From https://github.com/NeurodataWithoutBorders/nwbinspector/pull/433#issuecomment-1924676524


CodyCBakerPhD:
Confirmed that we're now correctly testing dev Inspector against dev DANDI testing suite, and confirmed fix from #433

@yarikoptic Sorry for the bother, this shouldn't happen again

@jwodder Thanks for the pointer on tox; it was even written in comments inside that action, but guess we missed that detail when following directions from https://github.com/dandi/dandi-cli/blob/master/DEVELOPMENT.md#running-tests-locally

PR #435, opened 2024-02-13
Separate workflows for testing dandi dev between live and no network

CodyCBakerPhD:
## Motivation

DANDI dev tests can sometimes fail due to live service issues like http://purl.obolibrary.org/ failing to respond (see https://github.com/dandi/dandi-cli/issues/1405)

We should definitely still run all tests; but I'd prefer a separation so that when I get a notification of a failure, I can do a direct comparison between with network and without network, indicating that I should just wait a certain amount of time before rerunning the with network ones

### DIFF
--- a/.github/workflows/dandi_release.yml
+++ b/.github/workflows/dandi_release.yml
@@ -1,5 +1,8 @@
-name: Test against latest DANDI release
+name: Test against latest DANDI release - no network
 on: workflow_call
+
+env:
+  DANDI_TESTS_NONETWORK: "1"
 
 jobs:
   build-and-test:
--- a/.github/workflows/dailies.yml
+++ b/.github/workflows/dailies.yml
@@ -3,7 +3,7 @@ name: Daily workflows
 on:
   workflow_dispatch:
   schedule:
-    - cron: "0 16 * * *"  # Daily at noon EST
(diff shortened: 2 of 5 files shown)


CodyCBakerPhD:
Log run: https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/7904129986/job/21573692473

shows successful separation and passing tests on non-network DANDI tests

.github/workflows/dandi_release.yml
```diff
@@ -1,6 +1,9 @@
-name: Test against latest DANDI release
+name: Test against latest DANDI release - no network
 on: workflow_call
 
+env:
+  DANDI_TESTS_NONETWORK: "1"
```
weiglszonja, line 5: Maybe it's just me, but at first I always read as "none twork" 😅 
```suggestion
  DANDI_TESTS_NO_NETWORK: "1"
```

CodyCBakerPhD: While I completely agree - this is the name of the environment variable that must be set to have this effect on the DANDI testing suite, so I cannot change it here

weiglszonja: ah okay, thanks for the clarification!

weiglszonja (approved):

PR #447, opened 2024-04-19
Method to import checks from any known extensions

CodyCBakerPhD:
## Motivation

One idea for removing the need to use `--modules ndx_multichannel_volume` every time

cc @dysprague - can you tell us if this works on your example? (you will probably have to do `pip install git+https://github.com/neurodatawithoutborders/nwbinspector.git@search_for_extension_checks`


CodyCBakerPhD:
From the discussion, we will go for a semi-automated opt-in only approach handled via config files

User trusted extensions would be registered via

```
nwbinspector register ndx-multichannel-volume
```

which would add that extension name to the config file such that any subsequent calls would automatically import that module, triggered the import of associated registered checks, without having to call `--modules ndx-multichannel-volume` explicitly every single time

bendichter:
We also might want to add `nwbinspector unregister` and `nwbinspector list`

PR #450, opened 2024-04-25
Pin mac runners to x64

CodyCBakerPhD:
Mac runners were just auto upgraded to x64; pinning to previous version here and will add arm64 tests in followup

### DIFF
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -33,6 +33,8 @@ jobs:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
     uses: ./.github/workflows/testing.yml
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   run-streaming-tests:
     needs: assess-file-changes
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -27,42 +27,42 @@ jobs:
           0
 
   run-doc-link-checks:
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/doc-link-checks.yml@dev
+    uses: ./.github/workflows/doc-link-checks.yml
 
   run-tests:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/testing.yml@dev
+    uses: ./.github/workflows/testing.yml
 
   run-streaming-tests:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/streaming-tests.yml@dev
+    uses: ./.github/workflows/streaming-tests.yml
 
   run-past-gallery:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/version-gallery.yml@dev
+    uses: ./.github/workflows/version-gallery.yml
 
   run-dev-gallery:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/dev-gallery.yml@dev
+    uses: ./.github/workflows/dev-gallery.yml
 
   test-dandi-latest:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/dandi_release.yml@dev
+    uses: ./.github/workflows/dandi_release.yml
 
   test-dandi-dev:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/dandi_dev.yml@dev
+    uses: ./.github/workflows/dandi_dev.yml
 
   test-dandi-dev-live:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/dandi_dev_live_service.yml@dev
+    uses: ./.github/workflows/dandi_dev_live_service.yml
 
   check-final-status:
     name: All tests passing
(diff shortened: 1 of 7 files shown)


.github/workflows/deploy-tests.yml
```diff
@@ -27,42 +27,42 @@ jobs:
           0
 
   run-doc-link-checks:
-    uses: neurodatawithoutborders/nwbinspector/.github/workflows/doc-link-checks.yml@dev
+    uses: ./.github/workflows/doc-link-checks.yml
```
CodyCBakerPhD, line 30: GitHub also started allowing relative paths to these sometime in the last year - making it a lot more convenient

.github/workflows/testing.yml
```diff
@@ -8,14 +8,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.8", "3.11"]
+        os: ["ubuntu-latest", "macos-13", "windows-latest"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
     steps:
-      - uses: s-weigand/setup-conda@v1
```
CodyCBakerPhD, line 14: `setup-conda` hasn't been updated in forever and is far less used than the official miniconda action

.github/workflows/testing.yml
```diff
@@ -8,14 +8,13 @@ jobs:
...
+      - uses: conda-incubator/setup-miniconda@v3
```
CodyCBakerPhD, line 18: `setup-conda` hasn't been updated in forever and is far less used than the official miniconda action

.github/workflows/testing.yml
```diff
@@ -26,12 +26,12 @@ jobs:
 
       - name: Install package
         run: |
-          pip install -e .
+          pip install .
```
CodyCBakerPhD, line 33: `-e` has been deprecated

.github/workflows/testing.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: Run pytest with coverage
         run: |
-          pytest -rsx --cov=./ --cov-report xml:./nwbinspector/nwbinspector/coverage.xml
+          pytest -rsx --cov=nwbinspector --cov-report xml:./nwbinspector/nwbinspector/coverage.xml
```
CodyCBakerPhD: Need to adjust name here to track coverage (used to rely on `-e`)

.github/workflows/testing.yml
```diff
@@ -26,12 +26,12 @@ jobs:
...
       - name: Install package
         run: |
-          pip install -e .
+          pip install .
           pip install dandi
       - name: Download testing data and set config path
         run: |
           dandi download "https://gui-staging.dandiarchive.org/#/dandiset/204919"
-          python -c "from nwbinspector.testing import update_testing_config; update_testing_config(key='LOCAL_PATH', value='./204919/testing_files/')"
+          python -c "from nwbinspector.testing import update_testing_config; update_testing_config(key='LOCAL_PATH', value='{nwbinspector.__file__}/204919/testing_files/')"
```
CodyCBakerPhD: Another consequence of not using `-e`

CodyCBakerPhD:
Various other things going wrong now as side effects

putting into draft

.github/workflows/streaming-tests.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.8", "3.11"]
+        os: ["ubuntu-latest", "windows-latest"]  # TODO: update mac and streaming methods
```
CodyCBakerPhD, line 11: Passing the buck here to a follow-up; swapping the conda activate action messes up the h5py install somehow, too much to figure out here

.github/workflows/testing.yml
```diff
@@ -8,14 +12,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.8", "3.11"]
+        os: ["ubuntu-latest", "macos-13", "windows-latest"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
```
CodyCBakerPhD, line 16: Action runners are easy to come by nowadays, and our suite is so light that I think it's fine to test against all versions

.github/workflows/testing.yml
```diff
@@ -43,14 +47,13 @@ jobs:
 
       - name: Run pytest with coverage
         run: |
-          pytest -rsx --cov=./ --cov-report xml:./nwbinspector/nwbinspector/coverage.xml
+          pytest -rsx --cov=nwbinspector --cov-report xml:./coverage.xml
           pytest tests/ -rsx tests/read_nwbfile_tests.py  # TODO when read_nwbfile is ported
       - if: ${{ matrix.python-version == '3.10' && matrix.os == 'ubuntu-latest' }}
         name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
```
CodyCBakerPhD, line 54: Bumpbing to v4 required some other changes, including a new token

CodyCBakerPhD:
@oruebel OK this is ready for review now

A long cascade of testing changes triggered by the big `macos` arm64 bump and revealed many items that were in need of maintenance over the past couple of years

.github/workflows/testing.yml
rly: When was `-e` of pip deprecated?

CodyCBakerPhD: Not sure - originally heard this from @h-mayorquin when we updated NeuroConv similarly

It's thrown a deprecation warning in console for a long time - this was the first time when trying it lead to an actual error in CI though

But reading into https://packaging.python.org/en/latest/discussions/setup-py-deprecated/ it seems it's not the method itself that's deprecated but rather the route through the CLI - so maybe that could still work

However, I'd rather swap the more official `pyproject.toml` approach anyhow in a follow-up since that is the most recommended way in 2024

rly: Oh I see. `pip install -e` is not deprecated with `pyproject.toml`, so the changes around removing `-e` could be resolved by using `pyproject.toml`.

h-mayorquin: Editable mode is a setuptools feature that might not be availabe in other installation backends (e.g. poetry).  pip just relies to the instructions to the backend:

https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs

But I think most backends implement this. Now, the real problematic part is that editable mode in setuptools changed dramatically on recent versions:

https://setuptools.pypa.io/en/latest/userguide/development_mode.html

My suggestion was that we should not use that API in the CI because it is unstable (within setuptools versions and across build backends implementations) and I suspected it was generating an error (I don't remember if I was right). This is in line what what setuptools suggest themselves (see the attention section in top of limitations here):
https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior

rly: Interesting. Thanks for the additional information @h-mayorquin ! In the pynwb/hdmf stack, we use `pip install -e 
 . ` to help `coverage` find the source files, but we can do it differently as @CodyCBakerPhD has done here. We'll make the same changes on the pynwb/hdmf stack.

rly (approved):

PR #458, opened 2024-05-03
add check for missing timezone. 

bendichter:
Also add best practice docs and tests for check


tests/unit_tests/test_nwbfile_metadata.py
```diff
@@ -56,6 +59,30 @@ def test_check_session_start_time_future_date_pass():
...
+        session_description="",
+        identifier=str(uuid4()),
+        session_start_time=datetime(2010, 1, 1, 0, 0, 0, 0, timezone.utc),
+    )
+    assert check_session_start_time_contains_time_zone(nwbfile) is None
+
+
+@pytest.mark.skipif(pynwb.__version__ < "2.7.0", reason="Feature not supported in pynwb < 2.7.0")
+def test_check_session_start_time_contains_time_zone_fail():
+    session_start_time = datetime(2010, 1, 1, 0, 0, 0, 0)
```
CodyCBakerPhD, line 84: Should there be a skip for it it's a `date` instead of `datetime`? If hourly precision isn't even known, I don't think the timezone matters a whole lot

bendichter:
It looks like this is (edit: were) failing because there is a workflow that explicitly tests older versions of pynwb and this fails in that case since the feature is only supported in 2.7.0+. What should we do here? Here are some options:
1. Add a `skipif` for the offending tests. I have implemented this here. This works for getting past the error, but the check will malfunction. That is not so bad in this particular case, since the error is erroneously silent, which is fine here since the user would not be able to fix that issue with their version of PyNWB anyway.
2. Since NWB Inspector is about best practices, it may make sense to have strict requirements about only using the latest PyNWB. This makes sense from a best practices perspective, but may not be practical for everyone. For example, the Allen Institute has in the past found it difficult to easily migrate to new environments.
3. We could introduce a `skipif` arg to the `@register_check` decorator, which would allow us to explicitly label checks as pertaining to certain conditions or configurations. This would solve the issue, but would fail to notify users that they could improve best practices if they upgrade pynwb. I suppose we could get around this downside by adding a check specifically for pynwb version, ensuring it is the latest release.

tests/unit_tests/test_nwbfile_metadata.py
bendichter: I don't think dates can contain timezones

src/nwbinspector/checks/nwbfile_metadata.py
```diff
@@ -36,6 +36,20 @@ def check_session_start_time_old_date(nwbfile: NWBFile):
...
 
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=NWBFile)
+def check_session_start_time_contains_time_zone(nwbfile: NWBFile):
+    """
+    Check if the session_start_time contains a time zone.
+
+    Best Practice: :ref:`best_practice_global_time_reference`
+    """
+    session_start_time = nwbfile.session_start_time
+    if session_start_time.tzinfo is None:
```
bendichter, line 49:
```suggestion
    if isinstance(session_start_time, datetime) and session_start_time.tzinfo is None:
```

bendichter: @CodyCBakerPhD , does this address your comment below?

bendichter:
Another question about this is how this would be integrated with NWB GUIDE. Does the GUIDE have a way to enter timezones yet? If not, we are setting users up for a best practice warning they are not able to address.

CodyCBakerPhD:
> It looks like this is (edit: were) failing because there is a workflow that explicitly tests older versions of pynwb and this fails in that case since the feature is only supported in 2.7.0+. What should we do here?

Yeah this has been a long-time question about how the Inspector should interact with PyNWB versions in general (in addition to cached schema within each file??): https://github.com/NeurodataWithoutBorders/nwbinspector/issues/263

Posted your ideas on that thread

Skipping per test is fine for now and has been the way forward for the time being

CodyCBakerPhD:
> Another question about this is how this would be integrated with NWB GUIDE. Does the GUIDE have a way to enter timezones yet? If not, we are setting users up for a best practice warning they are not able to address.

It looks like it still automatically sets based on computer clock

I think we were partially waiting on finding a better timestamp widget, but @garrettmflynn for the time being can you add a simple dropdown selector for $\pm \ X$ ($X \in \{0, ... 12\}$) to at least allow it? We can polish UX over time

src/nwbinspector/checks/nwbfile_metadata.py
CodyCBakerPhD: It does - specifically, it seems `datetime.date` does not even have the `tzinfo` attribute otherwise

Can you add a test for that skip case?

bendichter:
Codespell:

```
Error: ./tests/test_register_check.py:17: assertIn ==> asserting, assert in, assertion
```

Is codespell checking the spelling of code? How can we get it to skip this?

CodyCBakerPhD:
> Is codespell checking the spelling of code? How can we get it to skip this?

Hah, looks like an issue for very recent versions of codespell - already some others have noticed: https://github.com/codespell-project/codespell/issues/3430

Doing a local ignore now

.github/workflows/codespell.yml
CodyCBakerPhD, line 1: Also discovered that we added this to pre-commit at some point: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/458/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L13

but didn't remove the GitHub action

CodyCBakerPhD:
More pressing issue is some rather strange `datetime` comparison error that for whatever reason is fine on Windows but not elsewhere: https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/9214133616/job/25349652811?pr=458#step:12:232

bendichter:
@CodyCBakerPhD this is probably because `datetime` is a subclass of `date`. I personally find this counter-intuitive. Maybe it's not true on Windows?

```python
import datetime
isinstance(datetime.datetime.now(), datetime.date)
```
```
True
```

CodyCBakerPhD:
It's true on Windows. Very odd

PR #460, opened 2024-05-08
Pin Zarr for tests

CodyCBakerPhD:
Analog of https://github.com/catalystneuro/neuroconv/pull/845 for here as well

Tests just failed today due to a surprise break from Zarr release yesterday - some issue with decoding Blosc

Shouldn't effect actual typical usage of the backend feature aside from Blosc I assume, since none of the other compressor tests complain

https://github.com/hdmf-dev/hdmf-zarr/issues/192 so team knows about it, and if it's an issue with the way they specify the read from the IO level (I doubt it, but never know)

Downpinning in tests temporarily to fix the CI, will check over time if resolved automagically

### DIFF
--- a/.github/workflows/dev-gallery.yml
+++ b/.github/workflows/dev-gallery.yml
@@ -22,9 +22,8 @@ jobs:
           pip install pytest-cov
       - name: Install package
         run: |
-          pip install -e .
+          pip install -e ".[dandi]"
           pip install git+https://github.com/neurodatawithoutborders/pynwb@dev
-          pip install dandi
       - name: Download testing data and set config path
         run: |
           dandi download "https://gui-staging.dandiarchive.org/#/dandiset/204919"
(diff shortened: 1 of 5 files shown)


rly:
@CodyCBakerPhD I fixed the tests so that zarr has an upper bound when the dandi extra is installed and `pip install ".[dandi]"` is used instead of `pip install dandi`. I'm not sure what's up with the pynwb 2.0 test failing. OK if that fails?

CodyCBakerPhD:
I believe there was an old reason (so that joint installation requirements are not resolved simultaneously) but since that's working now OK to switch back

CodyCBakerPhD:
>  I'm not sure what's up with the pynwb 2.0 test failing. OK if that fails?

And sure, that's fine. Need to talk about it next LBNL meeting but most likely we'll stop supporting previous PyNWB versions with latest Inspector version

rly (approved):

PR #462, opened 2024-05-21
this check is missing

bendichter:
I also don't know how we would check this automatically


CodyCBakerPhD:
The name of the check is now `check_image_series_data_size` in the `image_series` submodule: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/checks/image_series.py#L55C5-L55C33

It allows the data to be written internally (with `gzip` or others) up to a threshold of 20GB - only applies to HDF5 at the moment

This best practice would also change when using the Zarr backend (I think recommendation would to use Jeremy's MP4 codec)

bendichter:
Let's look at updating this check to make sure it applies specifically to behavioral videos
* It should also apply to OnePhotonSeries
* The error message should specifically indicate behavioral videos.

CodyCBakerPhD (approved):

PR #475, opened 2024-06-17
Update testing.yml to install h5py with ROS3 first

rly:
If you install `h5py` from conda-forge to get ROS3 support before installing `nwbinspector`, `h5py` does not get re-installed, so there is no need to uninstall it and create conflicting conda/pip issues later.


.github/workflows/testing.yml
```diff
@@ -28,20 +28,14 @@ jobs:
...
         run: pip install ".[dandi]"
       - name: Download testing data and set config path
         run: |
           dandi download "https://gui-staging.dandiarchive.org/#/dandiset/204919"
           python -c "from nwbinspector.testing import update_testing_config; update_testing_config(key='LOCAL_PATH', value='./204919/testing_files/')"
-      - name: Uninstall h5py
-        run: pip uninstall -y h5py
-      - name: Install ROS3
-        run: conda install -c conda-forge h5py
-      - name: Install HDMF-zarr # temporary
```
rly, line 41: I think these temporary reinstalls from last year are no longer necessary (from https://github.com/NeurodataWithoutBorders/nwbinspector/pull/388#pullrequestreview-1580890623)

PR #480, opened 2024-07-30
[New Check]: Check timestamps ascending with nans

### LINKED ISSUE #476: [Bug]: check_timestamps_ascending is triggered when timestamps have NaNs
### What happened?

I am doing the Reimer Lab conversion and we have a timestamps array that contains NaNs. I just decided to leave them in. Then when running the inspector, this triggers the check_timestamps_ascending check, even though all the non-NaN values are ascending. I think the timestamps array should be flagged by the inspector, but this should be its own check, and if the defined values are ascending it should not trigger the check_timestamps_ascending check.

### Operating System

macOS

### Python Version

3.9

No

### Package Versions

_No response_

### Code of Conduct

alessandratrapani:
## Motivation
Fix #476

### DIFF
--- a/src/nwbinspector/checks/time_series.py
+++ b/src/nwbinspector/checks/time_series.py
@@ -97,7 +97,7 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
 def check_timestamps_with_nans(time_series: TimeSeries):
     """Check if there are NaN values in the timestamps array."""
     if time_series.timestamps is not None and np.isnan(time_series.timestamps).any():
-        return InspectorMessage(f"{time_series.name} timestamps contain NaN values.")
+        return InspectorMessage(message=f"{time_series.name} timestamps contain NaN values.")
 
 
 @register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=TimeSeries)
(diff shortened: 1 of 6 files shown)


src/nwbinspector/checks/time_series.py
```diff
@@ -93,6 +93,13 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
         return InspectorMessage(f"{time_series.name} timestamps are not ascending.")
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
+def check_timestamps_with_nans(time_series: TimeSeries):
```
bendichter: For these types of checks, we generally also include an optional arg `nelems` to allow for this check to only read a subset of the Dataset. This is there in case the datasets are huge and take too much time to read all the data.

bendichter:
@alessandratrapani this looks great! Can you also please add a section in the best practices document?

src/nwbinspector/checks/time_series.py
```diff
@@ -93,6 +93,13 @@ def check_timestamps_ascending(time_series: TimeSeries, nelems=200):
...
+    """Check if there are NaN values in the timestamps array."""
+    if time_series.timestamps is not None and np.isnan(time_series.timestamps).any():
+        return InspectorMessage(f"{time_series.name} timestamps contain NaN values.")
```
CodyCBakerPhD:
```suggestion
        return InspectorMessage(message=f"{time_series.name} timestamps contain NaN values.")
```

CodyCBakerPhD (approved):

PR #486, opened 2024-08-14
Implement scope control over check submodule

CodyCBakerPhD:
## Motivation

@stephprince After thinking about several alternatives, I do actually believe that this PR is the most Pythonic approach to the checks submodule. It maintains a fairly flat structure (rather than exposing submodules for each and every category of check) and is explicit about what gets imported and exposed

We've had some issues in the past, both with actual registration (in `available_checks`) and with the API docs where newly added checks didn't actually get included - this requirement makes it very clear what needs to occur in order for that to happen now

I also have ideas for improving the check testing infrastructure to ensure developers are notified right away on a PR if they forget to expose their newly created check function


CodyCBakerPhD:
@stephprince This is also a case where I'd say the CHANGELOG reminder can be ignored since the previous PR entries essentially describe this change

src/nwbinspector/checks/ecephys/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+    "All submodules below the level 'nwbinspector.checks'  have been deprecated. "
+    "Please retrieve the specific check function(s) you wish to use from the `available_checks` registry "
+    "at the top-level or import directly from 'nwbinspector.checks'."
+)
+
+warnings.warn(message=message, category=DeprecationWarning, stacklevel=2)
+
+# Still keep imports functional with warning for soft deprecation cycle
+# TODO: remove after 9/15/2024
+from .._behavior import *
```
stephprince:
```suggestion
from .._ecephys import *
```

I think all of the `__init__.py` files need to be updated, they are all currently importing from the behavior submodule on the L13

CodyCBakerPhD: @stephprince Great catch! Fixed

src/nwbinspector/checks/behavior/__init__.py
```diff
@@ -0,0 +1,13 @@
+import warnings
+
+message = (
+    "All submodules below the level 'nwbinspector.checks'  have been deprecated. "
+    "Please retrieve the specific check function(s) you wish to use from the `available_checks` registry "
```
stephprince, line 5: Should the `default_check_registry` be recommended  in this case instead of the `available_checks` list since a user would likely want to retrieve the specific check function(s) with the check name?

stephprince (commented):
Added a small comment but otherwise looks good to me!

Just so I understand clearly, when new checks get added the process will be:
1. add the check to the appropriate submodule
2. add the import to the `checks/__init__.py` file from that submodule
3. add the check name to the `__all__` list in the `checks/__init__.py`

CodyCBakerPhD:
> add the check to the appropriate submodule
add the import to the checks/__init__.py file from that submodule
add the check name to the __all__ list in the checks/__init__.py

Precisely. Not too onerous IMO - GitHub copilot also does a fantastic job autocompleting these structures

src/nwbinspector/checks/behavior/__init__.py
CodyCBakerPhD: That is an interesting question, keep it in mind going forward

I have other ideas for how to expose listings of checks that are distinct from the raw registry (since interactions with that underly the source of multiple issues), which I might even swap to be privately managed and not exposed

stephprince (approved):

PR #487, opened 2024-08-20
Final scoping fix

CodyCBakerPhD:
## Motivation

@stephprince My bad; we had previously fixed the `__init__.py` for this submodule to the new standard but forgot to mark the lower levels as private with soft deprecations

This should be the last one like this unless my eyes missed something else


stephprince (approved):
Sounds good! I also looked through and I think they're all fixed now.

PR #494, opened 2024-08-30
Fix import

bendichter:
## Motivation

fix #493

PR #495, opened 2024-08-31
Add cli with config tests

CodyCBakerPhD:
## Motivation

Turns out we actually didn't have a CLI test for the `--config dandi` case

Added here, will fail until #494 is merged (at which point it should then pass)

### DIFF
--- a/tests/test_inspector.py
+++ b/tests/test_inspector.py
@@ -3,11 +3,9 @@ from shutil import rmtree
 from tempfile import mkdtemp
 from pathlib import Path
 from unittest import TestCase
-from unittest.mock import patch
 from datetime import datetime
 
 import numpy as np
-import pytest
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
Proof of intended failure

image

tests/test_inspector.py
```diff
@@ -86,6 +88,44 @@ def add_simple_table(nwbfile: NWBFile):
 
 
 class TestInspector(TestCase):
+    """A common helper class for testing the NWBInspector."""
```
CodyCBakerPhD, line 89: Peeled out this class so the helpers for asserting log file contents can be used across the suite

Might be worth moving to `nwbinspector.testing` in a follow-up

bendichter (approved):

PR #496, opened 2024-08-31
Setup initial ruff

CodyCBakerPhD:
## Motivation

This is also a general way, without relying on testing suite, to catch any potential issues like those in #494 

Plan over next month was also to improve automatic linting and isorting in this manner, but easiest sometimes to add one selection at a time

### DIFF
--- a/src/nwbinspector/__init__.py
+++ b/src/nwbinspector/__init__.py
@@ -29,6 +29,7 @@ __all__ = [
     "Importance",
     "Severity",
     "InspectorMessage",
+    "validate_config",
     "load_config",
     "configure_checks",
     "InspectorOutputJSONEncoder",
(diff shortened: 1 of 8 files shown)


src/nwbinspector/__init__.py
```diff
@@ -29,6 +29,7 @@
     "Importance",
     "Severity",
     "InspectorMessage",
+    "validate_config",
```
CodyCBakerPhD, line 32: This caught one more

src/nwbinspector/_registration.py
```diff
@@ -1,9 +1,6 @@
 """Primary decorator used on a check function to add it to the registry and automatically parse its output."""
 
-from collections.abc import Iterable
```
CodyCBakerPhD, line 3: And cleaned up many others (unused imports)

bendichter (approved):

PR #505, opened 2024-09-06
Final deprecation of `insect_nwb`

CodyCBakerPhD:
Really making use of the 'or after' clause on this one


stephprince (commented):
> Really making use of the 'or after' clause on this one

Better late than never! Looks good - I found a couple of other locations that reference `inspect_nwb` that I think need to be updated as well: 
- `docs/user_guide/using_the_library.rst`
- `docs/user_guide/using_the_library_advanced.rst` 
- `src/nwbinspector/nwbinspector/__init__.py`

CodyCBakerPhD:
@stephprince Sharp eye!

When you add documentation tests, it might be nice to make them 'nitpicky' (throw error on warning, such as deprecations) to automatically catch such things

Removed those items

stephprince (approved):

PR #506, opened 2024-09-06
Remove robust_ros3_retry

CodyCBakerPhD:
Removed the ROS3 retry helper since it is no longer needed


src/nwbinspector/_nwb_inspection.py
```diff
@@ -145,8 +144,17 @@ def inspect_all(
...
         with pynwb.NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True) as io:
-            nwbfile = robust_s3_read(io.read)
-            identifiers[nwbfile.identifier].append(nwbfile_path)
+            try:
+                nwbfile = io.read()
+                identifiers[nwbfile.identifier].append(nwbfile_path)
+            except Exception as exception:
+                yield InspectorMessage(
+                    message=traceback.format_exc(),
+                    importance=Importance.ERROR,
```
CodyCBakerPhD, line 153: Also adding this in here now - it's a known problem that this pre-check can fail to collect and report messages if is read error occurs

Very hard to add tests for ATM though, might try in a follow-up that improves testing all around

src/nwbinspector/_nwb_inspection.py
```diff
@@ -458,22 +466,30 @@ def run_checks(
...
-                        output.importance = check_function.importance
-                    yield output
-                elif output is not None:
-                    for x in output:
-                        x.importance = check_function.importance
-                        yield x
+            if check_function.neurodata_type is not None and not issubclass(
+                type(nwbfile_object), check_function.neurodata_type
+            ):
+                continue
```
CodyCBakerPhD, line 440: Broke up this logic to reduce cyclomatic complexity

tests/unit_tests/test_time_series.py
```diff
@@ -371,12 +372,10 @@ def test_check_none_matnwb_resolution_pass():
...
-        nwbfile = robust_s3_read(command=io.read)
-        time_series = robust_s3_read(
-            command=nwbfile.processing["video_files"]["video"].time_series.get,
-            command_args=["20170203_KIB_01_s1.1.h264"],
-        )
-    assert check_resolution(time_series) is None
+        nwbfile = io.read()
+        time_series = nwbfile.processing["video_files"]["video"]["20170203_KIB_01_s1.1.h264"]
+
+        assert check_resolution(time_series=time_series) is None
```
CodyCBakerPhD, line 378: This one will just be removed in a follow-up. Not too concerned about it now; has to do with old versions of packages

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Upcoming
 
+### Improvements
+* Removed the `robust_ros3_retry` utility helper. [#506](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/506)
```
```suggestion
* Removed the `robust_ros3_read` utility helper. [#506](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/506)
```

src/nwbinspector/_nwb_inspection.py
stephprince: Sounds good. To clarify, at this point streaming is being redirected to `inspect_dandiset` earlier in `inspect_all`, so this except here is to catch local read errors?

stephprince (commented):
Looks good to me!

src/nwbinspector/_nwb_inspection.py
CodyCBakerPhD: Well, the check itself is a special one looking for duplicated file identifiers. That can't be done in an easy way as a normal check function, so it is done here before main deployment of the check registry

And yes, only applies to local folder contents. Have not implemented a Dandiset way of doing it yet

stephprince: Ah ok got it, thanks for explaining!

stephprince (approved):

PR #507, opened 2024-09-06
Swap to pyproject.toml setup

CodyCBakerPhD:
Modern Pythonic practice has moved away from `setup.py` +/- `.cfg` towards `pyproject.toml`

This updates the package accordingly


docs/.readthedocs.yaml
CodyCBakerPhD, line 1: I copied over these two RTD files to start a smooth transition to keeping the files there instead of the outer level to reduce clutter

It works fine as long as the settings on the RTD admin page are adjusted, which is why I will wait until after this PR is merged to do that and then remove the files still at the outer level (so it doesn't break on the other current PRs)

README.md
```diff
@@ -1,18 +1,40 @@
-<img src="docs/logo/logo.png" width="300">
+<p align="center">
+  <img src="https://raw.githubusercontent.com/neurodatawithoutborders/nwbinspector/dev/docs/logo/logo.png" width="250" alt="NWB Inspector logo"/>
```
CodyCBakerPhD, line 2: This change should fix the issue with the logo not rendering on PyPI: https://pypi.org/project/nwbinspector/

README.md
```diff
@@ -1,18 +1,40 @@
...
+  </p>
+  <p align="center">
+    <a href="https://github.com/psf/black"><img alt="Python code style: Black" src="https://img.shields.io/badge/python_code_style-black-000000.svg"></a>
+    <a href="https://github.com/astral-sh/ruff"><img alt="Python code style: Ruff" src="https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json"></a>
+  </p>
+  <p align="center">
+    <a href="https://nwbinspector.readthedocs.io/"><img alt="Documentation build status" src="https://readthedocs.org/projects/nwbinspector/badge/?version=dev"></a>
+    <a href="https://github.com/NeurodataWithoutBorders/nwbinspector/actions/workflows/dailies.yml/badge.svg"><img alt="Daily tests" src="https://github.com/NeurodataWithoutBorders/nwbinspector/actions/workflows/dailies.yml/badge.svg"></a>
+  </p>
+</p>
```
CodyCBakerPhD, line 20: These are stylistic changes to the front page

pyproject.toml
```diff
@@ -1,3 +1,67 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Operating System :: POSIX :: Linux",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "License :: OSI Approved :: BSD License",
+]
+requires-python = ">=3.9"
```
CodyCBakerPhD, line 33: This also removes Python 3.8 support, which ends its security upgrades next month

CodyCBakerPhD:
This will take a bit more work - will splinter a few items into smaller PRs

CodyCBakerPhD:
@stephprince OK this is ready now

I'll remove the RTD files on the outer level in a followup so the docs have a smooth transition

pyproject.toml
```diff
@@ -1,3 +1,67 @@
...
+    "requests",
+    "aiohttp",
+    "PyYAML",
+    "jsonschema",
+    "packaging",
+    "natsort",
+    "click",
+    "tqdm",
+    "isodate",
+    "numpy>=1.22.0,<2.0.0"  # TODO: remove 2.0 bound when HDMF supports
```
stephprince: I think hdmf should support numpy >2.0 now

```suggestion
    "numpy>=1.22.0"
```

pyproject.toml
```diff
@@ -1,3 +1,67 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Operating System :: POSIX :: Linux",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "License :: OSI Approved :: BSD License",
+]
+requires-python = ">=3.9"
+dependencies = [
+    "pynwb",
```
stephprince:
```suggestion
    "pynwb>=2.8",   # NWB Inspector should always be used with most recent minor versions of PyNWB
```

pyproject.toml
```diff
@@ -1,3 +1,67 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "nwbinspector"
+version = "0.6.0"
```
stephprince, line 7: not sure if the version was meant to be updated in this PR?

stephprince (approved):
Looks good to me. Added a couple small suggestions.

pyproject.toml
CodyCBakerPhD: Oh, this is good now? Thanks for checking

pyproject.toml
CodyCBakerPhD: It was. Just in case. (using a different build backend is a fundamental change; not per se one expected to break back compatibility, but it's fairly significant)

Maybe after the remaining big hard deprecations next March you might want to do an official v1.0.0 release

pyproject.toml
CodyCBakerPhD: Ah right, since that was merged first

Good catch

pyproject.toml
stephprince: That makes sense. And yes an official release after the big hard deprecations sounds like a good plan.

pyproject.toml
```diff
@@ -1,3 +1,67 @@
...
+    "requests",
+    "aiohttp",
+    "PyYAML",
+    "jsonschema",
+    "packaging",
+    "natsort",
+    "click",
+    "tqdm",
+    "isodate",
+    "numpy>=1.22.0"
```
CodyCBakerPhD:
```suggestion
    "numpy>=1.22.0,<2.0.0" # TODO: add compatibility for 2.0
```

Ahh looks like we still have some 2.0 incompatibilities local to Inspector here: https://github.com/NeurodataWithoutBorders/nwbinspector/actions/runs/10907985074/job/30272984050?pr=507#step:8:414

Will deal with this in a follow-up

PR #508, opened 2024-09-06
Remove 3.8; updating annotation typing

CodyCBakerPhD:
Pre-PR to #507 

Removes 3.8 and updates the relevant typing to 3.9+ standard (using built-in references instead of typing module classes)


stephprince (approved):
Looks good! Updated a couple extra files

PR #509, opened 2024-09-06
Simplify testing suite

CodyCBakerPhD:
While working on #507 noticed some needlessly complicated setup to a very particular corner of the testing suite

Replacing it all with a simple environment variable


stephprince (commented):
Looks good.

My only question is what is the case where the testing files get changed / need to be reuploaded to DANDI? Will that be done manually now since the `update-testing-files.yml` workflow was removed?

CodyCBakerPhD:
> My only question is what is the case where the testing files get changed / need to be reuploaded to DANDI? Will that be done manually now since the update-testing-files.yml workflow was removed?

While we once had aspirations for it to be all automated, the reality is that process will continue as it always has - manually

Manually create a specific (usually old) environment and run some code that only works in that specific old environment in order to create the 'bad' example file for testing

But even that should really just be for new 'bad' examples - the previous ones can just be downloaded from DANDI and shouldn't really ever need to be replaced for any reason I know of

stephprince:
> But even that should really just be for new 'bad' examples - the previous ones can just be downloaded from DANDI and shouldn't really ever need to be replaced for any reason I know of

Ok sounds good. Follow up question - since the nwbinspector test file dandiset is on the staging server (which from my understanding is not guaranteed to continue existing?), is there a backup of these files somewhere? I saw the dandiset said some files were from user submissions so I assume not all of them can be auto-generated from the package.

CodyCBakerPhD:
DANDI team would/should provide a big heads up on clearing any staging content or putting any such policy into place

At that point, if it happens, I'd recommend moving it to the main archive and publishing official persistent releases

stephprince (approved):

PR #265, opened 2022-09-20
[Do not merge until next PyNWB release]: Cached namespace validation

CodyCBakerPhD:
Related to PyNWB dev PR https://github.com/NeurodataWithoutBorders/pynwb/pull/1511 for allowing API access to validation using cached namespaces, this propagates it to the NWB Inspector validation capability and ensure it works correctly.

PR #510, opened 2024-09-06
Support only latest PyNWB

### LINKED ISSUE #263: [Bug]: Philosophy for interactions with PyNWB versions

CodyCBakerPhD:
As per previous discussion about what range of support NWB Inspector ought to have with past PyNWB/HDMF versions - we decided it should only support latest

This simplifies a lot of conditions on the core inspection logic, individual checks, and what the testing suite is responsible for checking

The only aspect that cannot be replaced is using old versions of PyNWB to generate example files with particular failures that are no longer possible to achieve in latest PyNWB - these are still good to check the inspection logic on since someone could write a file using an older version but inspect it in a different environment (DANDI validate/upload would outright require that) - so I'm keeping those helpers around


CodyCBakerPhD:
TODO in follow-up: re-implement #265

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 ### Improvements
 * Removed the `robust_ros3_read` utility helper. [#506](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/506)
+* Cleaned old references to non-recent PyNWB and HDMF versions. Current policy is that latest NWB Inspector releases should only support compatibility with latest PyNWB and HDMF. [#](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/)
```
```suggestion
* Cleaned old references to non-recent PyNWB and HDMF versions. Current policy is that latest NWB Inspector releases should only support compatibility with latest PyNWB and HDMF. [#510](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/510)
```

src/nwbinspector/_nwb_inspection.py
```diff
@@ -73,8 +71,8 @@ def inspect_all(
         all available CPUs minus x.
         Set to 1 (also the default) to disable.
     skip_validate : bool, optional
-        Skip the PyNWB validation step. This may be desired for older NWBFiles (< schema version v2.10).
```
stephprince, line 76: I think this is ok to keep as a comment here and below. 
From what I understand skipping the validation step for older NWBFiles is a different issue from using an older version of pynwb?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-pynwb
+pynwb>=2.8  # NWB Inspector should always be used with most recent minor versions of PyNWB
```
stephprince, line 1: Will this cause any difficulties when trying to generate example files with particular failures as in: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/4856bd65fa0650080b1ecb7f225596be6962208d/src/nwbinspector/testing/_testing.py#L90

maybe there should be a separate `requirements-dev.txt` with no minimum pynwb version specified?

stephprince (commented):
Looks good! Added a couple comments/questions.

CHANGELOG.md
CodyCBakerPhD: Thank you

requirements.txt
CodyCBakerPhD: When using that specific method, you can always do

```bash
pip install nwbinspector
pip install pynwb==2.1.0
```

as separate steps to obtain the desired result

Towards #507 there should be no excessive requirements files

src/nwbinspector/_nwb_inspection.py
CodyCBakerPhD: It's mostly that I don't want to make assumptions about why a user might want to skip validation - or in particular, give them a ready excuse to do so

Historically the main use is to prevent errors from occurring if there's some unexpected issue using an old PyNWB version on a newer file

But I trust modern PyNWB to either work on previous schema versions or otherwise trust the Inpspector to route an error to the right part of the messages

requirements.txt
stephprince: Following those `pip install` steps leads to some dependency conflict warnings, though the tests seem to run fine so that works.

stephprince (approved):

requirements.txt
CodyCBakerPhD: @stephprince Exactly, it's a warning rather than an error (if you do `pip install nwbinspector pynwb==2.1.0` or include `pynwb==2.1.0` in the requirements, it outright fails) but therefore lets you override it to 'force' the incompatability

PR #512, opened 2024-09-09
Remove dashboard CI

CodyCBakerPhD:
@stephprince I used to use this for my personal TODO list on CN side but since you're taking over this doesn't make sense to keep

Also LBNL dashboard is still here and working, so I assume y'all just use that


stephprince:
That makes sense, however I don't think any issues from this repo are currently getting added to the general NWB dashboard. 

If it's easy to swap in, could we replace it with something like this: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/.github/workflows/project_action.yml

CodyCBakerPhD:
Oh, I had thought we added that a while back

Could you take charge on adding the NWB one then? So that you know how it works and everything

stephprince (approved):

PR #520, opened 2024-09-27
Update all annotations and check with Mypy

### LINKED ISSUE #416: Fully type-annotate package
Running mypy on code that uses the latest version of `nwbinspector` fails with the error message "module is installed, but missing library stubs or py.typed marker".  A quick look at `nwbinspector`'s source indicates that it is making some use of type annotations, but a number of annotations are missing.

Please do the following:

- [x] Fully type-annotate all public items
- [x] Add a CI job to run mypy on the package, and ensure it passes
- [x] Add an empty file named `py.typed` inside `src/nwbinspector/`
- [x] Ensure that the `src/nwbinspector/py.typed` file is included in sdists & wheels built from the project.  This may involve configuration specific to your build backend (setuptools) that is beyond the scope of this comment.
    - You can list the contents of an sdist (the `dist/nwbinspector-{version}.tar.gz` file created when building the project) by running `tar ztf dist/nwbinspector-{version}.tar.gz` on it.
    - You can list the contents of a wheel (a `dist/nwbinspector-{version}-*-*-*.whl` file created when building the project) by running `zipinfo -1 dist/nwbinspector-{version}-*-*-*.whl` on it.

CodyCBakerPhD:
fix #416, at long last

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,7 +92,17 @@ extend-exclude = '''
 exclude = ["docs/*"]
 
 [tool.ruff.lint]
-select = ["F401", "F541", "I"] # TODO: eventually, expand to other 'D', and other 'F' linting
+# TODO: eventually, expand to other 'D', and other 'F' linting
+select = [
+    "F401",
+    "F541",
+    "I",
+    "ANN001",
+    "ANN201",
+    "ANN202",
+    "ANN205",
+    "ANN206",
+]
 fixable = ["ALL"]
 
 [tool.ruff.lint.per-file-ignores]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,6 +115,7 @@ fixable = ["ALL"]
 "src/nwbinspector/inspector_tools/__init__.py" = ["F401", "I"]
 "src/nwbinspector/version/__init__.py" = ["F401", "I"]
 "src/nwbinspector/register_checks/__init__.py" = ["F401", "I"]
+"tests/*" = ["ANN"]
 
 [tool.ruff.lint.isort]
 relative-imports-order = "closest-to-furthest"
--- a/src/nwbinspector/_configuration.py
+++ b/src/nwbinspector/_configuration.py
@@ -3,22 +3,20 @@
 import json
 from pathlib import Path
 from types import FunctionType
-from typing import Optional
+from typing import Optional, Union
 
 import jsonschema
 import yaml
+from typing_extensions import Callable
 
-from nwbinspector.utils import PathType
+from ._registration import Importance, available_checks
 
-from . import available_checks
-from ._registration import Importance
-
-INTERNAL_CONFIGS = dict(
+INTERNAL_CONFIGS: dict[str, Path] = dict(
     dandi=Path(__file__).parent / "_internal_configs" / "dandi.inspector_config.yaml",
 )
 
 
-def validate_config(config: dict):
+def validate_config(config: dict) -> None:
     """Validate an instance of configuration against the official schema."""
     config_schema_file_path = Path(__file__).parent / "_internal_configs" / "config.schema.json"
     with open(file=config_schema_file_path, mode="r") as fp:
@@ -26,7 +24,7 @@ def validate_config(config: dict):
     jsonschema.validate(instance=config, schema=schema)
 
 
-def _copy_function(function):
+def _copy_function(function: Callable) -> Callable:
     """Copy the core parts of a given function, excluding wrappers, then return a new function."""
     copied_function = FunctionType(
         function.__code__, function.__globals__, function.__name__, function.__defaults__, function.__closure__
@@ -34,10 +32,11 @@ def _copy_function(function):
 
     # in case f was given attrs (note this dict is a shallow copy)
     copied_function.__dict__.update(function.__dict__)
+
     return copied_function
 
 
-def copy_check(check):
+def copy_check(check: Callable) -> Callable:
     """
     Copy a check function so that internal attributes can be adjusted without changing the original function.
 
@@ -47,11 +46,12 @@ def copy_check(check):
     see https://github.com/NeurodataWithoutBorders/nwbinspector/pull/218 for explanation.
     """
     copied_check = _copy_function(function=check)
-    copied_check.__wrapped__ = _copy_function(function=check.__wrapped__)
+    copied_check.__wrapped__ = _copy_function(function=check.__wrapped__)  # type: ignore
+
     return copied_check
 
 
-def load_config(filepath_or_keyword: PathType) -> dict:
+def load_config(filepath_or_keyword: Union[str, Path]) -> dict:
     """
     Load a config dictionary either via keyword search of the internal configs, or an explicit filepath.
 
@@ -59,14 +59,15 @@ def load_config(filepath_or_keyword: PathType) -> dict:
         - 'dandi'
             For all DANDI archive related practices, including validation and upload.
     """
-    file = INTERNAL_CONFIGS.get(filepath_or_keyword, filepath_or_keyword)
+    file = INTERNAL_CONFIGS.get(str(filepath_or_keyword), filepath_or_keyword)
     with open(file=file, mode="r") as stream:
         config = yaml.safe_load(stream=stream)
+
     return config
 
 
 def configure_checks(
-    checks: list = available_checks,
+    checks: Optional[list] = None,
     config: Optional[dict] = None,
     ignore: Optional[list[str]] = None,
     select: Optional[list[str]] = None,
@@ -77,12 +78,12 @@ def configure_checks(
 
     Parameters
     ----------
+    checks : list of check functions, optional
+        If None, defaults to all registered checks.
     config : dict
         Dictionary valid against our JSON configuration schema.
         Can specify a mapping of importance levels and list of check functions whose importance you wish to change.
         Typically loaded via json.load from a valid .json file
-    checks : list of check functions
-        Defaults to all registered checks.
     ignore: list, optional
         Names of functions to skip.
     select: list, optional
@@ -100,6 +101,8 @@ def configure_checks(
 
         The default is the lowest level, BEST_PRACTICE_SUGGESTION.
     """
+    checks = checks or available_checks
+
     if ignore is not None and select is not None:
         raise ValueError("Options 'ignore' and 'select' cannot both be used.")
     if importance_threshold not in Importance:
@@ -107,9 +110,10 @@ def configure_checks(
             f"Indicated importance_threshold ({importance_threshold}) is not a valid importance level! Please choose "
             "from [CRITICAL_IMPORTANCE, BEST_PRACTICE_VIOLATION, BEST_PRACTICE_SUGGESTION]."
         )
+
+    checks_out: list = []
     if config is not None:
         validate_config(config=config)
-        checks_out = []
         ignore = ignore or []
         for check in checks:
             mapped_check = copy_check(check=check)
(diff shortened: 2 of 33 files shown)


pyproject.toml
```diff
@@ -92,7 +92,17 @@ extend-exclude = '''
...
+# TODO: eventually, expand to other 'D', and other 'F' linting
+select = [
+    "F401",
+    "F541",
+    "I",
+    "ANN001",
+    "ANN201",
+    "ANN202",
+    "ANN205",
+    "ANN206",
```
CodyCBakerPhD, line 104: These enforce enforce existence of annotations, which Mypy surprisingly does not do

src/nwbinspector/_configuration.py
```diff
@@ -3,41 +3,40 @@
...
+from typing import Optional, Union
 
 import jsonschema
 import yaml
+from typing_extensions import Callable
 
-from nwbinspector.utils import PathType
+from ._registration import Importance, available_checks
 
-from . import available_checks
```
CodyCBakerPhD, line 13: I removed localized `.` imports where possible to avoid circularlities

src/nwbinspector/_configuration.py
```diff
@@ -47,26 +46,28 @@ def copy_check(check):
...
+    file = INTERNAL_CONFIGS.get(str(filepath_or_keyword), filepath_or_keyword)
     with open(file=file, mode="r") as stream:
         config = yaml.safe_load(stream=stream)
+
     return config
 
 
 def configure_checks(
-    checks: list = available_checks,
+    checks: Optional[list] = None,
```
CodyCBakerPhD, line 70: This is something I also adjusted while I was here

Not only did it make Mypy feel better but it has been a long-standing issue with the API docs (since they then render the full listing of named check functions which floods the signature)

src/nwbinspector/_dandi_inspection.py
```diff
@@ -21,7 +21,7 @@ def inspect_dandiset(
     importance_threshold: Union[str, Importance] = Importance.BEST_PRACTICE_SUGGESTION,
     skip_validate: bool = False,
     show_progress_bar: bool = True,
-    client: Union["dandi.dandiapi.DandiAPIClient", None] = None,
+    client: Union["dandi.dandiapi.DandiAPIClient", None] = None,  # type: ignore
```
CodyCBakerPhD, line 24: Unfortunately does not play well with lazy imports, so those are flagged as skips

src/nwbinspector/_formatting.py
```diff
@@ -44,8 +45,8 @@ class FormatterOptions:
     """Class structure for defining all free attributes for the design of a report format."""
 
     def __init__(
-        self, indent_size: int = 2, indent: Optional[str] = None, section_headers: list[str] = ["=", "-", "~"]
-    ):
+        self, indent_size: int = 2, indent: Optional[str] = None, section_headers: tuple[str, ...] = ("=", "-", "~")
+    ) -> None:
```
CodyCBakerPhD, line 49: This was an old whoopsie on my end

Cardinal sin of Python to set an argument to mutable default

src/nwbinspector/_formatting.py
```diff
@@ -138,17 +139,17 @@ def _get_message_header(self, message: InspectorMessage):
...
-    def _get_message_increment(self, level_counter: list[int]):
+    def _get_message_increment(self, level_counter: list[int]) -> str:
         return (
             f"{'.'.join(np.array(level_counter, dtype=str))}.{self.message_counter}" f"{self.formatter_options.indent}"
         )
 
     def _add_subsection(
         self,
-        organized_messages: dict[str, Union[dict, list[InspectorMessage]]],
+        organized_messages: dict,
```
CodyCBakerPhD, line 149: While this could probably be typed with more detail as was attempted originally, it turns out to be pretty tricky to do right so maybe someone can tackle that later on its own PR

src/nwbinspector/_formatting.py
```diff
@@ -218,30 +219,37 @@ def format_messages(self) -> list[str]:
...
 
 
-def print_to_console(formatted_messages: list[str]):
+def print_to_console(formatted_messages: list[str]) -> None:
     """Print report file contents to console."""
     sys.stdout.write(os.linesep * 2)
     for line in formatted_messages:
         sys.stdout.write(line + "\n")
 
+    return None
```
CodyCBakerPhD, line 241: You may notice this around the place now

I'm not super crazy about it, but it is more explicit (and Python says explicit is better than implicit...)

Could theoretically suppress all return checks, but it did catch some pretty useful ones elsewhere

src/nwbinspector/_nwb_inspection.py
```diff
@@ -319,7 +323,9 @@ def _intercept_in_vitro_protein(nwbfile_object: pynwb.NWBFile, checks: Optional[
         "check_subject_proper_age_range",
     ]
     subject_related_dandi_requirements = [
-        check.importance == Importance.CRITICAL for check in checks if check.__name__ in subject_related_check_names
+        check.importance == Importance.CRITICAL  # type: ignore
+        for check in checks
+        if check.__name__ in subject_related_check_names
```
CodyCBakerPhD, line 328: Whenever you see skips like this, it really is indicative of how we probably never should have attached attributes to a function to begin with

We've discussed how that decision caused other issues (and all the copy function nonsense to theoretically solve it) so long term, if the design of the registry is refactored to be more appropriate, these ignores should go away

src/nwbinspector/tools/_read_nwbfile.py
```diff
@@ -66,8 +70,40 @@ def read_nwbfile(
...
+    # Note: do not be concerned about IO object closing due to garbage collection here
+    # (the IO object is attached as an attribute to the NWBFile object)
+    return nwbfile
+
+
+def read_nwbfile_and_io(
+    nwbfile_path: Union[str, Path],
+    method: Optional[Literal["local", "fsspec", "ros3"]] = None,
+    backend: Optional[Literal["hdf5", "zarr"]] = None,
+) -> tuple[NWBFile, HDMFIO]:
```
CodyCBakerPhD, line 106: This was a really bad design pattern on my part, I and should have known that  the time

Mypy particularly complained about the split typing, so I've properly separated it into two public methods that both call the same private helper

src/nwbinspector/tools/_nwb.py
```diff
@@ -1,17 +1,20 @@
...
-def get_nwbfile_path_from_internal_object(obj):
+def get_nwbfile_path_from_internal_object(neurodata_object: object) -> str:
     """Determine the file path on disk for a NWBFile given only an internal object of that file."""
-    if isinstance(obj, NWBFile):
-        return obj.container_source
-    return obj.get_ancestor("NWBFile").container_source
+    if isinstance(neurodata_object, NWBFile):
+        return neurodata_object.container_source
+
+    return neurodata_object.get_ancestor("NWBFile").container_source  # type: ignore
```
CodyCBakerPhD, line 20: This type of ignore is a potentially interesting one - it's because I type a generic NWB-type object as a Python `object` so any of the special class methods like `.get_ancestor` naturally are more specific than that - so it would be cool to have a common base neurodata object in PyNWB that all types inherit from (does that already exist? For both datasets and their containers / groups?)

CodyCBakerPhD: (recommend making more specific typing improvements down the road, and when ready remove the ignores here)

src/nwbinspector/inspector_tools/__init__.py
```diff
@@ -9,5 +9,12 @@
...
-from .._formatting import format_message, MessageFormatter, FormatterOptions, print_to_console, save_report
+from .._organization import organize_messages
+from .._formatting import (
+    format_messages,
+    MessageFormatter,
+    FormatterOptions,
+    print_to_console,
+    save_report,
+    _get_report_header,
+)
```
CodyCBakerPhD, line 20: Impressively, Mypy was even able to catch things like this (improper imports)

Quite curious why Ruff alone wasn't able to

src/nwbinspector/_formatting.py
```diff
@@ -75,14 +76,14 @@ class MessageFormatter:
...
+        messages: list[Optional[InspectorMessage]],
         levels: list[str],
         reverse: Optional[list[bool]] = None,
         detailed: bool = False,
         formatter_options: Optional[FormatterOptions] = None,
-    ):
+    ) -> None:
         self.nmessages = len(messages)
-        self.nfiles = len(set(message.file_path for message in messages))
+        self.nfiles = len(set(message.file_path for message in messages))  # type: ignore
```
CodyCBakerPhD, line 86: These types skips however are a weird false positive

We have to annotate various upstream signatures as `list[Optional[InspectorMessage]]` because a given check might have returned `None` - but the way we iterate the generator, those just don't show up in the final list of returned messages so these 'post-inspection' helpers really are typed for `list[InspectorMessage]`

CodyCBakerPhD:
@stephprince OK, sorry this one has quite a bit of a changelog (just a lot of the same things) - I tried excluding it piece by piece by the exclusion rules were not playing nice with pre-commit so went ahead and did it all

src/nwbinspector/utils/_utils.py
stephprince, line 1: If I understand correctly, `PathType` and `FilePathType` are still being used in a couple of places, but once I update those they should be able to be deprecated?

src/nwbinspector/testing/_testing.py
```diff
@@ -44,26 +47,28 @@ def check_streaming_tests_enabled() -> tuple[bool, Optional[str]]:
...
     """Generate a local copy of the NWB files required for the image series tests."""
     assert get_package_version(name="pynwb") == Version("2.1.0"), "Generating the testing files requires PyNWB v2.1.0!"
+    assert TESTING_FILES_FOLDER_PATH is not None, "The `TESTING_FILES_FOLDER_PATH` environment variable is not set!"
 
-    testing_config = load_testing_config()
-
-    local_path = Path(testing_config["LOCAL_PATH"])
-    local_path.mkdir(exist_ok=True, parents=True)
+    testing_folder = Path(TESTING_FILES_FOLDER_PATH)
+    testing_folder.mkdir(exist_ok=True)
```
stephprince, line 68:
```suggestion
    testing_folder.mkdir(exist_ok=True, parents=True)
```

might still be needed if nested directories?

src/nwbinspector/_formatting.py
stephprince: ah I see, good to know

src/nwbinspector/tools/_nwb.py
stephprince: There is the `AbstractContainer` class inherited from by both `Container` and `Data` types, but I don't know if there is a common base neurodata object specifically... 

Definitely worth discussing for typing improvements later on

stephprince (commented):
Looks good! Thanks for all the helpful comments and context information.

src/nwbinspector/utils/_utils.py
CodyCBakerPhD: Yep that sounds like a plan, I just did what I had to do to make Mypy happy, which was to sometimes fix that and other times not

src/nwbinspector/testing/_testing.py
CodyCBakerPhD: I caution against doing that very stringly, because if a target path is formatted incorrectly by mistake (say, like a WindowsPath) then calling `parents=True` will cause that folder to be created even if the root makes no sense for the OS (such as UNIX)

Ran into that one a couple times before.

If you want to ensure specific parents exist, I suggest going through them one at a time to make sure they exist (from top to bottom of course)

stephprince (approved):

PR #523, opened 2024-10-02
Change import of Callable for typing

### LINKED ISSUE #522: [Bug]: No module named 'typing_extensions'
### What happened?

On a clean install of the main branch, running `nwbinspector --config dandi path/to/file.nwb` results in:
```
Traceback (most recent call last):
  File "/Users/rly/mambaforge/envs/test/bin/nwbinspector", line 5, in <module>
    from nwbinspector._nwbinspector_cli import _nwbinspector_cli
  File "/Users/rly/Documents/NWB/nwbinspector/src/nwbinspector/__init__.py", line 3, in <module>
    from ._registration import available_checks, register_check
  File "/Users/rly/Documents/NWB/nwbinspector/src/nwbinspector/_registration.py", line 11, in <module>
    from typing_extensions import Callable
ModuleNotFoundError: No module named 'typing_extensions'
```

This reference was just added in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/520

I'm surprised the GitHub Actions tests passed. Perhaps `typing_extensions` is installed earlier in the GH Actions workflow. In this workflow, it looks like `pip install ".[dandi]"` is called. This installs `typing_extensions`, unlike `pip install .` I have not looked at all the workflows. If they all use `pip install ".[dandi]"`, I think we should have some tests without the `[dandi]` flag too.

### Operating System

Linux

### Python Version

3.12

None

### Package Versions

_No response_

### Code of Conduct

rly:
## Motivation

Fix #522

### DIFF
--- a/src/nwbinspector/_configuration.py
+++ b/src/nwbinspector/_configuration.py
@@ -1,13 +1,13 @@
 """Primary functions for inspecting NWBFiles."""
 
 import json
+from collections.abc import Callable
 from pathlib import Path
 from types import FunctionType
 from typing import Optional, Union
 
 import jsonschema
 import yaml
-from typing_extensions import Callable
 
 from ._registration import Importance, available_checks
 
--- a/src/nwbinspector/_registration.py
+++ b/src/nwbinspector/_registration.py
@@ -1,5 +1,6 @@
 """Primary decorator used on a check function to add it to the registry and automatically parse its output."""
 
+from collections.abc import Callable
 from functools import wraps
 from typing import List, Optional, Union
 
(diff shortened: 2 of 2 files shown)


rly:
Looks like something changed in the Windows runners or the conda action that is causing tests to fail. I'll quickly look into it.

rly:
I think the issue is due to conda being updated from 24.7.1 to 24.9.1 and the setup-conda action might be doing something weird with that. In any case, we can use setup-miniconda which we use elsewhere in NWB projects and is more widely used and maintained.

stephprince (approved):

PR #525, opened 2024-10-02
Adjust github actions to use setup-miniconda

rly:
Tests that use the s-weigand/setup-conda action are failing on Windows. Replace with conda-incubator/setup-miniconda. Also update some versions of GH Actions used.

### DIFF
--- a/.github/workflows/assess-file-changes.yml
+++ b/.github/workflows/assess-file-changes.yml
@@ -17,7 +17,7 @@ jobs:
 
     name: Test changed-files
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


rly:
Some workflows were using conda when it did not need to IMO. So I removed that call. Only the ros3 one needed conda.

stephprince (approved):
Looks good to me! I don't think a `CHANGELOG` is necessary for the workflow changes unless you want to add one.

And thanks for updating the workflows that don't need conda. I believe the workflows using conda that did not need to were the ones recently updated in the PRs that replaced `ros3` with `remfile`.

PR #531, opened 2024-10-24
Use cached namespaces during validation

### LINKED ISSUE #528: [Bug]: PyNWB validator not using cached schema
### What happened?

Using the dev branch, the PyNWB validator is not using cached extensions during validation. I think that is because we are using the `validate(io=io)` method instead of the `validate(paths=paths)` method. 

https://github.com/NeurodataWithoutBorders/nwbinspector/blame/ba7d845b255e8095c72978e91b2c014675ef55d6/src/nwbinspector/_nwb_inspection.py#L277

This appears to have been changed recently to support validation with Zarr.

See also https://github.com/NeurodataWithoutBorders/pynwb/issues/1777

I will look into solutions. We should add a test for this.

### Operating System

macOS

### Python Version

3.12

No

### Package Versions

_No response_

### Code of Conduct

stephprince:
## Motivation

Fix #528. 

Recent Zarr-related updates switched the inspector from using `pynwb.validate(paths=paths)` to `pynwb.validate(io=io)`,  resulting in the inspector not using cached namespaces during validation. We will temporarily revert this change, since pynwb validation of Zarr NWB files is not yet fully supported anyway due to pending issues that need to be addressed in hdmf/pynwb.

Since this is more of a bug on the pynwb side that using the `io` argument ignores the `use_cached_namespaces=True` setting in the validator (and will be updated in the next major pynwb release), I'm not sure if there are really any specific tests to add here.

While addressing this issue, I added a couple of other small updates to the Zarr support here:
1. Passing the path of an NWB file Zarr folder to the CLI (instead of the parent folder) would result in 0 files being inspected, due to some conditionals in `inspect_all` that detected whether the path provided was a directory of NWB files or an NWB file.
2. The `test_inspect_all_parallel` tests had been indented and were not running

### DIFF
--- a/src/nwbinspector/_nwb_inspection.py
+++ b/src/nwbinspector/_nwb_inspection.py
@@ -126,7 +126,9 @@ def inspect_all(
     if progress_bar_options is None:
         progress_bar_options = dict(position=0, leave=False)
 
-    if in_path.is_dir():
+    if in_path.is_dir() and (in_path.match("*.nwb*")) and (in_path / ".zgroup").exists():
+        nwbfiles = [in_path]  # if it is a zarr directory
+    elif in_path.is_dir():
         nwbfiles = list(in_path.rglob("*.nwb*"))
 
         # Remove any macOS sidecar files
--- a/src/nwbinspector/_nwb_inspection.py
+++ b/src/nwbinspector/_nwb_inspection.py
@@ -14,7 +14,7 @@ from tqdm import tqdm
 
 from ._configuration import configure_checks
 from ._registration import Importance, InspectorMessage, available_checks
-from .tools._read_nwbfile import read_nwbfile, read_nwbfile_and_io
+from .tools._read_nwbfile import read_nwbfile
 from .utils import (
     OptionalListOfStrings,
     PathType,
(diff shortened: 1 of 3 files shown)


src/nwbinspector/_nwb_inspection.py
```diff
@@ -126,7 +126,9 @@ def inspect_all(
     if progress_bar_options is None:
         progress_bar_options = dict(position=0, leave=False)
 
-    if in_path.is_dir():
+    if in_path.is_dir() and (in_path.match("*.nwb*")) and (in_path / ".zgroup").exists():
```
stephprince: I'm not sure if `(in_path / ".zgroup").exists()` is the appropriate check here for a Zarr NWB file or if it is even necessary

stephprince:
@rly you mentioned you had a related test file, can you let me know if this solves the issue or share the file?

src/nwbinspector/_nwb_inspection.py
rly: Can we use https://github.com/hdmf-dev/hdmf-zarr/blob/dev/src/hdmf_zarr/backend.py#L78 ?

stephprince:
@rly this should be ready for you to test with your example file and review

rly (approved):

PR #536, opened 2024-11-21
Remove numpy and h5py restrictions

rly:
## Motivation

numpy was pinned to <2.0 due to incompatibilities in pynwb and hdmf-zarr and maybe other tools. Those should be resolved now.

h5py was pinned to <3.12.0 due to an issue with windows that has been fixed and released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Upcoming
 
+### Improvements
+* Added support for Numpy 2 and h5py 3.12. [#536](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/536)
+
+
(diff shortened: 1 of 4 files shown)


rly:
The Windows failure is due to a change in Numpy 2.0:
> The default integer type on Windows is now int64 rather than int32, matching the behavior on other platforms

The tests need to be updated. I'll do this next week.

stephprince (commented):
should we pin to pynwb<3.0 now since I will need to make updates on the inspector side for the major pynwb release?

stephprince (approved):

rly:
Good idea @stephprince . I will do that here.

PR #537, opened 2024-11-21
update CHANGELOG for release 0.6.0

stephprince:
## Motivation

Prepare for release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Upcoming
 
-# v0.5.3
+# v0.6.0
 
 ### Deprecation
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Upcoming
 
+# v0.5.3
```
rly:
```suggestion
# v0.6.0
```

I would say this is more of a minor release since we added Zarr support. (If we decide 0.5.3 instead, then we'll need to update the version in `pyproject.toml`)

stephprince: that makes sense! Thanks, I missed that it had been updated in the `pyproject.toml` already.

rly (approved):

PR #549, opened 2025-01-09
rmv s3fs dependency

### LINKED ISSUE #548: [Feature]: rmv s3fs dependency

Currently, there is an  s3fs dependency. While not directly imported, it serves as the backend for fsspec when handling S3 URLs. Specifically, in src/nwbinspector/tools/_read_nwbfile.py, when reading files with "s3://" paths, the code uses fsspec.filesystem("s3", anon=True) which internally uses s3fs to handle the S3 file access.

The problem is that this causes a dependency chain that makes it difficult to use the latest packages. I think with the recent shift to using remfile, we should not need to maintain this apsect of the codebase. Do we want to be using fsspec at all? Are we even using the functions defined in _read_nwbfile.py at all? If not, let's refactor and simplify our dependencies for better environment management.

No.

### Code of Conduct

bendichter:
fix #548

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
     "pynwb>=2.8,<3",   # NWB Inspector should always be used with most recent minor versions of PyNWB
     "hdmf-zarr",
     "fsspec",
-    "s3fs",
     "requests",
     "aiohttp",
     "PyYAML",
@@ -63,7 +62,7 @@ nwbinspector = "nwbinspector._nwbinspector_cli:_nwbinspector_cli"
 
 [tool.black]
 line-length = 120
-target-version = ['py37']
+target-version = ['py39']
 include = '\.pyi?$'
 extend-exclude = '''
 /(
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -35,7 +35,6 @@ dependencies = [
     "pynwb>=2.8,<3",   # NWB Inspector should always be used with most recent minor versions of PyNWB
     "hdmf-zarr",
     "fsspec",
-    "s3fs",
```
stephprince, line 38: Since `read_nwbfile` with an `s3://` path would not technically work if we remove this as a dependency, I was wondering if we might want to make it an optional dependency instead? 

I added a more detailed discussion on the original issue.

stephprince (approved):

pyproject.toml
stephprince: Noted in the issue discussion that since this would be the only package in the optional dependency and the error message is clear, users can easily install this package if needed.

PR #552, opened 2025-01-13
Fix wrongly triggered compression check

### LINKED ISSUE #551: [Bug]: Compression check is wrongly triggered for large files
### What happened?

Currently, the warning is triggered if **there is** a compression indicator instead of being skipped as it should:

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/fc7491baecd266f873fd023e26cdbe4121dfbd79/src/nwbinspector/checks/_nwb_containers.py#L29-L41

It seems that this error was introduced here:

https://github.com/NeurodataWithoutBorders/nwbinspector/pull/513/files#r1913569891

On PR #513 

### Operating System

None

### Python Version

None

None

### Package Versions

<details> Copy and paste your output here </details>

### Code of Conduct

h-mayorquin:
## Motivation

Fix  #551

This also creates a definition for the size of the field/dataset and removes it from the if condition. I feel the other error got through partly because the if condition was a bit complicated and it is easy to miss logic the more difficult your assertions are.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ### Deprecation
 * Remove s3fs dependency, which was causing dependency management issues [#549](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/549)
 
+### Fixes
+* Fix wrongly triggered compression check [#552](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/552)
+
 ## Improvements
 * Added a section for describing the issues with negative timestamps in `TimeSeries` [#545](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/545)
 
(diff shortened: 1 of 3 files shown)


stephprince (commented):
thanks for catching @h-mayorquin!

It looks like we don't have any tests for the large dataset compression warning. Would you be able to add one?

I think replicating `test_check_small_dataset_compression_above_manual_threshold` using the `gb_lower_bound` option for `check_large_dataset_compression` would be fine to avoid creating a large dataset in the tests.

h-mayorquin:
Check the tests that I added.

Question: have you consider using pytest instead of unittest framework? I feel that using temporary files and directories is more friendly there:

https://docs.pytest.org/en/stable/how-to/tmp_path.html

This avoids intermittent errors on windows that result from naively trying to remove files on the test sometimes.

stephprince (approved):
Looks good to me!

I'm open to switching entirely to pytest. However, it looks like many tests currently use the `unittest.TestCase` class or the extended `TestCase` class from `hdmf.testing`. I think we would need to investigate a bit further to see how easy it is to migrate.

h-mayorquin:
Yep, you will probably need to change those. From my experience with other repos it is only the custom regex test for errors and warnings that is different but I also like the pytest version more for those. Plus, is easier to skip conditions.

PR #557, opened 2025-01-29
Add PyNWB 3.0 support

stephprince:
## Motivation
Address deprecations and validation changes in the upcoming [PyNWB 3.0 release](https://github.com/NeurodataWithoutBorders/pynwb/releases/tag/3.0.0rc1).

Note: I left some remaining TODO items that can be updated after the official release when 3.0 becomes the new minimum for NWBInspector.

To test:

```bash
cd nwbinspector
pip install -e . 
pip install pynwb==3.0.0rc1
pytest
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,25 +6,25 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black
-    rev: 24.10.0
+    rev: 25.1.0
     hooks:
     -   id: black
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.9.3
+  rev: v0.9.6
   hooks:
   - id: ruff
     args: [ --fix ]
 
 - repo: https://github.com/codespell-project/codespell
-  rev: v2.4.0
+  rev: v2.4.1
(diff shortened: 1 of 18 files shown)


rly:
Do you have any idea why the streaming tests are taking more than 1h 45m?

stephprince:
I'm not sure, I just noticed that. I can cancel the run for now and try to debug locally?

rly:
Sounds good.

Also, some time ago we had unofficially decided that NWB Zarr files should have the suffix ".nwb.zarr". I think we should keep that. https://github.com/NeurodataWithoutBorders/pynwb/issues/1976#event-15224332604 . NWB HDF5 files should still have suffix ".nwb" though. It can have the "hdf5" part before then to be explicit in these tests if we want.

stephprince:
> Also, some time ago we had unofficially decided that NWB Zarr files should have the suffix ".nwb.zarr". I think we should keep that. 

Ah right, thanks for catching. I was modifying the tests to address the new NWBHDF5IO warning to use the `.nwb` extension and forgot the recommendation was `.nwb.zarr` for Zarr files. I will fix those names.

stephprince:
@rly this is ready for you to re-review

rly (approved):

PR #564, opened 2025-02-24
Add python 3.13 support

### LINKED ISSUE #561: [Feature]: Add testing support for Python 3.13

Python 3.13 was released on Oct. 7, 2024. PyNWB will add 3.13 support in version 3.0 coming out in the next couple of weeks (3.0.0rc1 is out). 

It would be good to add official support for 3.13 soon after the PyNWB 3.0 release. 

No.

### Code of Conduct

- [x] Have you ensured this feature was not already [requested](https://github.com/neurodatawithoutborders/nwbinspector/issues)?

stephprince:
## Motivation

Fix #561. Add python 3.13 to testing workflows.

### DIFF
--- a/src/nwbinspector/checks/_image_series.py
+++ b/src/nwbinspector/checks/_image_series.py
@@ -1,7 +1,6 @@
 """Check functions specific to ImageSeries."""
 
-import ntpath
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Iterable, Optional
 
 from pynwb.image import ImageSeries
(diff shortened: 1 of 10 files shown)


src/nwbinspector/checks/_image_series.py
```diff
@@ -45,7 +44,7 @@ def check_image_series_external_file_relative(image_series: ImageSeries) -> Opti
         return None
     for file_path in image_series.external_file:
         file_path = file_path.decode() if isinstance(file_path, bytes) else file_path
-        if ntpath.isabs(file_path):  # ntpath required for cross-platform detection
+        if PureWindowsPath(file_path).is_absolute() or PurePosixPath(file_path).is_absolute():
```
stephprince, line 47: In Python 3.13, `ntpath.isabs()` no longer considers paths starting with exactly one slash to be absolute ([see note here](https://docs.python.org/3/whatsnew/3.13.html#os-path)).

From what I can tell, the intended behavior of `check_image_series_external_file_relative` is to detect whether `file_path` is an absolute path on either Windows/Posix systems and regardless of the current os the file is being checked on. Using the `PurePaths` lets us check the path on any system (using posixpath and ntpath).

rly (approved):

PR #572, opened 2025-03-06
Add check subject weight

### LINKED ISSUE #370: [Add Check]: Subject weight should have form "[numeric] [string]"

In the schema, subject weight is a string. In general, we want to encourage SI units for all fields, with exceptions. For weight, this is kg. In [PyNWB](https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.Subject.weight), we allow float/string to be entered, and if a float is entered, " kg" is appended to the end. 

If a unit is not part of the weight string, it is ambiguous what unit the weight is in. Therefore, I suggest that the subject weight string be checked to have the form "[numeric] [string]". I would argue this is a critical error.

This may be addressed in the schema, but for now, I think it would be good to address this in the inspector.

_No response_

No.

### Code of Conduct

anoushkajain:
Fix #370

### DIFF
--- a/docs/best_practices/nwbfile_metadata.rst
+++ b/docs/best_practices/nwbfile_metadata.rst
@@ -238,6 +238,16 @@ Check function: :py:meth:`~nwbinspector.checks._nwbfile_metadata.check_subject_a
 
 
 
+.. _best_practice_subject_weight:
+
+Subject Weight
+~~~~~~~~~~~~~
+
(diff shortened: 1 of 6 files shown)


bendichter:
This is looking good! However, this PR includes two checks, which is not ideal.

I think "70kg" should pass. Here, it fails

Also, could you add a sentence in the best practices document describing this?

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/docs/best_practices/nwbfile_metadata.rst#subject

PR #532, opened 2024-11-10
Raise best practice violation if ":" in container name

rly:
## Motivation

Related to https://github.com/hdmf-dev/hdmf-zarr/issues/219 and https://github.com/hdmf-dev/hdmf/pull/1202

PR #580, opened 2025-03-31
remove hdmf pin from workflows

stephprince:
## Motivation

There was previously a regression in hdmf 3.14.4 that required temporarily pinning to hdmf==3.14.3 in the dandi test workflows. This issue was addressed in the subsequent hdmf release.  

Removing this pin should fix one of the daily workflows that requires a newer version of hdmf for the best practice checks for colons in container names (https://github.com/NeurodataWithoutBorders/nwbinspector/pull/532)

### DIFF
--- a/.github/workflows/dandi-dev-live-services.yml
+++ b/.github/workflows/dandi-dev-live-services.yml
@@ -28,5 +28,4 @@ jobs:
           git clone https://github.com/dandi/dandi-cli
           cd dandi-cli
           pip install -e .[test]
-          pip install hdmf==3.14.3  # temporary: https://github.com/dandi/dandi-cli/issues/1494
           pytest -vv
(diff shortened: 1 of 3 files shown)


stephprince:
test-dandi-latest is now passing and I believe the other failures are unrelated

rly (approved):

rly:
Tests are failing due to zarr & numcodecs incompatibility described here https://github.com/hdmf-dev/hdmf-zarr/issues/269
which will be resolved in the hdmf/hdmf-zarr bugfix releases tomorrow

stephprince:
> Tests are failing due to zarr & numcodecs incompatibility described here https://github.com/hdmf-dev/hdmf-zarr/issues/269
> which will be resolved in the hdmf/hdmf-zarr bugfix releases tomorrow

These tests seem to be resolved with the latest release, but now there is a separate asciitree build error when installing dandi. I'll look into this.

PR #581, opened 2025-04-12
Add check for dimensionality of SpikeEventSeries in electrical series validation

### LINKED ISSUE #550: [Bug]: Inaccurate critical error for 2D SpikeEventSeries
### What happened?

According to the [nwb-schema](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/d5a931d2cc0ae059d917705b7583136d756a2c8b/core/nwb.ecephys.yaml#L68-L84), a SpikeEventSeries can have dimensions [num events, num samples] when it contains data from a single electrode. The other option is to have dimensions [num events, num channels, num samples] if it has data from multiple electrodes. 

However if an NWB file contains a 2D SpikeEventSeries, the inspector will return a critical error that the second dimension does not match the length of electrodes:

```
0  CRITICAL
===========

0.0  createESandSES.nwb: check_electrical_series_dims - 'SpikeEventSeries' object at location '/acquisition/spikedata1'
       Message: The second dimension of data does not match the length of electrodes. Your data may be transposed.

```

I think the check should be updated to warn about potentially transposed data only if a SpikeEventSeries has 3 dimensions.

### Operating System

macOS

### Python Version

None

None

### Package Versions

<details> Copy and paste your output here </details>

### Code of Conduct

bendichter:
fix #550 

> The bug described in issue #550 has been fixed. The check for transposed data in ElectricalSeries was updated so that for SpikeEventSeries, the warning about the second dimension not matching the number of electrodes is only triggered for 3D data, not for valid 2D data. A new test was added to ensure that 2D SpikeEventSeries do not trigger a warning, while 3D SpikeEventSeries with mismatched electrodes do. All unit tests pass, confirming the fix is correct and covered by tests.

Cline-assisted (optimus-alpha)

### DIFF
--- a/src/nwbinspector/checks/_ecephys.py
+++ b/src/nwbinspector/checks/_ecephys.py
@@ -3,7 +3,7 @@
 from typing import Optional
 
 import numpy as np
-from pynwb.ecephys import ElectricalSeries
+from pynwb.ecephys import ElectricalSeries, SpikeEventSeries
 from pynwb.misc import Units
 
 from .._registration import Importance, InspectorMessage, register_check
@@ -39,6 +39,26 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries) -> Optiona
     electrodes = electrical_series.electrodes
 
     data_shape = get_data_shape(data, strict_no_data_load=True)
+
+    # For SpikeEventSeries, only perform the check for 3D data
+    if isinstance(electrical_series, SpikeEventSeries):
+        if data_shape and len(data_shape) == 3 and data_shape[1] != len(electrodes.data):
+            if data_shape[0] == len(electrodes.data):
+                return InspectorMessage(
+                    message=(
+                        "The second dimension of data does not match the length of electrodes, "
+                        "but instead the first does. Data is oriented incorrectly and should be transposed."
+                    )
+                )
+            return InspectorMessage(
+                message=(
+                    "The second dimension of data does not match the length of electrodes. Your data may be transposed."
+                )
+            )
+        # Do not warn for 2D SpikeEventSeries
+        return None
+
+    # For other ElectricalSeries, keep the original logic
     if data_shape and len(data_shape) == 2 and data_shape[1] != len(electrodes.data):
         if data_shape[0] == len(electrodes.data):
             return InspectorMessage(
@@ -49,7 +69,7 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries) -> Optiona
             )
         return InspectorMessage(
             message=(
-                "The second dimension of data does not match the length of electrodes. Your " "data may be transposed."
+                "The second dimension of data does not match the length of electrodes. Your data may be transposed."
             )
         )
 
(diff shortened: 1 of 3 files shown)


tests/unit_tests/test_ecephys.py
```diff
@@ -156,6 +156,40 @@ def test_trigger_check_electrical_series_reference_electrodes_table(self):
...
+    # 3D data: [num_events, num_channels, num_samples] with mismatched num_channels (should trigger warning)
+    ses_3d = SpikeEventSeries(
+        name="spike_events_3d",
+        data=np.zeros((10, 4, 5)),  # 4 != 3 electrodes
+        electrodes=electrodes,
+        timestamps=[0.1 * i for i in range(10)],
+    )
+    result = check_electrical_series_dims(ses_3d)
+    assert isinstance(result, InspectorMessage)
+    assert "The second dimension of data does not match the length of electrodes" in result.message
```
stephprince:
```suggestion
    assert result == InspectorMessage(
        message=("The second dimension of data does not match the length of electrodes. Your data may be transposed."),
        importance=Importance.CRITICAL,
        check_function_name="check_electrical_series_dims",
        object_type="SpikeEventSeries",
        object_name="spike_events_3d",
        location="/",
        )
```

This could be updated to check for all InspectorMessage attributes to be more similar to other checks.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -39,6 +39,26 @@ def check_electrical_series_dims(electrical_series: ElectricalSeries) -> Optiona
...
+    # For SpikeEventSeries, only perform the check for 3D data
+    if isinstance(electrical_series, SpikeEventSeries):
+        if data_shape and len(data_shape) == 3 and data_shape[1] != len(electrodes.data):
+            if data_shape[0] == len(electrodes.data):
+                return InspectorMessage(
+                    message=(
+                        "The second dimension of data does not match the length of electrodes, "
+                        "but instead the first does. Data is oriented incorrectly and should be transposed."
+                    )
+                )
```
stephprince, line 52: This is technically missing test coverage right now. It is unlikely at least for files created with PyNWB to have the data be truly transposed since there are checks that the number of timestamps match the number of events when SpikeEventSeries are initialized. I suppose it could happen for files written with other software however.

We could add something like this to the tests:
```python
    # 3D data: [num_events, num_channels, num_samples] with flipped num_channels / num_events (should trigger warning)
    # this is an unlikely case in the latest PyNWB since there is a check that the number of timestamps and events are the same
    ses_3d_flipped = SpikeEventSeries(
        name="spike_events_flipped",
        data=np.zeros((3, 10, 5)),  # 10 != 3 electrodes
        electrodes=electrodes,
        timestamps=[0.1 * i for i in range(3)],  
    )
    result = check_electrical_series_dims(ses_3d_flipped)
    assert result == InspectorMessage(
        message=("The second dimension of data does not match the length of electrodes, but instead the first does. "
                 "Data is oriented incorrectly and should be transposed."),
        importance=Importance.CRITICAL,
        check_function_name="check_electrical_series_dims",
        object_type="SpikeEventSeries",
        object_name="spike_events_flipped",
        location="/",
        )

stephprince (commented):
looks good, thanks! Can you also add a changelog entry?

stephprince (approved):

stephprince:
I believe the dandi dev test failures are unrelated and this is good to merge

PR #584, opened 2025-04-15
Add check for empty datasets in NWB containers

### LINKED ISSUE #324: [Add Check]: Datasets are not empty

From https://github.com/NeurodataWithoutBorders/pynwb/issues/1002, check that datasets are not empty.

Maybe need to consider edge case of ImageSeries with external file format, memory doesn't reclal if the `Dataset` there is `None`, empty, or if the `data` attribute is not set at all in that case.

_No response_

Yes.

### Code of Conduct

anoushkajain:
Fix https://github.com/NeurodataWithoutBorders/nwbinspector/issues/324

### DIFF
--- a/src/nwbinspector/checks/_nwb_containers.py
+++ b/src/nwbinspector/checks/_nwb_containers.py
@@ -4,6 +4,7 @@ import os
 from typing import Iterable, Optional
 
 import h5py
+import numpy as np
 import zarr
 from pynwb import NWBContainer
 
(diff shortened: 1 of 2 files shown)


src/nwbinspector/checks/_nwb_containers.py
```diff
@@ -42,6 +42,39 @@ def check_large_dataset_compression(
...
+    nwb_container: NWBContainer
+        The NWB container to check for empty datasets.
+
+    Returns
+    -------
+    Optional[Iterable[InspectorMessage]]
+        Inspector messages for each empty dataset found, or None if no empty datasets are found.
+    """
+    for field_name, field in getattr(nwb_container, "fields", dict()).items():
+        if not isinstance(field, (h5py.Dataset, zarr.Array)):
```
stephprince, line 64: I think we may also want to check for other empty array-like objects here, including numpy arrays and DataIO objects. Could use something like: 
```suggestion
        is_array_data = hasattr(value, "shape") and hasattr(value, "dtype")
        if not is_array_data:
```

src/nwbinspector/checks/_nwb_containers.py
```diff
@@ -42,6 +42,39 @@ def check_large_dataset_compression(
...
+            continue
+
+        # Check if the dataset has zero elements
+        if field.size == 0:
```
stephprince, line 68: I am not sure if it will be necessary but for the DataIO objects you may need to use `get_data_shape` from the `nwbinspector.utils` module.

src/nwbinspector/checks/_nwb_containers.py
```diff
@@ -42,6 +42,39 @@ def check_large_dataset_compression(
     return None
 
 
+@register_check(importance=Importance.CRITICAL, neurodata_type=NWBContainer)
```
stephprince, line 45: I am not sure yet whether this should be a critical check or a best practice violation. There was a discussion here about whether empty datasets should be allowed: https://github.com/NeurodataWithoutBorders/pynwb/issues/2065.

stephprince (changes requested):
Thanks for the PR @anoushkajain!

I think this looks good but there are a couple more pieces to add before merging.

1. We may need to expand the check to other array-like dataset objects that could be present in NWB files that haven’t been written to disk yet. I've added a suggestion to the code related to this. 
2. Can you add tests including passing and failing cases with the different types of datasets.
3. Can you add a new best practice section to the documentation recommending that objects with empty datasets should not be included in the file.

Also in the future, feel free to @ me or request a review whenever you are ready for a PR to be looked at!

PR #585, opened 2025-04-15
Remove unused setuptools reference on pyproject.toml.

h-mayorquin:
While debugging a build issue with this package, I discovered that the library uses `setuptools` directives even though the build system is `hatchling`.

I'm interpreting the original intent and translating the `setuptools` configuration directly to `hatchling`. @rly 

I'm also updating the `pynwb` version as mentioned in the comment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,8 +54,8 @@ dandi = [
     "remfile",
 ]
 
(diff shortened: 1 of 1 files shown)


stephprince (approved):
thanks @h-mayorquin!

PR #607, opened 2025-08-04
Add check for NWB schema version to ensure official releases

### LINKED ISSUE #432: [Add Check]: Check for unofficial versions of schema
AFAIK ATM we do not impose any limits on versions of nwb schema in the files really... I agree indeed it could be at least some kind of a WARNING level annotation by nwb inspector... and I would not mind ERROR level for "dandi" profile to avoid cases of unloadable files the future.

_Originally posted by @yarikoptic in https://github.com/NeurodataWithoutBorders/pynwb/issues/1837#issuecomment-1922021678_

bendichter:
Add `check_nwb_schema_version_official_release` to validate that NWB files use official release versions rather than development or pre-release versions. This check is configured as CRITICAL for DANDI validation to ensure production data compatibility and reproducibility.

## Motivation

fix #432


src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -20,6 +20,54 @@
...
+    Best Practice: Use only official, released versions of the NWB schema for
+    production data to ensure compatibility and reproducibility.
+    """
+    try:
+        import pynwb
+
+        # Get the schema version from the namespace catalog
+        manager = pynwb.get_manager()
+        core_namespace = manager.namespace_catalog.get_namespace("core")
+        schema_version = core_namespace.version
```
stephprince, line 41: This check uses the loaded namespace from PyNWB to check the schema version validity and not the nwbfile argument provided. I think it would be more direct to check the `nwb_version` attribute of the nwbfile provided. Thoughts?

tests/unit_tests/test_nwbfile_metadata.py
```diff
@@ -619,3 +620,75 @@ def test_check_subject_id_with_slashes():
...
+    assert check_nwb_schema_version_official_release(minimal_nwbfile) is None
+
+
+def test_check_nwb_schema_version_official_release_fail():
+    """Test that unofficial schema versions fail the check."""
+    # We need to mock the schema version to test different scenarios
+    import unittest.mock
+
+    # Test development version
+    with unittest.mock.patch("pynwb.get_manager") as mock_manager:
```
stephprince, line 637: See the comment on the check, but I think these tests could also be updated/simplified using mock nwbfiles with the `nwb_version` attribute manually changed instead of patching the namespace manager

src/nwbinspector/checks/_nwbfile_metadata.py
bendichter: Agreed, it reads the file's own nwb_version now, through the IO it was read with. In-memory files are skipped.

tests/unit_tests/test_nwbfile_metadata.py
bendichter: Done, the tests attach a mock IO with the version instead.

Issue #567, opened 2025-02-27
[Documentation] Should we add "use the nwb extension" as a good practice in the docs?


h-mayorquin:
I am thinking on the context of this:

https://github.com/NeurodataWithoutBorders/pynwb/pull/1978

stephprince:
I think that would be good to add as a best practice in the inspector docs. 

Also to add the recommendation to use the `.nwb.zarr` extension, which I don't think is documented anywhere yet (but a similar [warning in hdmf-zarr](https://github.com/hdmf-dev/hdmf-zarr/issues/230) is currently in progress).

h-mayorquin:
Ok, so the policy is that:
* hdf5 - add only .nwb 
* zarr - add .nwb.zarr

Why the difference? is because the later is a directory/folder? Is the discussion of the choice documented somewhere?

bendichter:
I like the idea of recommending attaching the backend as a second suffix. Usually when you have multiple extensions the earlier extensions are more about meaning and the later are more about form (e.g. x.schema.json), so nwb.zarr would be better than zarr.nwb.

We could technically also have .nwb.h5, but I don't think it's necessary since h5 is sort of the default. Also, just for historical reasons, since we started with just .nwb. There is precedent for having HDF5 files and using a different suffix (e.g. .mat).

Yes, technically .nwb.h5 would be more explicit and would make it clear to the user that they can open it with any h5 reader, though might make things confusing for users that are not familiar with HDF5. Also, it would require a lot of ecosystem support, particularly with the inspector and the dandi cli. It's not impossible to support but I don't think if would provide enough value to justify the effort that this point.

stephprince:
This was addressed in #625

PR #625, opened 2025-10-08
Add best practices documentation and check for NWB file extension

stephprince:
## Motivation

Address #567. See also discussion summary in https://github.com/hdmf-dev/hdmf-zarr/pull/265#issuecomment-2698893106.

Adds a best practice suggestion that an nwbfile uses either a `.nwb`, `.nwb.zarr` or `.nwb.h5` file extension.


src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -352,3 +355,44 @@ def check_subject_id_no_slashes(subject: Subject) -> Optional[InspectorMessage]:
...
+    NWB files should use appropriate extensions based on their backend:
+    - .nwb (minimum recommendation), .nwb.h5 (HDF5), or .nwb.zarr (Zarr)
+
+    Best Practice: :ref:`best_practice_file_extension`
+    """
+    file_path = get_nwbfile_path_from_internal_object(nwbfile)
+
+    # Only perform the check if we can determine the file path
+    if file_path is not None:
+        file_extension = "".join(Path(file_path).suffixes)  # Get all suffixes for multi-part extensions
```
rly:
```suggestion
        file_extension = "".join(Path(file_path).suffixes)  # Concatenate all suffixes for multi-part extensions
```

Adding this because I was confused as to what this does.

src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -352,3 +355,44 @@ def check_subject_id_no_slashes(subject: Subject) -> Optional[InspectorMessage]:
...
+        all_valid_extensions = [".nwb", ".nwb.h5", ".nwb.zarr"]
+
+        if isinstance(nwbfile.get_read_io(), NWBHDF5IO):
+            valid_extensions = [".nwb", ".nwb.h5"]
+            backend = "HDF5"
+        elif isinstance(nwbfile.get_read_io(), NWBZarrIO):
+            valid_extensions = [".nwb", ".nwb.zarr"]
+            backend = "Zarr"
```
rly:
```suggestion
        read_io = nwbfile.get_read_io()
        if isinstance(read_io, NWBHDF5IO):
            valid_extensions = [".nwb", ".nwb.h5"]
            backend = "HDF5"
        elif isinstance(read_io, NWBZarrIO):
            valid_extensions = [".nwb", ".nwb.zarr"]
            backend = "Zarr"
```

src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -352,3 +355,44 @@ def check_subject_id_no_slashes(subject: Subject) -> Optional[InspectorMessage]:
...
+
+        # check the extension contains .nwb or .nwb.h5/.nwb.zarr
+        msg = (
+            f"The file extension '{file_extension}' does not follow the recommended naming convention. "
+            f"{backend} NWB files should use one of the following file name extensions: {', '.join(valid_extensions)}."
+        )
+        if not any(pattern in file_extension for pattern in valid_extensions):
```
rly: This check and the check below checks for `pattern in file_extension` which allows for "file.xx.nwb.xx" to be valid. I think we should check that the file ends with one of the accepted extensions instead. What do you think?

stephprince: Right, I was thinking if there were cases where someone would want to include additional extensions (e.g. I saw `example.nwb.lindi.json` used in the lindi docs). But since this is a best practice suggestion for final file names (and lindi is not currently included in this check) maybe this approach is overly permissive?

rly: Yeah, I think since these are just best practices, it is OK to be more strict. I would say as LINDI (and other backends) gets more integrated into the core tools, then we update the best practices to support the LINDI extension

stephprince: Sounds good - updated the code to check that the file ends with the recommended extensions! I also added a condition in case we can't get the read_io for some reason to just check against all potential valid extensions

src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -352,3 +355,48 @@ def check_subject_id_no_slashes(subject: Subject) -> Optional[InspectorMessage]:
...
+            valid_extensions = [".nwb", ".nwb.h5"]
+            backend = "HDF5"
+        elif isinstance(read_io, NWBZarrIO):
+            valid_extensions = [".nwb", ".nwb.zarr"]
+            backend = "Zarr"
+        else:
+            valid_extensions = all_valid_extensions
+            backend = ""
+
+        # check the extension contains .nwb or .nwb.h5/.nwb.zarr
```
rly:
```suggestion
        # check the extension ends with .nwb or .nwb.h5/.nwb.zarr
```

stephprince: thanks for catching!

rly (approved):

rly:
Looks good!

Issue #626, opened 2025-10-09
[Add Check]: Duration of TimeSeries


bendichter:
### What would you like to see added to the NWBInspector?

This PR for DANDI CLI: https://github.com/dandi/dandi-cli/pull/1714 contains code for checking the duration of files. We should include similar code here and flag if any of the timeseries etc. are above a certain duration (e.g. 1 year), as this is likely an error.

### Pseudocode or other logic of the check

```shell

```

### Do you have any interest in helping implement the check function?

Yes.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [x] Have you ensured this check was not already [requested](https://github.com/neurodatawithoutborders/nwbinspector/issues)?

PR #627, opened 2025-10-09
Add time series duration check and corresponding tests

bendichter:
related to #626


src/nwbinspector/checks/_time_series.py
```diff
@@ -202,3 +202,78 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+    # Calculate duration from timestamps if available
+    if time_series.timestamps is not None:
+        timestamps_shape = get_data_shape(time_series.timestamps)
+        if timestamps_shape is not None and timestamps_shape[0] > 1:
+            first_timestamp = time_series.timestamps[0]
+            last_timestamp = time_series.timestamps[-1]
+            duration = float(last_timestamp - first_timestamp)
+
+    # Calculate duration from starting_time and rate if timestamps not available
+    elif time_series.starting_time is not None and time_series.rate is not None and time_series.rate > 0:
```
stephprince:
```suggestion
    elif time_series.rate is not None and time_series.rate > 0:
```

I don't think the starting_time check is needed here since it is not used?

src/nwbinspector/checks/_time_series.py
```diff
@@ -202,3 +202,78 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+        # Convert threshold to years for the message (assuming 1 year = 365.25 days)
+        threshold_years = duration_threshold / 31557600.0
+        duration_years = duration / 31557600.0
+        return InspectorMessage(
+            message=(
+                f"TimeSeries '{time_series.name}' has a duration of {duration:.2f} seconds ({duration_years:.2f} years), "
+                f"which exceeds the threshold of {duration_threshold:.2f} seconds ({threshold_years:.2f} years). "
+                "Please verify that this is correct."
+            )
+        )
```
stephprince:
```suggestion
        # Convert duration to years for the message
        duration_years = duration / 31557600.0
        return InspectorMessage(
            message=(
                f"TimeSeries '{time_series.name}' has an unusually long duration of {duration:.2f} seconds ({duration_years:.2f} years), "
                f"which may indicate an error in the timestamps or rate data. "
                "Please verify that this is correct."
            )
        )
```

I think this message could be a little clearer as to the motivation why we are flagging this.

src/nwbinspector/checks/_time_series.py
```diff
@@ -202,3 +202,78 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+
+    return None
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=TimeSeries)
```
stephprince: Should this be critical or a best practice violation?

src/nwbinspector/checks/_time_series.py
```diff
@@ -202,3 +202,78 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+
+    if time_series.rate is not None and 0 < time_series.rate < low_rate_threshold:
+        period = 1.0 / time_series.rate
+        return InspectorMessage(
+            message=(
+                f"TimeSeries '{time_series.name}' has a sampling rate of {time_series.rate}Hz (period of {period:.2f} seconds), "
+                f"which is below the expected threshold of {low_rate_threshold}Hz. "
+                "This may indicate that the period was specified instead of the rate. "
+                f"If the intended period is {time_series.rate} seconds, the rate should be {1.0 / time_series.rate}Hz."
+            )
```
stephprince:
```suggestion
                f"TimeSeries '{time_series.name}' has a sampling rate of {time_series.rate}Hz (period of {period:.2f} seconds). "
                "This low sampling rate may indicate that the period was specified instead of the rate. "
                f"If the intended period is {time_series.rate} seconds, the rate should be {1.0 / time_series.rate}Hz."
            )
```

Similar to above, since the threshold is somewhat arbitrary I think it is fine to just state the motivation here

tests/unit_tests/test_time_series.py
```diff
@@ -413,3 +415,208 @@ def test_check_rate_is_positive_fail():
...
+def test_check_rate_not_below_threshold_pass_at_threshold():
+    """Test that a rate exactly at the threshold passes."""
+    time_series = pynwb.TimeSeries(
+        name="test_time_series",
+        unit="test_units",
+        data=np.zeros(shape=100),
+        starting_time=0.0,
+        rate=0.01,  # Exactly at the default threshold
+    )
+    assert check_rate_not_below_threshold(time_series) is None
```
stephprince:
```suggestion
```

I think this test is unnecessary since the threshold is arbitrary

tests/unit_tests/test_time_series.py
```diff
@@ -413,3 +415,208 @@ def test_check_rate_is_positive_fail():
...
+        starting_time=0.0,
+        rate=period_value,
+    )
+    result = check_rate_not_below_threshold(time_series)
+    # Should pass with default threshold since 2.0 > 0.01
+    assert result is None
+
+    # But should fail with a custom threshold
+    result = check_rate_not_below_threshold(time_series, low_rate_threshold=5.0)
+    assert result is not None
```
stephprince:
```suggestion
```

I think this test covers the same behavior as the test above and below?

tests/unit_tests/test_time_series.py
```diff
@@ -413,3 +415,208 @@ def test_check_rate_is_positive_fail():
...
+        name="long_time_series",
+        unit="test_units",
+        data=np.zeros(shape=100),
+        timestamps=np.linspace(0, one_year + 1000, 100),  # Exceeds 1 year
+    )
+    result = check_time_series_duration(time_series)
+    assert result is not None
+    assert "long_time_series" in result.message
+    assert "exceeds the threshold" in result.message
+    assert result.importance == Importance.BEST_PRACTICE_SUGGESTION
```
stephprince: Can you update these to be checked against the full inspector message? Similarly in the other tests.

e.g. `check_time_series_duration(time_series) == InspectorMessage(...`

tests/unit_tests/test_time_series.py
```diff
@@ -413,3 +415,208 @@ def test_check_rate_is_positive_fail():
...
+    one_year = 31557600.0
+    rate = 1.0  # 1 Hz
+    num_samples = int(one_year + 1000)  # More than 1 year worth of samples
+    time_series = pynwb.TimeSeries(
+        name="long_time_series",
+        unit="test_units",
+        data=np.zeros(shape=num_samples),
+        starting_time=0.0,
+        rate=rate,
+    )
```
stephprince, line 498: We might want to avoid creating a large array here just for tests, could lower the rate instead to minimize the number of samples

tests/unit_tests/test_time_series.py
```diff
@@ -413,3 +415,208 @@ def test_check_rate_is_positive_fail():
...
+def test_check_time_series_duration_pass_two_samples():
+    """Test that TimeSeries with only two samples passes (duration cannot be calculated reliably)."""
+    time_series = pynwb.TimeSeries(
+        name="test_time_series",
+        unit="test_units",
+        data=np.zeros(shape=2),
+        timestamps=[0, 1],
+    )
+    # Should pass - even though there's a duration, with only 2 samples we skip
+    assert check_time_series_duration(time_series) is None
```
stephprince:
```suggestion
```

This is not skipping duration calculations for two samples (and I don't think it should). It passes because the duration is short. Think it's fine to remove this test instead.

src/nwbinspector/checks/_time_series.py
```diff
@@ -202,3 +202,78 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=TimeSeries)
+def check_time_series_duration(
+    time_series: TimeSeries, duration_threshold: float = 31557600.0
+) -> Optional[InspectorMessage]:
+    """
+    Check if the TimeSeries duration is longer than the specified threshold.
+
+    The default threshold is 1 year (31,557,600 seconds = 365.25 days).
```
stephprince, line 223: Is there a reason for the year long threshold? I imagine anything longer than a month or two could be worth flagging since it is a best practice suggestion.

stephprince (commented):
Added my review, could you also add a CHANGELOG entry?

src/nwbinspector/checks/_time_series.py
bendichter: Hard to know, really. A few labs are going as long as a month. I'm mostly trying to catch when people use milliseconds or microseconds instead of seconds.

tests/unit_tests/test_time_series.py
h-mayorquin: I think this was was addressed in [6cf9922](https://github.com/NeurodataWithoutBorders/nwbinspector/commit/6cf9922d7390fcd197ba009e005d009511547a06)

tests/unit_tests/test_time_series.py
```diff
@@ -428,3 +430,193 @@ def test_check_rate_is_positive_fail():
...
+    time_series = pynwb.TimeSeries(
+        name="test_time_series",
+        unit="test_units",
+        data=np.zeros(shape=100),
+        timestamps=np.linspace(0, 100, 100),
+    )
+    assert check_rate_not_below_threshold(time_series) is None
+
+
+def test_check_rate_not_below_threshold_pass_zero_rate():
```
h-mayorquin: The zero-rate case is already caught by [`check_rate_is_not_zero`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/1ff328fcbf96c7b7fc8a22b51d981cd84d694abf/src/nwbinspector/checks/_time_series.py#L187-L200) at `CRITICAL` importance so maybe it is better to remove this and leave the work to that check?

src/nwbinspector/checks/_time_series.py
```diff
@@ -211,3 +211,76 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
         )
 
     return None
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=TimeSeries)
```
h-mayorquin: This was [discussed before](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/627#discussion_r2417413877) and I don't know if you and @stephprince talked about this in person but I think both should be `BEST_PRACTICE_VIOLATION`.

These two new checks are heuristic error detectors that flag values which likely indicate a mistake in the data (timestamps in the wrong units, sampling period used instead of rate). This is different from the other `BEST_PRACTICE_SUGGESTION` checks in the library, which are more like style or metadata completeness checks. For example, [`check_experimenter_exists`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/73f99488fa6d35c0dbd7ea2e8e256bd95f39f33f/src/nwbinspector/checks/_nwbfile_metadata.py#L71) reminds you to add an experimenter name, [`check_keywords`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/73f99488fa6d35c0dbd7ea2e8e256bd95f39f33f/src/nwbinspector/checks/_nwbfile_metadata.py#L131) reminds you to add keywords, and [`check_file_extension`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/73f99488fa6d35c0dbd7ea2e8e256bd95f39f33f/src/nwbinspector/checks/_nwbfile_metadata.py#L387) suggests using a standard file extension. Those are "nice to have" suggestions where the data itself is still correct. In contrast, a TimeSeries spanning years or a sampling rate of 0.001 Hz almost certainly indicates an error that the user should know about. A user who knows their data is correct can always ignore the warning.

An important practical difference is that DANDI sets its [`importance_threshold` to `BEST_PRACTICE_VIOLATION`](https://github.com/dandi/dandi-cli/blob/ccdb6593597c50cd9d8d7a905643b20d81caf3e6/dandi/files/bases.py#L549), so `BEST_PRACTICE_SUGGESTION` checks are skipped entirely during DANDI validation. Neither level blocks upload (both map to [`Severity.WARNING` or `Severity.HINT`](https://github.com/dandi/dandi-cli/blob/ccdb6593597c50cd9d8d7a905643b20d81caf3e6/dandi/files/bases.py#L51-L55)), but making both `VIOLATION` ensures DANDI users see the warning and if there is an error correct it.

src/nwbinspector/checks/_time_series.py
```diff
@@ -211,3 +211,76 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
...
+    if time_series.rate is not None and 0 < time_series.rate < low_rate_threshold:
+        period = 1.0 / time_series.rate
+        return InspectorMessage(
+            message=(
+                f"TimeSeries '{time_series.name}' has a sampling rate of {time_series.rate}Hz (period of {period:.2f} seconds). "
+                "This low sampling rate may indicate that the period was specified instead of the rate. "
+                f"If the intended period is {time_series.rate} seconds, the rate should be {1.0 / time_series.rate}Hz."
+            )
+        )
+
```
h-mayorquin, line 289: If someone passed a rate of 0.001 Hz this message reads:

> TimeSeries 'test_time_series' has a sampling rate of 0.001Hz (period of 1000.00 seconds). This low sampling rate may indicate that the period was specified instead of the rate. If the intended period is 0.001 seconds, the rate should be 1000.0Hz.

I find it confusing that the word "period" appears twice with different meanings: the first time as the implied period of the current rate (1000.00 seconds), and the second time as the hypothesized actual sampling period (0.001 seconds). Consider rewording to make it easier to parse:

> TimeSeries 'test_time_series' has a sampling rate of 0.001 Hz (one sample every 1000.00 seconds). This low value may indicate the sampling period was provided instead of the rate. If the sampling period of the data is indeed 0.001 seconds, the rate should be set to 1000.0 Hz instead.

h-mayorquin (changes requested):
This looks good, some questions and suggestions. I also fixed the read the docs problems here:
https://github.com/NeurodataWithoutBorders/nwbinspector/pull/662

src/nwbinspector/checks/_time_series.py
```diff
@@ -211,3 +211,80 @@ def check_rate_is_positive(time_series: TimeSeries) -> Optional[InspectorMessage
         )
 
     return None
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=TimeSeries)
```
bendichter:
```suggestion
@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
```

tests/unit_tests/test_time_series.py
```diff
@@ -428,3 +430,193 @@ def test_check_rate_is_positive_fail():
...
+    duration = one_year + 1000
+    duration_years = duration / 31557600.0
+    expected_message = (
+        f"TimeSeries 'long_time_series' has an unusually long duration of {duration:.2f} seconds ({duration_years:.2f} years), "
+        f"which may indicate an error in the timestamps or rate data. "
+        "Please verify that this is correct."
+    )
+    assert check_time_series_duration(time_series) == InspectorMessage(
+        message=expected_message,
+        importance=Importance.BEST_PRACTICE_SUGGESTION,
```
bendichter:
```suggestion
        importance=Importance.BEST_PRACTICE_VIOLATION,
```

tests/unit_tests/test_time_series.py
```diff
@@ -428,3 +430,193 @@ def test_check_rate_is_positive_fail():
...
+    duration = (num_samples - 1) / rate
+    duration_years = duration / 31557600.0
+    expected_message = (
+        f"TimeSeries 'long_time_series' has an unusually long duration of {duration:.2f} seconds ({duration_years:.2f} years), "
+        f"which may indicate an error in the timestamps or rate data. "
+        "Please verify that this is correct."
+    )
+    assert check_time_series_duration(time_series) == InspectorMessage(
+        message=expected_message,
+        importance=Importance.BEST_PRACTICE_SUGGESTION,
```
bendichter:
```suggestion
        importance=Importance.BEST_PRACTICE_VIOLATION,
```

src/nwbinspector/checks/_time_series.py
bendichter: I'm happy with this rewording

bendichter: @copilot

tests/unit_tests/test_time_series.py
```diff
@@ -428,3 +430,193 @@ def test_check_rate_is_positive_fail():
...
+    """Test that zero rate passes (handled by different check)."""
+    time_series = pynwb.TimeSeries(
+        name="test_time_series",
+        unit="test_units",
+        data=np.zeros(shape=1),
+        starting_time=0.0,
+        rate=0.0,
+    )
+    # Zero rate should pass this check (it's handled by check_rate_is_not_zero)
+    assert check_rate_not_below_threshold(time_series) is None
```
bendichter:
```suggestion
```

bendichter:
@copilot fix the tests

h-mayorquin (approved):
LGTM

PR #628, opened 2025-10-09
Add check for duration of time columns in DynamicTable

bendichter:
#627


bendichter:
@stephprince this PR has been sitting for over a month. Can you please review it?

stephprince:
Yes I can take a look this afternoon! Sorry about that - thanks for the ping

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +292,90 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+    # Check for timestamp column (possibly with duration)
+    if "timestamp" in table.colnames and len(table["timestamp"]) > 0:
+        timestamp_data = table["timestamp"]
+        start_times.append(float(timestamp_data[0]))
+
+        if "duration" in table.colnames and len(table["duration"]) > 0:
+            duration_data = table["duration"]
+            end_times.append(float(timestamp_data[-1] + duration_data[-1]))
+        else:
+            end_times.append(float(timestamp_data[-1]))
```
stephprince, line 341: Do we recommend storing data this way somewhere? This seems relatively arbitrary to look for (vs. the `start_time`/`end_time` and `spike_times` columns we know exist in DynamicTables from the schema).
```suggestion
```

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +292,90 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+                start = float(np.min(np.r_[st_data[0], st_data[idxs[:-1]]]))
+            else:
+                start = float(st_data[0])
+
+            end = float(np.max(st_data[idxs - 1]))
+            start_times.append(start)
+            end_times.append(end)
+
+    # Calculate duration if we found any time data
+    if start_times and end_times:
```
stephprince, line 364: I think with the current warning it may be unclear to the user what is causing the duration errors. 

Thoughts on splitting into two different checks, one for spike times and one for start/end times to provide additional information as to what columns are triggering the InspectorMessage? The message would then be more specific e.g. something like, 

```python
(f"The 'start_time' and 'end_time' columns in DynamicTable '{table.name}' have a total duration of {duration:.2f} seconds "
f"({duration_years:.2f} years), which may indicate an error in the units used."
f"Please verify that this is correct." )
```

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +292,90 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+            st_data = table["spike_times"].target
+
+            if len(idxs) > 1:
+                start = float(np.min(np.r_[st_data[0], st_data[idxs[:-1]]]))
+            else:
+                start = float(st_data[0])
+
+            end = float(np.max(st_data[idxs - 1]))
+            start_times.append(start)
+            end_times.append(end)
```
stephprince, line 361: I think this could be simplified to get the duration from the max/min of all spike time data unless I am missing something. If the spike times are relative to the session start time then we are really looking for any case of spike times with a duration more than a year apart.


```suggestion
    if "spike_times" in table.colnames and len(table["spike_times"]) > 0:
        all_spike_times = np.asarray(units_table["spike_times"].target.data[:]) 
        duration = np.max(all_spike_times) - np.min(all_spike_times)
```

src/nwbinspector/checks/_tables.py
stephprince: There are some other inspector checks that search for column names ending with `_time`. I think those could be checked for duration

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +292,90 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+        return None  # Empty table
+
+    start_times = []
+    end_times = []
+
+    # Check for start_time and stop_time columns (e.g., trials)
+    if "start_time" in table.colnames and len(table["start_time"]) > 0:
+        start_times.append(float(table["start_time"][0]))
+        if "stop_time" in table.colnames and len(table["stop_time"]) > 0:
+            end_times.append(float(table["stop_time"][-1]))
```
stephprince, line 330: We could also check max/min of all start and stop times if we don't want to assume that these columns are also following ascending order best practices

tests/unit_tests/test_tables.py
```diff
@@ -498,3 +499,82 @@ def test_table_time_columns_are_not_negative_multidimensional_pass():
...
+    one_year = 31557600.0
+    table = DynamicTable(name="events", description="test events")
+    table.add_column(name="timestamp", description="event timestamps")
+    table.add_column(name="duration", description="event durations")
+    table.add_row(timestamp=0.0, duration=10.0)
+    table.add_row(timestamp=one_year, duration=1000.0)
+
+    result = check_table_time_columns_duration(table)
+    assert result is not None
+    assert "exceeds the threshold" in result.message
```
stephprince, line 570: See comment in `_tables.py` about these column names and whether we should keep these tests

tests/unit_tests/test_tables.py
```diff
@@ -498,3 +499,82 @@ def test_table_time_columns_are_not_negative_multidimensional_pass():
     test_table.add_row(test_time=[0.0, 1.0, 2.0, 3.0])
 
     assert check_table_time_columns_are_not_negative(test_table) is None
+
+
+def test_check_table_time_columns_duration_pass_short():
```
stephprince, line 504: Can you add tests for the Units table / spike_times case?

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # v0.6.6 (Upcoming)
 
 ### New Checks
+* Added checks for the duration of DynamicTables by checking start_time, stop_time, timestamp, duration, and spike_times columns. [#628](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/628)
 * Added `check_file_extension` for NWB file extension best practice recommendations (`.nwb`, `.nwb.h5`, or `.nwb.zarr`) [#625](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/625)
 
 ### Improvements
-* Added documentation to API and CLI docs on how to use the dandi config option. [#624](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/624)
+* Added documentation to API and CLI docs on how to use the dandi config option. [#628](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/628)
```
```suggestion
* Added documentation to API and CLI docs on how to use the dandi config option. [#624](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/624)
```

stephprince (commented):
@bendichter added my review - can you also add a best practice documentation section?

src/nwbinspector/checks/_tables.py
bendichter: This is for [ndx-events](https://github.com/rly/ndx-events). I should have mentioned that somewhere, huh?

src/nwbinspector/checks/_tables.py
bendichter: Yeah, I think you are right, maybe this would be better as 2 (or even 3) checks

src/nwbinspector/checks/_tables.py
bendichter: The idea here is to get the first and last spike times for each unit. This allows us to avoid reading the entire dataset (though in practice it may be faster to just read the entire dataset. I'll admit I have not checked)

stephprince:
@bendichter I think we can close this in favor of #635 and #636?

PR #635, opened 2025-11-21
feat: Add `check_time_intervals_duration` to verify TimeIntervals table duration and include unit tests.

bendichter:
a more modular PR for #628

### DIFF
--- a/src/nwbinspector/checks/__init__.py
+++ b/src/nwbinspector/checks/__init__.py
@@ -140,6 +140,7 @@ __all__ = [
     "check_column_binary_capability",
     "check_dynamic_table_region_data_validity",
     "check_time_interval_time_columns",
+    "check_time_intervals_duration",
     "check_time_intervals_stop_after_start",
     "check_table_values_for_dict",
     "check_table_time_columns_are_not_negative",
--- a/src/nwbinspector/checks/__init__.py
+++ b/src/nwbinspector/checks/__init__.py
@@ -75,8 +75,8 @@ from ._tables import (
     check_table_time_columns_are_not_negative,
     check_table_values_for_dict,
     check_time_interval_time_columns,
-    check_time_intervals_stop_after_start,
     check_time_intervals_duration,
+    check_time_intervals_stop_after_start,
 )
 from ._time_series import (
     check_data_orientation,
--- a/src/nwbinspector/checks/__init__.py
+++ b/src/nwbinspector/checks/__init__.py
@@ -76,6 +76,7 @@ from ._tables import (
     check_table_values_for_dict,
     check_time_interval_time_columns,
     check_time_intervals_stop_after_start,
+    check_time_intervals_duration,
 )
 from ._time_series import (
     check_data_orientation,
--- a/src/nwbinspector/checks/_tables.py
+++ b/src/nwbinspector/checks/_tables.py
@@ -1,7 +1,6 @@
 """Check functions that can apply to any descendant of DynamicTable."""
 
 from numbers import Real
-from tkinter import N
 from typing import Iterable, Optional
 
 import numpy as np
--- a/src/nwbinspector/checks/_tables.py
+++ b/src/nwbinspector/checks/_tables.py
@@ -18,6 +18,7 @@ from ..utils import (
 )
 
 NELEMS = 200
+MAX_DURATION = 3600 * 24 * 365.25 # default: 1 year
 
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=DynamicTableRegion)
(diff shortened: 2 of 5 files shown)


src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +293,61 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
                 )
 
     return None
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=TimeIntervals)
```
h-mayorquin, line 298: Should this be a critical error? We are basically saying that they won't ever be such cases. Seems too strong to me.

We should have a way of handling "things that are usually errors but there is a small probability that the user knows what they are doing and they can move forward"

The solution could a non-trivial barrier to enable this like an environment variable that could skip this error (as a CLI argument would be hard to propagate to DANDI) or something of the like.

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +293,61 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+    end_times = []
+
+    # Check for start_time and stop_time columns
+    if "start_time" in time_intervals.colnames and len(time_intervals["start_time"]) > 0:
+        start_times.append(float(time_intervals["start_time"][0]))
+        if "stop_time" in time_intervals.colnames and len(time_intervals["stop_time"]) > 0:
+            end_times.append(float(time_intervals["stop_time"][-1]))
+
+    # Check for other time columns
+    for column_name in time_intervals.colnames:
```
h-mayorquin, line 327: I think that all the times in the time columns should be smaller than the  `max(time_intervals[`stop_time`].data), right? Maybe that could be a check on its own.

src/nwbinspector/checks/_tables.py
```diff
@@ -292,3 +293,61 @@ def check_table_time_columns_are_not_negative(table: DynamicTable) -> Optional[I
...
+        Maximum expected duration in seconds. Default is 1 year (365.25 days).
+    """
+    if len(time_intervals.id) == 0:
+        return None
+
+    start_times = []
+    end_times = []
+
+    # Check for start_time and stop_time columns
+    if "start_time" in time_intervals.colnames and len(time_intervals["start_time"]) > 0:
```
h-mayorquin: Here and with the other time columns we are assuming that the time columns are already well-ordered. Is there a way of running checks in order? Maybe this check does not make sense if the other fails and the output will confuse rather than clarify if the ascending order check fails.

h-mayorquin (commented):
Some questions.

src/nwbinspector/checks/_tables.py
bendichter: Yes, this check requires that the times are in order. If they are not, this check may fail to raise a message. I suppose we could just read all the data. That's probably fine in most cases -- I doubt we'll come across many datasets where these datasets are large. Do you think it's better to just read all of the time arrays?

h-mayorquin: It is definitely more safe. 

If we want to be more efficient maybe we can just combine these two checks:

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/d8382c925a34cf245e75a889f2144a707515138f/src/nwbinspector/checks/_tables.py#L52-L80

To avoid reading them twice.

I think either way is fine.

stephprince: I agree it would be better to not assume the columns are following ascending order best practices. Most of the tables checks limit the search to NELEMS (default 200). Maybe for this check we could read the first 200 start times and last 200 stop times?

I believe if we use the `cache_data_selection` function as used in the `is_ascending_series` then we will avoid reading the data twice for the minimum start time check. I would lean towards keeping the checks separate since they cover slightly different best practices.

src/nwbinspector/checks/_tables.py
stephprince: I imagine there could be some other trial related variables like "reward/punishment_end_time"  that experimenters do not want to consider as part of a trial start/stop_time.... but I am curious if others think that is a good check to add that the start/stop time contain all other _time columns.

tests/unit_tests/test_tables.py
```diff
@@ -498,3 +499,72 @@ def test_table_time_columns_are_not_negative_multidimensional_pass():
...
+    table.add_row(start_time=150.0, stop_time=200.0, custom_time=one_year + 1000)
+
+    result = check_time_intervals_duration(table)
+    assert result is not None
+    assert "trials" in result.message
+    assert "exceeds the threshold" in result.message
+
+
+def test_check_time_intervals_duration_pass_without_start_stop():
+    """Test that tables with only other time columns work correctly."""
```
stephprince:
```suggestion
def test_check_time_intervals_duration_pass_without_additional_time_columns():
    """Test that check passes additional time columns that are within the threshold."""
```

src/nwbinspector/checks/_tables.py
stephprince:
```suggestion
```

`start_time` is a required column for a TimeIntervals table and we check that it is not empty above. I think technically both of these conditionals could be removed? But correct me if I'm missing a case.

Similarly with `if "stop_time"` below

src/nwbinspector/checks/__init__.py
stephprince, line 142:
```suggestion
    "check_time_interval_time_columns",
    "check_time_intervals_duration",
```

src/nwbinspector/checks/_tables.py
stephprince: > Should this be a critical error? We are basically saying that they won't ever be such cases. Seems too strong to me.

I think this is a good question that will apply to all the related duration PRs as well... what combination of duration threshold and check importance do we want to apply? As we discussed in other PRs the original check purpose is to catch when people are using milliseconds or microseconds instead of seconds. I think a 1-2 year threshold for a single NWB session is reasonable but maybe these cases are starting to exist.

> We should have a way of handling "things that are usually errors but there is a small probability that the user knows what they are doing and they can move forward"

We have one case (`check_data_orientation`) where the inspector default is CRITICAL but the dandi inspector config deelevates it to a BEST_PRACTICE_VIOLATION. We could do something similar here. However I'm not sure how useful this is if most of the documentation recommends users to run the inspector with the dandi config.

bendichter: > The solution could a non-trivial barrier to enable this like an environment variable that could skip this error

yes, we have the ability to skip checks with config files and with passed variables. We could use environment variables as well.

The definition of CRITICAL, according to the docs is:

> High likelihood of incorrect data that can’t be caught by PyNWB validation

[source](https://nwbinspector.readthedocs.io/en/dev/contributing_checks.html#choose-the-right-importance-level)

I feel pretty confident that there won't be any NWB files that are a single session over 1 year in the near future. As soon as we see a real case I am fine with demoting this to a best practice violation.

src/nwbinspector/checks/_tables.py
bendichter: stop_time can contain NaNs, in which case the other columns may be relevant

src/nwbinspector/checks/_tables.py
h-mayorquin: > yes, we have the ability to skip checks with config files and with passed variables.

How do you think is best to convey this information?

I think we could append the following message to the error:

"If your recording is really longer than {max_expected_duration} check the following link for how to bypass this" 

or something along these lines.

I think it would be really frustrating if a user with such an experiment comes along and they are just unable to upload their data without further recourse or information.

h-mayorquin (approved):
This looks good to me. I think we could improve the way that we convey how users can bypass heuristic based checks but I don't think this should block this PR. I will open an issue.

bendichter:
@h-mayorquin I agree, I think this issue makes it clear that there are checks that should block but should also be ignorable. The CLI tool supports   

--ignore TEXT                   Comma-separated names of checks to skip.

but that is not very well-documented. Do you think adding a note to our docs would suffice? Maybe we could try passing that arg to the dandi cli as well?

h-mayorquin:
> @h-mayorquin I agree, I think this issue makes it clear that there are checks that should block but should also be ignorable. The CLI tool supports
> 
> --ignore TEXT Comma-separated names of checks to skip.
> 
> but that is not very well-documented. Do you think adding a note to our docs would suffice? Maybe we could try passing that arg to the dandi cli as well?

Yes, this makes sense.

PR #636, opened 2025-11-21
Add check for units table duration

bendichter:
Modular PR for #628


src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,71 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+    spike_times = units["spike_times"].target
+    if len(idxs) > 1:
+        start = np.min(np.r_[spike_times[0], spike_times[idxs[:-1]]])
+    else:
+        start = spike_times[0]
+
+    end = np.max(spike_times[idxs - 1])
+
+    if len(spike_times) == 0:
+        return None
```
bendichter:
```suggestion
    spike_times = units["spike_times"].target
    if len(spike_times) == 0:
        return None
    if len(idxs) > 1:
        start = np.min(np.r_[spike_times[0], spike_times[idxs[:-1]]])
    else:
        start = spike_times[0]

    end = np.max(spike_times[idxs - 1])
```

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,70 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
                 )
             )
     return None
+
+
+@register_check(importance=Importance.CRITICAL, neurodata_type=Units)
```
stephprince, line 147: See discussion in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635#discussion_r2560318425. Whatever we decide there should be consistent across PRs.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,70 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+    if len(idxs) > 1:
+        start = np.min(np.r_[spike_times[0], spike_times[idxs[:-1]]])
+    else:
+        start = spike_times[0]
+
+    end = np.max(spike_times[idxs - 1])
+
+    start = float(np.min(spike_times))
+    end = float(np.max(spike_times))
+    duration = end - start
```
stephprince, line 207: If we want to be more consistent with discussion in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635#discussion_r2560374912 and not assume that spike times are ascending, then we could also read in all the data. There is one example of this in `check_negative_spike_times` so maybe it is not too much of a performance concern? I am fine either way.

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/672c1690aaaba7d7c6b4ca1115ce2e49299f3ba3/src/nwbinspector/checks/_ecephys.py#L16-L20

stephprince (commented):
Looks good to me - main comments are just related to consistency with #635

src/nwbinspector/checks/_ecephys.py
bendichter: I'd rather not read all the spikes if we don't have to. I think this could be changed in the other check as well

src/nwbinspector/checks/_ecephys.py
bendichter: I think this should be critical

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,87 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+    last_spike_idxs = idxs[idxs != 0] - 1
+
+    # Combine into single array of indices to read, then read all at once
+    all_indices = np.concatenate([first_spike_idxs, last_spike_idxs])
+    all_indices = np.unique(all_indices)  # Remove duplicates for efficiency
+
+    # Read only the needed spike times in one operation
+    spike_times_data = units_table["spike_times"].target.data
+
+    # print(f"{idxs=}", flush=True)
```
h-mayorquin: This print statements should be removed.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,87 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+    -------
+    Optional[InspectorMessage]
+        An InspectorMessage if the duration exceeds the threshold, None otherwise.
+    """
+    if "spike_times" not in units_table:
+        return None
+
+    # Read the index array (cumulative indices marking end of each unit's spikes)
+    # This is small - just one integer per unit
+    idxs = np.asarray(units_table["spike_times"].data[:])
```
h-mayorquin, line 178: I think this code is a bit harder to read than it should because of hdmf data model for ragged arrays.  I was confused for some minutes on why the variable is called indices until I remember that this is a ragged array. 

I think that converting the unit to data frames would make this logic simpler.

tests/unit_tests/test_ecephys.py
```diff
@@ -312,3 +313,100 @@ def test_ascending_spike_times_empty(self):
...
+    """Test that units table with a single unit and long duration is detected."""
+    units = Units(name="units")
+    units.add_unit(spike_times=[0.0, 50000000.0])  # ~1.58 years
+
+    result = check_units_table_duration(units)
+    assert result is not None
+    assert "Units table has a duration of 50000000.00 seconds" in result.message
+
+
+def test_check_units_table_duration_first_unit_no_spikes():
```
h-mayorquin, line 395: this and the next tests seems a bit repetitive.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,87 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+
+    if duration > duration_threshold:
+        duration_years = duration / DURATION_THRESHOLD
+        threshold_years = duration_threshold / DURATION_THRESHOLD
+        return InspectorMessage(
+            message=(
+                f"Units table has a duration of {duration:.2f} seconds "
+                f"({duration_years:.2f} years), which exceeds the threshold of "
+                f"{duration_threshold:.2f} seconds ({threshold_years:.2f} years). "
+                "This may indicate that spike_times are in the wrong units or there is a data quality issue."
```
h-mayorquin:
```suggestion
                "This may indicate that spike_times are not in seconds or a data quality issue."
```

This is more specific on why are we throwing the error and also suggest the fix.

h-mayorquin (commented):
Seem good to me, some minor comments.

src/nwbinspector/checks/_ecephys.py
bendichter: I'm trying to avoid reading all the spike times. I am just reading the first and last spike for each unit. Converting this to a dataframe would read all of the spike times.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -140,3 +142,87 @@ def check_ascending_spike_times(units_table: Units, nelems: Optional[int] = NELE
...
+
+    # Read only the needed spike times in one operation
+    spike_times_data = units_table["spike_times"].target.data
+
+    # print(f"{idxs=}", flush=True)
+    # print(f"{first_spike_idxs=}", flush=True)
+    # print(f"{last_spike_idxs=}", flush=True)
+    # print(f"{all_indices=}", flush=True)
+    # print(f"{spike_times_data=}", flush=True)
+    # print(f"{all_indices=}", flush=True)
```
bendichter:
```suggestion
```

src/nwbinspector/checks/_ecephys.py
h-mayorquin: Good point. It is unfortunate that we don't have data frame like lazy reading.

bendichter: That was the idea behind https://github.com/catalystneuro/lazy_ops

oruebel: > Good point. It is unfortunate that we don't have data frame like lazy reading.

You should be able to read individual rows as a DataFrame though

bendichter: you'd still end up reading all the spike times, which is what I am trying to avoid

tests/unit_tests/test_ecephys.py
bendichter: no these are actually two different conditions that need to be tested independently. The first unit it special because it does not have a starter index, so when the first unit has no spike times, the index reads [0,...] and for any other unit it's [...,x,x,...].

src/nwbinspector/checks/_ecephys.py
bendichter: fixed.

tests/unit_tests/test_ecephys.py
h-mayorquin: I see, thanks.

tests/unit_tests/test_ecephys.py
```diff
@@ -312,3 +313,100 @@ def test_ascending_spike_times_empty(self):
...
+    """Test that units table where first unit has no spikes works correctly."""
```
h-mayorquin:
```suggestion
    """Test handling of empty spike_times for the first unit.
    
    The first unit is a special case in the index structure: it has no preceding
    index value, so when the first unit has no spikes, the index array starts as
    [0, ...] (the first value equals the second). This test ensures the duration
    calculation correctly handles this edge case without index errors.
    """```

File: tests/unit_tests/test_ecephys.py [hunk 10]
```diff
@@ -312,3 +313,100 @@ def test_ascending_spike_times_empty(self):
...
+    units = Units(name="units")
+    units.add_unit(spike_times=[])  # First unit has no spikes
+    units.add_unit(spike_times=[0.0, 1.0, 2.0])  # Second unit has spikes
+
+    # Should pass - duration is only 2 seconds
+    assert check_units_table_duration(units) is None
+
+
+def test_check_units_table_duration_second_unit_no_spikes():
+    """Test that units table where second unit has no spikes works correctly."""
```
h-mayorquin:
```suggestion
    """Test handling of empty spike_times for a non-first unit.
    
    Non-first units with no spikes produce a different index pattern than the
    first unit: the index array contains [..., x, x, ...] where consecutive
    values are equal. This test ensures the duration calculation correctly
    handles this distinct edge case.
    """
```

h-mayorquin (approved):

tests/unit_tests/test_ecephys.py
```diff
@@ -312,3 +313,112 @@ def test_ascending_spike_times_empty(self):
...
+
+
+def test_check_units_table_duration_first_unit_no_spikes():
+    """Test handling of empty spike_times for the first unit.
+
+    The first unit is a special case in the index structure: it has no preceding
+    index value, so when the first unit has no spikes, the index array starts as
+    [0, ...] (the first value equals the second). This test ensures the duration
+    calculation correctly handles this edge case without index errors.
+    """```
```
bendichter:
```suggestion
    """
```

PR #638, opened 2025-11-25
Improve `TimeSeries` wrong orientation message

h-mayorquin:
## Motivation

Currently the error does not give the shape which is, I think, crucial context for the users.

The RP improves in this direction and also provides a suggestion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Improvements
 * Added documentation to API and CLI docs on how to use the dandi config option. [#624](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/624)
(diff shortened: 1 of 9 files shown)


bendichter:
@h-mayorquin you've changed "Time should be in the first dimension, and is usually the longest dimension." to "Time should be the longest dimension, which is usually the first.". The first sentence is correct. The second implies that time is always longest (which is not true) and that it can sometimes not be first (which is also not true)

h-mayorquin:
> @h-mayorquin you've changed "Time should be in the first dimension, and is usually the longest dimension." to "Time should be the longest dimension, which is usually the first.". The first sentence is correct. The second implies that time is always longest (which is not true) and that it can sometimes not be first (which is also not true)

My bad, corrected it now to use the previous message but printing the shape and a suggestion.

stephprince (approved):
Looks good to me - thanks @h-mayorquin!

PR #640, opened 2025-11-27
Fix link targets for badges in README for nwbinspector project

ehennestad:
Fix link targets for README badges

## Motivation

Links for python version and codecov badges point to a different project. This PR fixes those link targets


stephprince (approved):
thanks @ehennestad !

PR #647, opened 2025-12-24
feat(checks): add check_subject_weight to validate weight format

### LINKED ISSUE #370: [Add Check]: Subject weight should have form "[numeric] [string]"

bendichter:
Add new check that ensures subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '25kg'). Without a unit, the weight is ambiguous.

- Add weight_form_regex pattern for validation
- Register check with CRITICAL importance
- Add unit tests for valid/invalid weight formats

Closes #370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Added `check_units_table_duration` to detect if the duration of spike times in a Units table exceeds a threshold (default: 1 year), which may indicate spike_times are in the wrong units or there is a data quality issue.
 * Added `check_time_intervals_duration`, which makes sure that `TimeInterval` objects do not have a duration greater than 1 year.
 [#635](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635)
-* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '25kg'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
+* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '10 g'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
 
 ### Improvements
 * Added documentation to API and CLI docs on how to use the dandi config option. [#624](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/624)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Added `check_file_extension` for NWB file extension best practice recommendations (`.nwb`, `.nwb.h5`, or `.nwb.zarr`) [#625](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/625)
 * Added `check_time_intervals_duration`, which makes sure that `TimeInterval` objects do not have a duration greater than 1 year.
 [#635](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635)
-* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '25kg'). [#370](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/370)
+* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '25kg'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
 
 ### Improvements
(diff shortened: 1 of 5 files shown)


src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -19,6 +19,7 @@
     r"?M)?(\d+(?:\.\d+)?S)?)?$"
 )
 species_form_regex = r"([A-Z][a-z]* [a-z]+)|(http://purl.obolibrary.org/obo/NCBITaxon_\d+)"
+weight_form_regex = r"^\d+(\.\d+)?\s*[A-Za-z]+$"
```
h-mayorquin: This regex matches units that are not from the international metric system. I suggest that we make it more strict in two ways:
1) Only accept units from the SI 
2) Enforce a space between the number and the units.

This regex should do it
```suggestion
weight_form_regex = r"(?i)^\d+(\.\d+)? (kg|g|mg|ug|μg|ng|pg)$"
```

docs/best_practices/nwbfile_metadata.rst
```diff
@@ -249,6 +249,19 @@ The ``strain`` of a :ref:`nwb-schema:sec-Subject` should be set to further indic
 
 
 
+.. _best_practice_subject_weight:
+
+Subject Weight
```
h-mayorquin, line 254: I think we can cite this document:

image

https://www.bipm.org/en/publications/si-brochure

tests/unit_tests/test_nwbfile_metadata.py
```diff
@@ -662,3 +663,65 @@ def test_check_file_extension_fail():
             result = check_file_extension(read_nwbfile)
             msg = f"The file extension '{ext}' does not follow the recommended naming convention."
             assert msg in result.message
+
+
+def test_check_subject_weight_pass():
+    """Test that valid weight formats pass the check."""
+    valid_weights = ["2.3 kg", "25 kg", "2.5kg", "100g", "1.5 lbs", "0.5 kg"]
```
h-mayorquin:
```suggestion
    valid_weights = ["2.3 kg", "25 kg", "0.5 kg", "100 g"]
```
I think we should not accept spaces and keep it to SI units.

h-mayorquin (commented):
This looks good. I think that if we are gonna make the requirement more strict here we should follow the SI standard in full.

src/nwbinspector/checks/_nwbfile_metadata.py
bendichter: ok, but then the error message needs to state this.

tests/unit_tests/test_nwbfile_metadata.py
h-mayorquin, line 1: We should add a test that fails if there is no space between the umber and the units and this should be good to go.

h-mayorquin (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Added `check_units_table_duration` to detect if the duration of spike times in a Units table exceeds a threshold (default: 1 year), which may indicate spike_times are in the wrong units or there is a data quality issue.
 * Added `check_time_intervals_duration`, which makes sure that `TimeInterval` objects do not have a duration greater than 1 year.
 [#635](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635)
+* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '25kg'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
```
bendichter:
```suggestion
* Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '10 g'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
```

tests/unit_tests/test_nwbfile_metadata.py
bendichter: done

h-mayorquin (approved):
LGTM.

PR #650, opened 2025-12-24
feat: add check for unscaled ElectricalSeries data

### LINKED ISSUE #395: [Add Check]: ElectricalSeries of dtype int16 with conversion=1.0 and offset=0.0 is potentially wrong?

bendichter:
Add `check_electrical_series_unscaled_data` to warn when ElectricalSeries has integer data type with default conversion (1.0) and offset (0.0), which suggests raw acquisition units are not properly scaled to Volts.

The check also considers `channel_conversion` for per-channel scaling to avoid false positives when proper conversion factors are specified.

Includes documentation for the best practice and changelog entry.

Closes #408
Closes #395

### DIFF
--- a/tests/unit_tests/test_ecephys.py
+++ b/tests/unit_tests/test_ecephys.py
@@ -523,8 +523,19 @@ class TestCheckElectricalSeriesUnscaledData(TestCase):
             rate=30.0,
         )
         result = check_electrical_series_unscaled_data(electrical_series)
-        assert result is not None
-        assert "uint16" in result.message
+        assert result == InspectorMessage(
+            message=(
+                "ElectricalSeries 'elec_series' has data with dtype 'uint16' and "
+                "conversion=1.0 and offset=0.0. This suggests the data is in raw acquisition units "
+                "which is not in Volts. Please set the 'conversion' and/or 'offset' fields to convert "
+                "the data to Volts, or use 'channel_conversion' for per-channel conversion factors."
+            ),
+            importance=Importance.BEST_PRACTICE_VIOLATION,
+            check_function_name="check_electrical_series_unscaled_data",
+            object_type="ElectricalSeries",
+            object_name="elec_series",
+            location="/",
+        )
 
     def test_fail_with_channel_conversion_all_ones(self):
         """Test that int data with channel_conversion all 1.0 triggers a warning."""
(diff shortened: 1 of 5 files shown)


tests/unit_tests/test_ecephys.py
```diff
@@ -422,3 +423,130 @@ def test_check_units_table_duration_second_unit_no_spikes():
...
+            data=np.zeros((100, 3), dtype=np.int16),
+            electrodes=self.electrodes,
+            rate=30.0,
+            channel_conversion=[1.0, 1.0, 1.0],  # All default values
+        )
+        result = check_electrical_series_unscaled_data(electrical_series)
+        assert result is not None
+        assert "int16" in result.message
+
+    def test_pass_with_empty_data(self):
```
h-mayorquin: I feel that we should not allow `ElectricalSeries` with empty data and that this test might break or conflict with a check for non-empty data. What do you think?

tests/unit_tests/test_ecephys.py
```diff
@@ -422,3 +423,130 @@ def test_check_units_table_duration_second_unit_no_spikes():
...
+            name="elec_series",
+            description="desc",
+            data=np.zeros((100, 3), dtype=np.int16),
+            electrodes=self.electrodes,
+            rate=30.0,
+            channel_conversion=[1.0, 1.0, 1.0],  # All default values
+        )
+        result = check_electrical_series_unscaled_data(electrical_series)
+        assert result is not None
+        assert "int16" in result.message
```
h-mayorquin: The convention I have seen here in this repo is to test against the full message.

h-mayorquin (changes requested):
Two minor points and this should be good to go. 

Also, I am thinking that the schema might be improved to reduce the likelihood of errors:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/671

tests/unit_tests/test_ecephys.py
h-mayorquin: Actually, I pushed this to avoid some work for you. Let me know if you want me to revert it.

h-mayorquin (approved):

tests/unit_tests/test_ecephys.py
```diff
@@ -422,3 +423,152 @@ def test_check_units_table_duration_second_unit_no_spikes():
...
+    def test_pass_with_empty_data(self):
+        """Test that empty data does not trigger a warning."""
+        electrical_series = ElectricalSeries(
+            name="elec_series",
+            description="desc",
+            data=np.zeros((0, 3), dtype=np.int16),
+            electrodes=self.electrodes,
+            rate=30.0,
+        )
+        assert check_electrical_series_unscaled_data(electrical_series) is None
```
bendichter:
```suggestion
```

PR #651, opened 2025-12-24
feat: add publication format and DOI validation checks

### LINKED ISSUE #419: [Add Check]: Comma-separated entries in releated_publication should be list of entries

From https://github.com/dandi/dandi-cli/issues/1308#issuecomment-1843671333

_No response_

Yes.

### Code of Conduct

bendichter:
Add two new inspector checks for related_publications metadata:
- check_publication_list_format: detects comma-separated DOIs/URLs that should be separate list entries
- check_publication_doi_resolves: verifies DOI URLs resolve to valid publications via network requests

Includes documentation updates for best practices guidance.

Closes #419

## Motivation


h-mayorquin (commented):
I think the `check_publication_list_format` severity should be a BestPracticeViolation or even a CRITICAL one.

The [schema definition for `related_publications`](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/a6b23f073f749538eccd33eebcc25533afd0ad08/core/nwb.file.yaml#L219-L226) specifies this field as a 1-D array with `dims: [num_publications]` and `shape: [null]`. Each element in the array is meant to be one publication reference. When someone stuffs multiple DOIs into a single comma-separated string, they are violating the intended data structure, which I think warrants a stronger severity. I am also unable to think of false positives that would justify making this a weaker severity. What do you think?

Regarding `check_publication_doi_resolves`, I am wary of adding a network-based check. I think this can fail in many ways even if the DOI is valid:

1. The user has no internet connection or is running the inspector offline.
2. The user is behind a firewall or proxy that blocks outbound HTTP (common in institutional and CI environments).
3. `doi.org` is temporarily down, rate-limiting, or returning transient errors.
4. DOI blocks certain types of requests when coming from specific places. This happens to some of the link checks on the GitHub CI and I am expecting this will be on the CI pipeline causing problems in the same direction.

In its current state the check cannot distinguish between "the DOI is invalid" and "the network is not reaching it," which means it would report false positives in all of these cases.

A possible pattern to make this more robust: before checking the user's DOIs, first test against a well-known DOI that we know resolves (a canary request). If the canary fails, we know the problem is the network, not the data, and we skip the check entirely. If the canary succeeds, we can trust that any subsequent failures are real. With that safeguard in place, I think the check should be CRITICAL as URLs that do not resolve are clearly wrong, right?

What do you say if I open another PR with the `check_publication_list_format` so we merge that quickly and we think a bit more about this as network checks are uncharted territory for this repo?

bendichter:
Per @h-mayorquin's suggestion above, I've split the `check_publication_list_format` check out into its own PR so it can be merged quickly: #710. It uses the stronger `BEST_PRACTICE_VIOLATION` severity as discussed.

The network-based `check_publication_doi_resolves` check still needs more design discussion (handling offline/firewall/transient-failure false positives, the canary-request safeguard), so closing this PR in favor of #710. We can open a fresh PR for the DOI-resolution check once the approach is settled.

PR #653, opened 2025-12-24
feat: add check for starting_frame without external_file in ImageSeries

### LINKED ISSUE #235: [Add Check]: imageseries arg check

If there is no external file, there should be no starting frame.

_No response_

Yes.

### Code of Conduct

bendichter:
Add check_image_series_starting_frame_without_external_file to detect when starting_frame is set on an ImageSeries without external_file. The starting_frame attribute is only relevant when using external files, and setting it otherwise indicates a legacy issue from older PyNWB versions.

Includes documentation, unit tests, and changelog entry.

Closes #235

### DIFF
--- a/src/nwbinspector/checks/_image_series.py
+++ b/src/nwbinspector/checks/_image_series.py
@@ -78,3 +78,22 @@ def check_image_series_data_size(image_series: ImageSeries, gb_lower_bound: floa
         return InspectorMessage(message="ImageSeries is very large. Consider using external mode for better storage.")
 
     return None
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
+def check_image_series_starting_frame_without_external_file(image_series: ImageSeries) -> Optional[InspectorMessage]:
+    """
+    Check if starting_frame is set when external_file is not used.
+
+    The starting_frame attribute is only relevant when using external files.
+    If there is no external file, there should be no starting_frame.
+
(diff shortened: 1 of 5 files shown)


h-mayorquin (approved):
LGTM. This will detect old files with this problem

If we get to the point where performance starts to matter for this task we should skip this check if we detect the version of the file is larger than  PyNWB>=2.2.0.

src/nwbinspector/checks/_image_series.py
```diff
@@ -78,3 +78,26 @@ def check_image_series_data_size(image_series: ImageSeries, gb_lower_bound: floa
...
     return None
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
+def check_image_series_starting_frame_without_external_file(image_series: ImageSeries) -> Optional[InspectorMessage]:
+    """
+    Check if starting_frame is set when external_file is not used.
+
+    The starting_frame attribute is only relevant when using external files.
+    If there is no external file, there should be no starting_frame.
```
h-mayorquin, line 89:
```suggestion
    If there is no external file, there should be no starting_frame.
    
    Note: it is only possible to create such a file with PyNWB<2.2.0
```
Maybe we should write this here for the benefit of our future selves. Following on the comment here:
https://github.com/NeurodataWithoutBorders/nwbinspector/issues/235

bendichter: This does not account for files that can be created outside of PyNWB.

PR #654, opened 2025-12-24
docs: add CLI documentation for --ignore and --select flags

### LINKED ISSUE #645: Improve documentation for skipping heuristic based checks
On #636 we added a heuristic based check for session duration that should raise a critical error when the session is longer than a year. While this makes sense and should work in most cases it is still a heuristic and we should provide better documentation on how to skip them if a session with those characteristics actually shows up.

@bendichter already has some suggestions that makes sense but I am opening an issue so we don't forget this:

https://github.com/NeurodataWithoutBorders/nwbinspector/pull/635#issuecomment-3682439084

bendichter:
Add new sections to the CLI user guide explaining how to:
- Skip specific checks using the --ignore flag with single or comma-separated check names
- Run only specific checks using the --select flag

Include a note about heuristic-based checks that may produce false positives for edge cases like long-running experiments.

fix #645


h-mayorquin (approved):
LGTM.

I think it could be improved by mentioning how to discover test names but this is good as it stands.

PR #661, opened 2026-02-05
Nwb schema in register checks

bendichter:
Add args to `register_check` so the checks can be filtered by what version of the NWB schema they apply to

### DIFF
--- a/src/nwbinspector/_nwb_inspection.py
+++ b/src/nwbinspector/_nwb_inspection.py
@@ -15,6 +15,8 @@ from tqdm import tqdm
 from ._configuration import configure_checks
 from ._registration import Importance, InspectorMessage, available_checks
 from .tools._read_nwbfile import read_nwbfile, read_nwbfile_and_io
+from packaging import version
+
 from .utils import (
     OptionalListOfStrings,
     PathType,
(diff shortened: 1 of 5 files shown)


src/nwbinspector/_nwb_inspection.py
```diff
@@ -428,7 +433,21 @@ def run_checks(
     else:
         check_progress = checks
 
+    # Get NWB schema version from the nwbfile's read_io if not provided
+    if nwb_schema_version is None:
+        nwb_version_info = getattr(getattr(nwbfile, "read_io", None), "nwb_version", None)
```
h-mayorquin, line 439: is this defensive cod really necessary, when is this None?

h-mayorquin (commented):
This looks good to me. As far as I can the version of the nwbfile will always be available making the code more readable. What do you think of removing that assumption?

src/nwbinspector/_nwb_inspection.py
```diff
@@ -416,6 +418,9 @@ def run_checks(
         Defaults to not displaying progress per set of checks over an individual file.
     progress_bar_options : dict, optional
         Dictionary of keyword arguments to pass directly to the `progress_bar_class`.
+    nwb_schema_version : packaging.version.Version, optional
+        The NWB schema version of the file being inspected.
+        If not provided, will be read from nwbfile.read_io.nwb_version if available.
```
bendichter, line 423:
```suggestion
        If not provided, will be read from nwbfile.read_io.nwb_version if available.
        This arg is mostly used for tests. Usually it is best to leave as None.
```

h-mayorquin (approved):

PR #667, opened 2026-02-18
Remove deprecated warning in version submodule

bendichter:
I don't see a need to deprecate this. This is the way most people check for version


h-mayorquin (approved):
I think the common pattern is:

`from nwbinspector import __version__`

Not the one here which is:
`from nwbinspector.version import __version__` 

I do like removing code for the sake of token efficiency for our robot overlords, but I don’t think it matters in this case. I’d say it’s your call whether you remove the code or the warning.

PR #403, opened 2023-09-18
[New Check]: TimeSeries with empty dataset

alessandratrapani:
Related to Issue #324 
add check function for empty dataset

- [x] add check function
- [ ] add unit tests
- [ ] add check function documentation 
- [ ] update CHANGELOG.md

PR #668, opened 2026-02-19
Add check_time_series_data_is_not_empty

h-mayorquin:
This consolidates the proposals from PR #403 and PR #584. The check is registered at the `NWBDataInterface` level to exclude metadata-only containers (`Device`, `Subject`, `ImagingPlane`), and skips `ImageSeries` with `external_file` set where empty data is intentional per the NWB spec. Emptiness is detected via `.size == 0` rather than `len()` to correctly catch multidimensional arrays with zero elements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Added `check_subject_weight` to ensure subject weight follows the form '[numeric] [unit]' (e.g., '2.3 kg' or '10 g'). [#647](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/647)
 * Added `check_image_series_starting_frame_without_external_file` to verify that `starting_frame` is not set when `external_file` is not used in an `ImageSeries`. [#235](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/235)
 * Added `check_sweeptable_deprecated` to detect usage of the deprecated `SweepTable` in NWB files with schema version >= 2.4.0, which should use `IntracellularRecordingsTable` instead. [#657](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/657)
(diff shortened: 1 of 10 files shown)


bendichter:
I think this is over-inclusive. This would check any extension that contains data, which may not be the desired behavior.

h-mayorquin:
> I think this is over-inclusive. This would check any extension that contains data, which may not be the desired behavior.

Do I get it right that the problem is that maybe there are valid uses of empty data and we will be triggering a false positive here?

In that case, what do you think of just having a big list of known neuro data types whoe data should be non-empty? Explicit is better than implicit I guess.

bendichter:
What data types to do imagine listing? If it's less than 4, I'd say just create separate checks for them. That way we can link to different severity, schema versions, etc.

h-mayorquin:
The original issue (#324) was not very explicit.

I was assuming that in general `.data` should not be empty, so I was thinking we could grow the list by exclusion (like we do here with ImageSeries with external files) instead of having a really large inclusion list. I don't have a sense of how to divide this by neurodata type in a way that the severity would be different.

Maybe it is best for us to not spend too much time thinking on edge cases and, like #403, just restrict to `TimeSeries` with the exception of `ImageSeries` that we already include here. What do you think?

bendichter:
I just think we could be getting ourselves into trouble if we try to extend this to all NWBDataInterface types, including those from extensions. If we only want to limit this to TimeSeries, then it should be a check registered to the TimeSeries type.

h-mayorquin:
Yes, I think this makes sense. Let's limit ourselves to TimeSeries

h-mayorquin:
Done.

bendichter (approved):

PR #673, opened 2026-02-23
Allow "/" as valid subject age for unspecified age range

bendichter:
## Summary
- `check_subject_age` required the lower bound of an age range to be a valid ISO 8601 duration, but allowed the upper bound to be empty. Now both sides are treated symmetrically, so `"/"`, `"/P3D"`, and `"P1D/"` are all valid.
- Added two new tests for the newly accepted formats (`"/P3D"` and `"/"`).

## Test plan
- [x] All existing `check_subject_age` and `check_subject_proper_age_range` tests pass
- [x] New tests for `age="/P3D"` and `age="/"` pass


h-mayorquin (approved):
LGTM.

h-mayorquin:
So, "/" means full uncertainty, this is related to:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/672

PR #382, opened 2023-06-28
Fix #375

h-mayorquin:
This should fix #375. 

Give the discussion in #375 this should probably be the only function that is checked. All the other time columns might not be in ascending order for perfectbly acceptable reasons I think.

Issue #375, opened 2023-06-02
Shoud check for ascending series should apply to stop_times of the interval?


h-mayorquin:
I have the following situation:

```
sleep_states = nwbfile.processing["behavior"]["SleepStates"]
sleep_states.to_dataframe().head(8)
```

Which outputs the following:

| id | start_time | stop_time | label                |
|----|------------|-----------|----------------------|
|  0 |          1 |         1 | WAKEtheta_ThDt       |
|  1 |          1 |         3 | QWake_ThDt           |
|  2 |          1 |         3 | QWake_noRipples_ThDt |
|  3 |          1 |        50 | WAKEnontheta         |
|  4 |          1 |       100 | WAKEstate            |
|  5 |          3 |         4 | WAKEtheta_ThDt       |
|  6 |          5 |        13 | QWake_ThDt           |
|  7 |          5 |        13 | QWake_noRipples_ThDt |

And it is flagged by the inspector with the following warning:
```
0.1  fCamk1_200827_sess9.nwb: check_time_interval_time_columns - 'TimeIntervals' object with name 'SleepStates'
       Message: ['stop_time'] are time columns but the values are not in ascending order. All times should be in seconds with respect to the session start time.
```

Looking at the check it seems that it assumes that all time columns should be ascending which does not seem to be true for my case. 

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/2ddada2cf7481127d35616ece0e6e7d17e332d29/src/nwbinspector/checks/tables.py#L47-L71

Am I missing something? What do you think?

bendichter:
Hmm yes I guess if you have overlapping intervals it might be impossible for start and stop time to both be ascending. That appears to be the case here. What importance level is this in the inspector?

h-mayorquin:
```

0  BEST_PRACTICE_VIOLATION
==========================
...
0.1  fCamk1_200827_sess9.nwb: check_time_interval_time_columns - 'TimeIntervals' object with name 'SleepStates'
       Message: ['stop_time'] are time columns but the values are not in ascending order. All times should be in seconds with respect to the session start time.
...

````

bendichter:
OK, maybe we should disable this for `stop_time`?

h-mayorquin:
Yes, that makes sense to me. What about other columns?

CodyCBakerPhD:
Very interesting

Solution would indeed just be to skip the `stop_time` column for that check

[...]

PR #678, opened 2026-03-13
Add check: flag if all start_time values are 0

### LINKED ISSUE #677: [Add Check]: flag if all start and stop times of a trial table are 0

start times should not be all 0
start times should be ascending
stop times should be greater than their corresponding start time (if non-nan)

```shell

```

Yes.

### Code of Conduct

bendichter:
## Summary
- Adds `check_time_intervals_start_time_all_zero` (CRITICAL) to flag `TimeIntervals` tables where all `start_time` values are 0, indicating times were likely not set relative to session start
- Skips tables with 0 or 1 rows (a single trial at t=0 is normal)
- Updates `check_time_intervals_stop_after_start` to also flag when stop_time == start_time (zero-duration intervals)
- The other check from #677 (ascending start times) already exists as `check_time_interval_time_columns`

Closes #677

## Test plan
- [x] `test_check_time_intervals_start_time_all_zero_fail` — two rows both at 0
- [x] `test_check_time_intervals_start_time_all_zero_pass` — normal ascending times
- [x] `test_check_time_intervals_start_time_all_zero_pass_empty` — empty table
- [x] `test_check_time_intervals_start_time_all_zero_pass_single_row` — single row at 0
- [x] `test_check_time_intervals_stop_equal_start` — stop_time == start_time flagged


src/nwbinspector/checks/_tables.py
```diff
@@ -101,7 +132,7 @@ def check_time_intervals_stop_after_start(
     if np.any(
         np.asarray(cache_data_selection(data=time_intervals["stop_time"].data, selection=slice(nelems)))
         - np.asarray(cache_data_selection(data=time_intervals["start_time"].data, selection=slice(nelems)))
-        < 0
+        <= 0
```
h-mayorquin, line 146: Good catch.

h-mayorquin (commented):
What do you think of just making check_time_interval_time_columns strictly ascending rather than non-decreasing? This should cover not only this problem but also and similar ones.
                                                                                                                                      
That said, I think it is good to have this very specific error as it gives a very actionable hint to the user of what went wrong. We should have a way of running checks in order though so the more specific errors fail and are displayed first for the benefit of the user.

h-mayorquin (approved):
I expanded the check from all-zeros to all-identical `start_time` values and renamed it to `check_time_intervals_start_time_not_constant` (following the convention of naming checks after the correct state). A table with `start_time = [5.0, 5.0, 5.0]` is just as broken as all zeros, so the check now compares against `start_times[0]` instead of hardcoding 0. Added a test for the non-zero case.

I also looked into my earlier suggestion of making `check_time_interval_time_columns` strictly ascending, but that would false-positive on overlapping state annotations like the `SleepStates` table from issue #375, where multiple concurrent states share the same `start_time`. So I kept the non-strict ascending (`>=`) there and added a docstring explaining why the check only covers `start_time` despite the function name (the history from PR #382).

Updated the "Timing Columns" best practice docs to cover both `start_time` (non-decreasing, not constant) and `stop_time` (must be strictly greater than `start_time`), with common signs of unconverted times.

bendichter:
I don't know about this change. I don't think it's crazy for a `TimeIntervals` table to have overlapping intervals, and I also don't think it's crazy for intervals to have the same start time.

h-mayorquin:
> , and I also don't think it's crazy for intervals to have the same start time.

Just to be clear, this check only flags when all start_time values are identical. Overlapping intervals with different start times pass fine, and intervals sharing a start time also pass as long as not every single one is the same value. 

What kind of case are you envisioning where having all start times be the same value would make sense?

bendichter:
ah ok, let me take another look

Issue #676, opened 2026-03-13
better inspection for dandiset 1610


bendichter:
### What would you like to see added to the NWBInspector?

There are a numbers of problems with this dataset, yet the dataset passed with no critical errors. Problems with the dataset include:
* The spike times are mostly negative
* The spike times are not ascending
* The spike times have a duration of 7.32 seconds
* The trials have a start and stop time of 0
* The other times in the trials table are clearly relative

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/CODE_OF_CONDUCT.rst)
- [x] Have you ensured this feature was not already [requested](https://github.com/neurodatawithoutborders/nwbinspector/issues)?

h-mayorquin:
What is 1610? Is it a [dandiset](https://dandiarchive.org/dandiset/001610?search=1610&pos=1)?

bendichter:
yes

h-mayorquin:
Ok, I took a look into this. I downloaded and inspected the full dandiset and the issues are consistent across the files. I think we could have avoided this by not accepting descending spike times, which are currently flagged as BEST_PRACTICE_VIOLATION but should be critical (see PR #684).

Problem and analysis by neurodata type:

**Units (spike_times).** Spike times are trial-relative (centered on a trial event, ranging [-7, +7] seconds) rather than session-absolute. Within each unit, spikes from all ~570 trials are concatenated, causing times to descend at every trial boundary (e.g., 5.8 -> -5.9). About 50% of spikes are negative. Resolution is not set.

Within a single unit, the stored spike times look like:

```
Trial 1 spikes:  [-6.5, -3.2, -1.0, 0.5, 2.1, 5.8]    (relative to trial 1 event)
Trial 2 spikes:  [-5.9, -2.8, 0.1, 1.3, 4.7, 6.9]      (relative to trial 2 event)
Trial 3 spikes:  [-6.1, -4.0, -0.3, 2.5, 5.1]           (relative to trial 3 event)
...
Concatenated:    [-6.5, -3.2, -1.0, 0.5, 2.1, 5.8, -5.9, -2.8, 0.1, 1.3, ...]
                                                    ^^^^
                                          descends here (5.8 -> -5.9) at trial boundary
```

[...]

PR #680, opened 2026-03-13
Add check for spike times duration much shorter than recording

### LINKED ISSUE #679: [Add Check]: flag if spike times duration is suspiciously short

Flag when the total duration of spike times in a Units table is suspiciously short (e.g., 7.32 seconds), which may indicate that spike times were stored in sample indices rather than seconds.

Sub-issue of #677.

Compare the duration of spike times against the TimeSeries durations in the NWBFile. If the spike times duration is much shorter than the recording duration, flag it.

Yes.

### Code of Conduct

bendichter:
## Summary
- Adds `check_units_spike_times_short_duration` (BEST_PRACTICE_VIOLATION) that flags when spike times duration is less than 10 seconds, which may indicate spike times are not aligned to session start.
- Refactors `check_units_table_duration` to use the new `Units.get_duration()` method from PyNWB.

**Note:** This PR depends on the next release of PyNWB, which adds the `Units.get_duration()` method.

Closes #679

## Test plan
- [x] `test_check_units_spike_times_short_duration_fail` — spike times span 7.32s
- [x] `test_check_units_spike_times_short_duration_pass` — spike times span more than 10s
- [x] `test_check_units_spike_times_short_duration_pass_no_spike_times` — Units table without spike_times
- [x] All existing `check_units_table_duration` tests still pass after refactor to `get_duration()`

### DIFF
--- a/src/nwbinspector/checks/_ecephys.py
+++ b/src/nwbinspector/checks/_ecephys.py
@@ -251,6 +251,36 @@ def check_units_table_duration(
     return None
 
 
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=Units)
+def check_units_spike_times_short_duration(
+    units_table: Units, duration_threshold: float = 10.0
+) -> Optional[InspectorMessage]:
+    """
+    Check if spike times duration is suspiciously short.
+
+    Parameters
+    ----------
+    units_table : Units
+        The Units table to check.
+    duration_threshold : float, optional
+        The minimum expected duration in seconds. Default is 10.0 seconds.
+    """
+    duration = units_table.get_duration()
+    if duration is None:
+        return None
+
+    if duration < duration_threshold:
+        return InspectorMessage(
+            message=(
+                f"The spike times have a duration of {duration:.2f} seconds. "
+                "This may indicate that spike times are not aligned to the session start time "
+                "or that there is a data quality issue."
+            )
+        )
+
+    return None
+
+
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=NWBFile)
 def check_electrodes_location_allen_ccf(nwbfile: NWBFile) -> Optional[Iterable[InspectorMessage]]:
     """
@@ -260,7 +290,6 @@ def check_electrodes_location_allen_ccf(nwbfile: NWBFile) -> Optional[Iterable[I
 
     Best Practice: :ref:`best_practice_ecephys_ontologies`
     """
-<<<<<<< HEAD
     if nwbfile.subject is None:
         return None
     species = nwbfile.subject.species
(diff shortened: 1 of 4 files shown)


src/nwbinspector/checks/_ecephys.py
```diff
@@ -232,42 +232,10 @@ def check_units_table_duration(
...
-    spike_times_data = units_table["spike_times"].target.data
-
-    # needed to get tests to work on example data that is a list, not an h5py dataset
-    if isinstance(spike_times_data, list):
-        spike_times_data = np.array(spike_times_data)
-
-    boundary_spike_times = spike_times_data[all_indices]
-
-    if len(boundary_spike_times) == 0:
+    duration = units_table.get_duration()
```
h-mayorquin, line 235: I suspect the tests are failing because this was not released.

src/nwbinspector/checks/_ecephys.py
```diff
@@ -283,6 +251,36 @@ def check_units_table_duration(
...
+    duration_threshold : float, optional
+        The minimum expected duration in seconds. Default is 10.0 seconds.
+    """
+    duration = units_table.get_duration()
+    if duration is None:
+        return None
+
+    if duration < duration_threshold:
+        return InspectorMessage(
+            message=(
```
h-mayorquin, line 274: I am unsure what failure mode are we trying to detect, the one that stated in the issue will result in overly large sipke times not short ones as here:

https://github.com/NeurodataWithoutBorders/nwbinspector/issues/679#issue-4070602933

And duration of spike train should be time invariant to session start time offsets. 

Could you clarify that?

bendichter: This is based on Dandiset 1610, which has this problem. Most of the spikes were negative and the max was like 7 seconds. I think what happened was spikes were aligned to a trial event, not the session start time. This error could also occur if someone tries to convert from milli/micro/nano seconds to seconds and accidentally does it twice.

h-mayorquin (commented):
I think this is not necessary and coule be closed. TFirst, part of this is already detected by the existing [`check_negative_spike_times`](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/889acab720f349ad1181557df8aea8afdb8f5e18/src/nwbinspector/checks/_ecephys.py#L20-L33). Moreover, in the context of [dandiset 001610](https://dandiarchive.org/dandiset/001610), as I mentioned in [the issue about the dandiset](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/676#issuecomment-4121031673), the most important fix is to flag descending spike times as a critical error (see #684).

I do think it would be good to have a check for detecting whether spike times are offset to something other than the session start time (see #688) but I am unsure this is good enough to fit the bill. The duration heuristic has some edge cases: the dandiset 1610 spike times span [-7, +7] = 14 seconds, which actually exceeds the 10-second threshold, so this check wouldn't have caught the motivating case. It would also flag legitimately short recordings as false positives, and miss trial-relative spikes with longer trial windows (e.g., 30-second trials spanning [0, 30] seconds).

PR #681, opened 2026-03-18
Update requirements to HDMF 5

### LINKED ISSUE #675: Compatibility with HDMF 5: DynamicTableRegion constructor validation breaks tests
## Description

HDMF 5 added constructor-time validation for `DynamicTableRegion` that checks all index values in `data` are valid row indices for the referenced `table`. When creating new objects (not reading from file), an `IndexError` is now raised if any index is out of bounds. When reading from file (construct mode), only a warning is issued to maintain backwards compatibility.

This breaks two NWB Inspector tests that intentionally create invalid `DynamicTableRegion` objects to test the inspector's own validation:

- `test_check_dynamic_table_region_data_validity_gt_len` — `data=[0, 20]` where `20` exceeds the table length
- `test_check_dynamic_table_region_data_validity_lt_zero` — `data=[-1, 0]` where `-1` is negative

With HDMF 4, creating these invalid objects was allowed, and NWB Inspector would catch the bad data. With HDMF 5, HDMF itself prevents creating these invalid objects, so the tests fail during setup before the inspector checks can run.

## Traceback

```
E           IndexError: DynamicTableRegion values [20] are out of bounds for <class 'hdmf.common.table.DynamicTable'> 'test_table'.
```

```
E           IndexError: DynamicTableRegion values [-1] are out of bounds for <class 'hdmf.common.table.DynamicTable'> 'test_table'.
```

## Suggested fix

The tests should create the invalid `DynamicTableRegion` objects by bypassing HDMF's constructor validation (e.g., writing invalid data to an NWB file and reading it back, or constructing the object in construct mode). HDMF only warns (rather than raising) when reading from file, so the inspector checks are still useful in that context — which is the real-world scenario the inspector is designed for.

## Related

- NeurodataWithoutBorders/pynwb#2171
- NeurodataWithoutBorders/pynwb#2172

rly:
HDMF 5 added constructor-time validation for `DynamicTableRegion` that raises `IndexError` for out-of-bounds indices. Pass `validate_data=False` in tests that intentionally create invalid `DynamicTableRegion` objects to test the inspector's own validation checks.

Closes #675

Note that pynwb does not yet support HDMF 5 so the dependencies cannot yet be resolved except in local installations where tests pass. This is a WIP until the next pynwb release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 
 ### Improvements
+* Updated HDMF dependency to version 5+. [#675](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/675)
 
 ### Fixes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ classifiers = [
 requires-python = ">=3.10"
 dependencies = [
     "pynwb>=3.1",   # NWB Inspector should always be used with most recent minor versions of PyNWB
+    "hdmf>=5",
     "hdmf-zarr",
     "fsspec",
     "requests",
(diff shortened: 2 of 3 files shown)


rly:
Superseded by #707

PR #683, opened 2026-03-24
Add check for spike times stored as sample indices

h-mayorquin:
This adds `check_spike_times_not_in_samples` (CRITICAL) to detect when spike times are stored as sample indices rather than seconds. The check flags spike times that are all integer-valued, since real spike times in seconds have fractional parts at any common electrophysiology sampling rate.

The existing `check_units_table_duration` catches this error when the apparent duration exceeds 1 year, but 15-minute recordings (common in ephys) at 30 kHz produce sample indices that look like 312 days of "seconds," slipping under that threshold. The integer-valued heuristic catches these cases regardless of recording length. The check samples the last 200 spike time values for efficiency, which is sufficient because the frames-to-seconds conversion error is all-or-nothing.

If the `resolution` field on the Units table is set to >= 1.0, the check is skipped. This serves as an escape hatch for the unlikely case where spike time resolution is truly 1 second or lower. The error message instructs users to set `Units(resolution=1.0)` if this applies to them.

Relates to #679 and #676

### DIFF
--- a/src/nwbinspector/checks/_ecephys.py
+++ b/src/nwbinspector/checks/_ecephys.py
@@ -34,9 +34,7 @@ def check_negative_spike_times(units_table: Units) -> Optional[InspectorMessage]
 
 
 @register_check(importance=Importance.CRITICAL, neurodata_type=Units)
-def check_spike_times_not_in_samples(
-    units_table: Units, nelems: Optional[int] = 200
-) -> Optional[InspectorMessage]:
+def check_spike_times_not_in_samples(units_table: Units, nelems: Optional[int] = 200) -> Optional[InspectorMessage]:
     """
     Check if spike times appear to be sample indices rather than seconds.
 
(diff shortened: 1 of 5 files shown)


src/nwbinspector/checks/_ecephys.py
```diff
@@ -33,6 +33,55 @@ def check_negative_spike_times(units_table: Units) -> Optional[InspectorMessage]
...
+    spike_times_data = units_table["spike_times"].target.data
+
+    if isinstance(spike_times_data, list):
+        spike_times_data = np.array(spike_times_data)
+
+    # Sample the last nelems values
+    if nelems is not None and len(spike_times_data) > nelems:
+        sample = np.array(spike_times_data[-nelems:])
+    else:
+        sample = np.array(spike_times_data[:])
```
bendichter:
```suggestion
    sample = spike_times_data[:nelems]
```

I think you can just do this. If `nelems` is `None` it will be ignored. Is there a reason you wanted to pull from the end and not the beginning?

h-mayorquin: Yes.

I simplified to `spike_times_data[:nelems]`. 

I think I was sampling at the end at some point because I was using it for estimating length but that is no longer the case and I think it is better for consistency to do this.

bendichter (approved):

PR #685, opened 2026-03-24
Add check_spatial_series_unit for SpatialSeries unit validation

h-mayorquin:
SpatialSeries has a free-form `.unit` field that defaults to `"meters"` in PyNWB but accepts any string. The NWB spec says it "should be SI unit" but nothing enforces this, so files in practice contain values like `"m"`, `"cm"`, `"unknown"`, or empty strings. This check flags unrecognized unit strings to push the ecosystem toward consistent values.

The valid unit allowlist uses full SI words in plural form, matching the PyNWB convention: `"meters"`, `"centimeters"`, `"millimeters"`, `"micrometers"`, `"degrees"`, `"radians"`, and `"pixels"`. I chose full words over abbreviations to be consistent with what PyNWB produces as defaults and what the existing `check_compass_direction_unit` already expects. SI length prefixes are included following the same principle as `check_subject_weight`, which accepts the full range of SI mass prefixes (`kg`, `g`, `mg`, etc.).

The check skips SpatialSeries objects inside a CompassDirection container since those are already covered by `check_compass_direction_unit`.

Related to #208 (general SI unit validation). This PR addresses one specific type rather than the broader question of enforcing SI conventions across all TimeSeries unit fields.


bendichter:
This is tricky since sometimes the units are really unknown or arbitrary, but I suppose that's fine if it's just a violation and not a critical.

bendichter (approved):

h-mayorquin:
> This is tricky since sometimes the units are really unknown or arbitrary, but I suppose that's fine if it's just a violation and not a critical.

Maybe we can add a escape hatch of allowing them to intentionally write "n.a." in case that is the case, what do you think?

h-mayorquin:
Related:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/672
And
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/675

bendichter:
yeah I like the idea of allowing n.a.

PR #686, opened 2026-03-24
Add check that Units resolution is set

h-mayorquin:
The `resolution` field on the Units table indicates the smallest possible difference between two spike times, in seconds (i.e., `1/sampling_rate`). This field is almost never set in practice but it carries meaningful information about the precision of spike timing data. Downstream users need this to judge whether fine-timescale analyses are appropriate for a given dataset, for example, cross-correlograms for monosynaptic connection detection require sub-millisecond timing precision that not all recording systems provide.

This PR adds `check_units_resolution_is_set`, which flags at `BEST_PRACTICE_VIOLATION` level when a Units table has spike_times but resolution is not a meaningful positive float (None, NaN, or <= 0). The message tells users exactly what to set and why: `Units(resolution=1/30000)` for a 30 kHz system.

Relates to https://github.com/NeurodataWithoutBorders/nwb-schema/pull/666, #676, #683.

c.c @alejoe91

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### New Checks
 
 * Added `check_imaging_plane_location_allen_ccf`, `check_electrodes_location_allen_ccf`, and `check_intracellular_electrode_location_allen_ccf` to validate location fields against Allen Mouse Brain CCF ontology terms when subject species is mouse. [#671](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/671)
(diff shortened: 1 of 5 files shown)


bendichter:
Let's go even further here. We know reasonable range for resolution would be. I'm worried users will enter sampling rate instead of resolution. Let's say anything > .01 (sampling rate of 100) should be flagged as probably incorrect. The most common error this would catch is if someone accidentally enters 30000.

h-mayorquin:
> Let's go even further here. We know reasonable range for resolution would be. I'm worried users will enter sampling rate instead of resolution. Let's say anything > .01 (sampling rate of 100) should be flagged as probably incorrect. The most common error this would catch is if someone accidentally enters 30000.

Yes, this makes sense. I added tests in this direction.

bendichter (approved):

PR #689, opened 2026-03-24
Add `check_spike_times_without_nans`

h-mayorquin:
Spike times are physical event timestamps representing when action potentials occurred. A spike either happened at a specific time or it didn't, so NaN is never a meaningful value. NaN in spike times typically indicates a conversion bug or array padding that was not cleaned up before writing to NWB.

Unlike other spike time checks that sample the first N elements, this check scans the full array because NaN values can appear at any position in the data.


docs/checks_by_importance.rst
h-mayorquin, line 1: @bendichter was this added accidentally?

bendichter: I think so, yes

rly: I'm deleting the file

rly:
This looks good to me. One minor concern is that as recordings get bigger, both `check_spike_times_without_nans` and `check_ascending_spike_times` load the whole array into memory and will take some time to run.

Typical large recording today:
- ~1,000–5,000 units (multi-probe Neuropixels)
- mean rate ~5 Hz, session ~1–2 h (~3,600–7,200 s)
- → 5,000 × 5 × 7,200 ≈ 1.8e8 spikes

Some labs are doing many-hour recordings, so on the order of 1e9 spike times is plausible.

With 1e9 spike times:

`data[:]` materializes 8 GB (float64) and `np.isnan` allocates another ~1 GB bool array, so ~9 GB transient. On a 16 GB laptop running the inspector that risks an OOM.

On my Mac M1 laptop, it would take about 35 seconds to read gzip-compressed `data[:]` + check, with a warm cache.

I think that's OK, but something to keep in mind as data gets bigger.

rly (approved):

bendichter (approved):

PR #710, opened 2026-06-30
feat: add check_publication_list_format check

### LINKED ISSUE #419: [Add Check]: Comma-separated entries in releated_publication should be list of entries

bendichter:
Split out from #651 per @h-mayorquin's [review suggestion](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/651#issuecomment) to merge the non-network check quickly while the DOI-resolution (network-based) check gets more design discussion.

## Changes

- Adds `check_publication_list_format`, which detects comma-separated DOIs/URLs stuffed into a single `related_publications` entry (e.g. `"https://doi.org/10.1234/abc,https://doi.org/10.5678/def"`) that should instead be separate list entries.
- Per review feedback, severity is set to `BEST_PRACTICE_VIOLATION` (stronger than the original `BEST_PRACTICE_SUGGESTION`): the `related_publications` schema defines a 1-D array where each element is one publication reference, so combining multiple into one comma-separated string violates the intended data structure, and there are no obvious false positives.
- Includes unit tests, CHANGELOG entry, and best-practices documentation.

The network-based `check_publication_doi_resolves` check from #651 is intentionally **not** included here; it needs further discussion (offline/firewall/transient-failure false positives, canary-request safeguard).

fix #419


src/nwbinspector/checks/_nwbfile_metadata.py
```diff
@@ -168,6 +168,35 @@ def check_doi_publications(nwbfile: NWBFile) -> Optional[Iterable[InspectorMessa
...
+    if not nwbfile.related_publications:
+        return None
+    for publication in nwbfile.related_publications:
+        publication = publication.decode() if isinstance(publication, bytes) else publication
+        # Check for comma-separated DOIs or URLs within a single entry
+        # Look for patterns like "doi:xxx,doi:yyy" or "https://doi.org/xxx,https://doi.org/yyy"
+        if "," in publication:
+            # Check if the comma appears to separate multiple DOIs/URLs
+            parts = [p.strip() for p in publication.split(",")]
+            doi_indicators = ["doi:", "doi.org/", "dx.doi.org/"]
```
rly:
```suggestion
            doi_indicators = ["doi:", "doi.org/"]
```
Minor: "doi.org/" is a substring of "dx.doi.org/" so "dx.doi.org/" doesn't capture anything and is dead code.

rly (approved):

PR #717, opened 2026-07-14
Add check_image_series_external_file_forward_slashes

### LINKED ISSUE #697: Add check for backslashes in ImageSeries external_file paths
NWB files organized on Windows can end up with backslash separators in `external_file` paths (e.g. `ses-1_image\abc123_external_file_0.mp4`). This happens because `dandi organize` was using `os.path.join` to build these paths, which produces backslashes on Windows. A fix is in progress at [dandi/dandi-cli#1832](https://github.com/dandi/dandi-cli/pull/1832).

However, NWB files with this problem already exist on the archive (e.g. Dandiset 001771), and the issue could reappear from other tools that construct `external_file` paths using platform-dependent path joining.

The existing `check_image_series_external_file_relative` already validates that paths are not absolute, using both `PureWindowsPath` and `PurePosixPath`. A similar check could flag `external_file` values that contain backslashes, since DANDI asset keys always use forward slashes and backslash paths will silently fail when resolved against the archive.

For context, this caused `get_dandi_video_info()` to return empty results for Dandiset 001771, as reported in [catalystneuro/nwb-video-widgets#33](https://github.com/catalystneuro/nwb-video-widgets/issues/33).

bendichter:
Closes #697

## Summary

Adds a check that flags backslashes in `ImageSeries.external_file` paths. These are produced by platform-dependent path joining on Windows (e.g. `ses-1_image\abc123_external_file_0.mp4`) and silently fail to resolve on POSIX systems and against archive asset keys, which always use forward slashes. This is what caused `get_dandi_video_info()` to return empty results for Dandiset 001771.

The check decodes bytes paths like the neighboring checks do, and yields one message per offending entry so a series with several external files reports each bad one.

## Severity

Registered at `BEST_PRACTICE_VIOLATION`, paralleling the sibling `check_image_series_external_file_relative`. A path that is genuinely broken locally is already reported as `CRITICAL` by `check_image_series_external_file_valid`. Happy to raise this to `CRITICAL` if reviewers feel the silent failure on the archive warrants it.

## Changes

- New `check_image_series_external_file_forward_slashes` in `src/nwbinspector/checks/_image_series.py`, exported from `checks/__init__.py`
- New best-practice section in `docs/best_practices/image_series.rst` under the `best_practice_image_series_external_file_forward_slashes` anchor referenced by the docstring
- CHANGELOG entry under "New Checks"

## Testing

Five new unit tests cover the passing forward-slash path, a non-external series, the backslash trigger, a bytestring path, and multiple external files. All tests in `test_image_series.py` pass; `ruff` and `black` are clean. Also ran end-to-end through `inspect_nwbfile` on a written NWB file containing one backslash and one forward-slash `ImageSeries` — the check fires on the backslash one only.

### DIFF
--- a/src/nwbinspector/checks/_image_series.py
+++ b/src/nwbinspector/checks/_image_series.py
@@ -56,6 +56,33 @@ def check_image_series_external_file_relative(image_series: ImageSeries) -> Opti
 
 
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
+def check_image_series_external_file_forward_slashes(
+    image_series: ImageSeries,
+) -> Optional[Iterable[InspectorMessage]]:
+    """
+    Check that the external_file paths specified by an ImageSeries use forward slashes as separators.
+
+    Backslashes are produced by platform-dependent path joining on Windows, but they do not resolve as separators
+    on POSIX systems or against archive asset keys such as those used by DANDI.
+
+    Best Practice: :ref:`best_practice_image_series_external_file_forward_slashes`
+    """
+    if image_series.external_file is None:
+        return None
+    for file_path in image_series.external_file:
+        file_path = file_path.decode() if isinstance(file_path, bytes) else file_path
+        if "\\" in file_path:
+            yield InspectorMessage(
+                message=(
+                    f"The external file '{file_path}' contains a backslash ('\\'). "
+                    "Please use forward slashes ('/') as path separators so the path resolves on all platforms."
+                )
+            )
+
+    return None
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
 def check_image_series_data_size(image_series: ImageSeries, gb_lower_bound: float = 20.0) -> Optional[InspectorMessage]:
     """
     Check if an ImageSeries stored is larger than gb_lower_bound and suggests external file.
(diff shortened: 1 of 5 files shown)


src/nwbinspector/checks/_image_series.py
```diff
@@ -55,6 +55,33 @@ def check_image_series_external_file_relative(image_series: ImageSeries) -> Opti
...
+    """
+    if image_series.external_file is None:
+        return None
+    for file_path in image_series.external_file:
+        file_path = file_path.decode() if isinstance(file_path, bytes) else file_path
+        if "\\" in file_path:
+            yield InspectorMessage(
+                message=(
+                    f"The external file '{file_path}' contains a backslash ('\\'). "
+                    "Please use forward slashes ('/') as path separators so the path resolves on all platforms."
```
bendichter, line 78: add an exclamation point here

h-mayorquin (changes requested):
This looks good to me.

On the severity, I think it should be `CRITICAL`. On the machine that wrote the file the backslash path opens fine, so `check_image_series_external_file_valid` passes and at this level the upload goes through with a warning. The path only breaks on the archive, where the asset keys use forward slashes and the video can't be found by any reader. That is the same outcome as a missing external file, which we already report as `CRITICAL`, and the upload is the one moment where the writer can still fix it with a one-line change. A warning they can scroll past doesn't do that.

Nothing already on the archive changes, as the server only validates metadata, so the only files this blocks are re-uploads of ones that are already broken.

PR #721, opened 2026-08-11
Fix DynamicTableRegion upper-bound validation

AtomicGlance:
## Summary

DynamicTableRegion stores zero-based row indices, so the largest valid value is len(table) - 1. The validator previously allowed an index equal to len(table), which is out of bounds.

This change:

- rejects indices greater than or equal to the target table length;
- updates the diagnostic message to describe the inclusive upper bound; and
- adds a regression test for the exact boundary value.

## Validation

- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/unit_tests/test_tables.py -q — 48 passed
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/unit_tests -q — 294 passed, 7 skipped

### DIFF
--- a/src/nwbinspector/checks/_tables.py
+++ b/src/nwbinspector/checks/_tables.py
@@ -26,10 +26,11 @@ def check_dynamic_table_region_data_validity(
     dynamic_table_region: DynamicTableRegion, nelems: Optional[int] = NELEMS
 ) -> Optional[InspectorMessage]:
     """Check if a DynamicTableRegion is valid."""
-    if np.any(np.asarray(dynamic_table_region.data[:nelems]) > len(dynamic_table_region.table)):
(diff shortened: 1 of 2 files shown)


AtomicGlance:
I verified the DynamicTableRegion boundary fix locally; the focused table tests pass (48 passed), and the available CI checks are green. The PR is ready for review.

AtomicGlance:
Rebased the DynamicTableRegion boundary fix onto the current dev branch and pushed the refreshed branch. The focused boundary tests pass (4 passed), and the change remains limited to the table check and regression coverage.

AtomicGlance:
The branch is now rebased onto the latest `dev` and force-updated. The full table unit module passes locally (53 passed); the boundary fix remains isolated to the intended check and tests.

PR #724, opened 2026-08-27
[New Check] Detect single uncompressed HDF5 chunks

### LINKED ISSUE #722: [Add Check]: check if a dataset is just one big uncompressed chunk

Some datasets, e.g. https://neurodata-type-inheritance-m.neurosift.pages.dev/nwb?url=https%3A%2F%2Fapi.dandiarchive.org%2Fapi%2Fassets%2Fba9fa55d-6b94-46cf-b72f-6135e92e8f45%2Fdownload%2F&dandisetId=001780&dandisetVersion=draft&tab=hdf5, have Datasets that are just a single, uncompressed chunk. If these are medium sized, they would be better as unchunked. If they are large, they would be better as chunked and compressed, but a single uncompressed chunk is certainly suboptimal

```shell

```

Yes.

### Code of Conduct

AtomicGlance:
## Motivation

Closes #722.

Some HDF5 datasets are created with chunking enabled but the chunk shape is large enough to cover the entire dataset. When compression is disabled, this creates a single uncompressed chunk: it loses the flexibility of chunked storage without gaining compression. Contiguous storage is usually a better fit for smaller datasets, while larger datasets should use multiple chunks and compression.

## Changes

- Add `check_single_chunk_dataset_compression` for HDF5 datasets larger than the configurable 50 MB threshold.
- Ignore contiguous datasets, compressed datasets, and datasets that span multiple chunks.
- Emit an actionable suggestion that distinguishes the small/contiguous and large/chunked-compressed choices.
- Add boundary and storage-layout tests, export the check, document it with the other compression checks, and update the changelog.

## Validation

- `python -m ruff check src/nwbinspector/checks/_nwb_containers.py src/nwbinspector/checks/__init__.py tests/unit_tests/test_nwb_containers.py`
- `PYTHONPATH=src python -m pytest tests/unit_tests/test_nwb_containers.py tests/test_check_configuration.py -q` (23 passed)
- `python -m compileall -q src/nwbinspector tests/unit_tests/test_nwb_containers.py`

This contribution was developed with Codex assistance and reviewed against the existing check and test conventions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.7.3 (Upcoming)
 
 ### New Checks
+* Added `check_single_chunk_dataset_compression` to flag large HDF5 datasets stored as a single uncompressed chunk, with guidance to use contiguous storage for smaller data or chunking and compression for larger data. [#722](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/722)
 * Added `check_spike_times_without_nans` to detect NaN values in Units spike times, which indicate conversion bugs or unclean array padding. [#689](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/689)
 * Added `check_subject_age_reference` to validate that `Subject.age__reference`, when present, is one of the supported values (`"birth"` or `"gestational"`). This catches invalid references in files written by tools that do not enforce the schema constraint. [#250](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/250)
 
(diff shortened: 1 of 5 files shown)


AtomicGlance:
Rebased the single-uncompressed-chunk check onto the current dev branch and pushed the refreshed branch. The focused tests in tests/unit_tests/test_nwb_containers.py pass (5 passed for the single-chunk cases). The diff remains limited to the new check, its tests, documentation, and changelog; ready for review.

AtomicGlance:
The branch is now rebased onto the latest `dev` (including the subject-age fix) and force-updated. The full nwb-container unit module passes locally (14 passed); ready for review.

bendichter:
Thanks for taking this on. The check is close, but the 50 MB threshold puts it in the wrong place. Above that size an uncompressed dataset already trips `check_small_dataset_compression` or `check_large_dataset_compression`, so everything this check flags gets two messages, while the medium-sized case from #722, a dataset written as one chunk that would be better contiguous, sits below the threshold and never fires. The message then hedges across both situations.

I am closing this in favor of #765, which keeps a single check about the layout, fires at any size, and picks the message by size so the existing compression checks keep reporting the missing compression on their own.

PR #725, opened 2026-08-29
Check Units waveform electrode dimensions

### LINKED ISSUE #714: Check Units.waveforms electrode dimension against the electrodes column (analogous to check_electrical_series_dims)
### Feature request

Add a check that validates the doubly-ragged `Units.waveforms` column against the Units `electrodes` column, analogous to the existing `check_electrical_series_dims`.

### Motivation

nwbinspector currently has no check that touches `Units.waveforms` at all (the `Units` checks are all about `spike_times`, resolution, and duration). In particular, nothing compares the number of waveforms per spike event (implied by `waveforms_index`) to the number of electrodes a unit is associated with (its `electrodes` entry).

This matters because the `waveforms` input order is easy to get wrong. `add_unit` expects a per-unit 3-D array of shape `(num_spikes, num_electrodes, num_samples)`, but the `add_unit` docstring says the third dimension is electrodes (see NeurodataWithoutBorders/pynwb#2220). Following the docstring transposes the electrode and sample axes, and it fails silently — no exception, and the resulting file is internally inconsistent with the `electrodes` column.

Concrete example (2 electrodes, 10 samples, 3 spikes, 2 electrode references):

```
# CORRECT: (num_spikes=3, num_electrodes=2, num_samples=10)
#   waveforms.data  : (6, 10)
#   waveforms_index : [2, 4, 6]      # 2 electrodes/spike -> matches the 2 electrode refs

# DOCSTRING ORDER (wrong): (num_spikes=3, num_samples=10, num_electrodes=2)
#   waveforms.data  : (30, 2)        # each "waveform" is now length 2, not 10
#   waveforms_index : [10, 20, 30]   # claims 10 electrodes/spike, but electrodes column has 2
```

The broken file has waveforms claiming 10 electrodes per spike while the `electrodes` column has 2. The NWB schema requires the waveforms' electrode dimension to match the `electrodes` column, so this is a detectable inconsistency — but nothing currently detects it.

### Precedent

`check_electrical_series_dims` (`src/nwbinspector/checks/_ecephys.py`) already does the analogous thing for `ElectricalSeries`: it uses the length of the linked electrodes region to detect a transposed data array (both 2-D `ElectricalSeries` and 3-D `SpikeEventSeries`), warning "The second dimension of data does not match the length of electrodes…". A `Units.waveforms` check would be the doubly-ragged equivalent.

### Proposed check

For a `Units` table that has both a `waveforms` column and an `electrodes` column, for each unit:

- Derive the number of waveforms per spike event from the `waveforms_index` / `waveforms_index_index` datasets (i.e. the per-spike-event increment of `waveforms_index` within that unit's range).
- Compare it to the number of electrodes for that unit (`len(units['electrodes'][unit])`).
- If they differ (or the per-spike-event count is not constant within a unit), emit a message that the waveforms' electrode dimension is inconsistent with the `electrodes` column and the data may be transposed.

This can be done from the small index datasets alone (`waveforms_index`, `waveforms_index_index`, and the electrodes index), without loading the full `waveforms` data — so it stays cheap on large files. Importance likely `CRITICAL`, mirroring `check_electrical_series_dims`.

AtomicGlance:
Fixes #714. Adds a critical Units check that compares the per-spike waveform row count from the doubly-ragged waveforms indices with each unit's electrodes references. It reports affected unit rows, catches transposed or inconsistent dimensions, and reports malformed or incomplete index metadata. The implementation reads only the small index arrays and never loads waveform samples. Added regression coverage for valid and transposed layouts, multiple units, inconsistent spike dimensions, missing columns, empty tables, and malformed indices. Validation: 63 focused ecephys tests passed; ruff passed; inspector end-to-end tests passed (24 passed, 22 skipped); HDF5-backed NWB smoke test passed.

### DIFF
--- a/docs/best_practices/ecephys.rst
+++ b/docs/best_practices/ecephys.rst
@@ -111,6 +111,19 @@ Check functions: :py:meth:`~nwbinspector.checks._ecephys.check_units_resolution_
 :py:meth:`~nwbinspector.checks._ecephys.check_units_resolution_is_valid`
 
 
+.. _best_practice_units_waveforms_electrodes:
+
+Waveform Electrode Dimension
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a ``Units`` table stores individual ``waveforms``, each spike's waveform must contain one waveform row for
+every electrode listed in that unit's ``electrodes`` entry. A mismatch usually means the electrode and sample axes
+were transposed when the data was written. Keep the electrode order consistent with the referenced electrode table
+and verify the dimensions before sharing the NWB file.
+
+Check function: :py:meth:`~nwbinspector.checks._ecephys.check_units_waveforms_electrodes`
+
+
 .. _best_practice_spike_times_not_in_samples:
 
 Spike Times in Seconds
--- a/src/nwbinspector/checks/__init__.py
+++ b/src/nwbinspector/checks/__init__.py
@@ -19,6 +19,7 @@ from ._ecephys import (
     check_units_resolution_is_valid,
     check_units_table_duration,
     check_units_table_has_spikes,
+    check_units_waveforms_electrodes,
 )
 from ._general import (
     check_description,
@@ -200,4 +201,5 @@ __all__ = [
     "check_electrodes_location_allen_ccf",
     "check_intracellular_electrode_location_allen_ccf",
     "check_units_table_has_spikes",
+    "check_units_waveforms_electrodes",
 ]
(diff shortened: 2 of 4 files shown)


AtomicGlance:
Rechecked this PR: the Read the Docs and pre-commit checks are passing, and the branch is mergeable. The check stays scoped to waveform/electrode dimensions without loading the full waveform payload. It is ready for review when convenient.

AtomicGlance:
Rebased the branch onto the current dev branch (including the latest zero-row table fixes) and reran tests/unit_tests/test_ecephys.py: 63 passed. The waveform/electrode check remains bounded to ragged indices and does not load waveform payloads. The updated branch is ready for review.

AtomicGlance:
The branch picked up the latest `dev` commit as well and was force-updated again. The full ecephys unit module still passes locally (63 passed); the check remains payload-free and ready for review.

PR #727, opened 2026-09-04
Fix inspect_all with n_jobs > 1 when a config is given

### LINKED ISSUE #726: inspect_all with n_jobs > 1 fails when a config is given
Running the inspector over a folder with both a config and more than one job fails before any file is inspected:

```
nwbinspector path/to/folder --config dandi --n-jobs 2
```

```
AttributeError: Can't get local object 'register_check.<locals>.register_check_and_auto_parse.<locals>.auto_parse_some_output'
```

The same happens from the API with `inspect_all(path, config=load_config("dandi"), n_jobs=2)`.

The cause is in how configured checks are built. `configure_checks` copies each check with `_copy_function` so that its importance can be changed without touching the registry. `FunctionType` sets `__qualname__` from the code object, which for the registered wrapper is the local name inside `register_check`, and `functools.wraps` is not re-applied to the copy. When `inspect_all` submits those copies to a `ProcessPoolExecutor`, pickle tries to locate them by qualified name and fails. Setting `__qualname__` on the copy is not enough on its own, because pickle then finds the original registry function at that name, sees that it is a different object, and refuses to pickle it (and if it did, the worker would receive the unconfigured original and the importance changes would be lost silently).

The existing parallel test never passes a config, which is why CI does not catch this.

I think the fix is to stop sending function objects to the workers at all. The parent can send the list of selected check names along with the config, importance threshold, and module list, and each worker can call `configure_checks` itself to rebuild the same list. That keeps custom checks from external modules working the same way they do today, since the worker imports those modules before configuring.

bendichter:
Fixes #726.

`inspect_all` with `n_jobs > 1` and a config failed with a pickling error before inspecting anything, because the configured checks are function copies made by `configure_checks` and those copies cannot be pickled. Restoring `__qualname__` on the copies is not enough, since pickle would then find the original registry function at that name and refuse it as a different object, and if it did not refuse, the worker would receive the unconfigured original and silently drop the importance changes.

This PR changes what is sent to the workers. The parent now sends the list of selected check names together with the config, the importance threshold, and the module list, and each worker calls `configure_checks` itself to rebuild the same list. Selection, ignore, and SKIP filtering are already resolved into the name list on the parent side, so the worker only needs to reapply the importance mapping. The CLI passes its `--modules` list through to `inspect_all` so that custom checks from external modules are importable in spawn-based workers.

Two tests are added to `TestInspectorAPIAndCLIHDF5` (and inherited by the Zarr variant): one runs the same config serially and in parallel and asserts the results match and that the remapped importances are present, and one checks that a SKIP entry is honored in parallel mode.


h-mayorquin (approved):
LGTM. Now, we will be able to use the config with parallelism.

bendichter:

PR #729, opened 2026-09-04
Guard four table checks against zero-row tables

### LINKED ISSUE #728: Zero-row tables with columns produce ERROR-level messages from four table checks
A DynamicTable that has columns but no rows is valid NWB, and `check_empty_table` already reports it as a best practice violation. Four other table checks do not guard against this case and instead read element 0 of each column, so the report for such a file also contains four ERROR-level messages:

```
During evaluation of 'check_column_binary_capability' - <class 'IndexError'>: Index (0) out of range for empty dimension
During evaluation of 'check_table_values_for_dict' - <class 'IndexError'>: Index (0) out of range for empty dimension
During evaluation of 'check_col_not_nan' - <class 'IndexError'>: Index (0) out of range for empty dimension
During evaluation of 'check_table_time_columns_are_not_negative' - <class 'IndexError'>: Index (0) out of range for empty dimension
```

Reproduction:

```python
import numpy as np
from pynwb import NWBHDF5IO
from pynwb.epoch import TimeIntervals
from nwbinspector import inspect_nwbfile_object, Importance
from nwbinspector.testing import make_minimal_nwbfile

nwbfile = make_minimal_nwbfile()
trials = TimeIntervals(name="trials", description="d")
trials.add_column(name="c", description="d", data=np.array([], dtype=float))
nwbfile.trials = trials
with NWBHDF5IO("empty.nwb", "w") as io:
    io.write(nwbfile)
with NWBHDF5IO("empty.nwb", "r") as io:
    errors = [m for m in inspect_nwbfile_object(io.read()) if m.importance == Importance.ERROR]
    print(len(errors))  # 4
```

An empty trials or electrodes table with predeclared columns is a plausible outcome of a conversion where a session happened to have no events, so this shows up in real reports. The four checks should return early when the table has no rows and leave the reporting to `check_empty_table`.

bendichter:
Fixes #728.

A table with declared columns but no rows caused `check_column_binary_capability`, `check_table_values_for_dict`, `check_col_not_nan`, and `check_table_time_columns_are_not_negative` to raise `IndexError` when they read element 0 of a column, and the inspector reported each of those as an ERROR-level message. The table itself is valid NWB and `check_empty_table` already reports it as a best practice violation, so the other four checks now return early when the table has no rows.

The unit tests cover each of the four checks on an in-memory empty table, and one round-trip test writes an NWB file with an empty trials table and asserts that inspection produces no ERROR-level messages and still reports `check_empty_table`. All five new tests fail on `dev` and pass with this change.


h-mayorquin (approved):
This is good. The four crashes go away and a file with an empty table now gets reported instead of failing outright. I ran every registered check against an empty `DynamicTable`, an empty `TimeIntervals` and an empty `Units` on both branches. Those four are the only ones that raise on `dev` and none of them raise here, so the set is complete.

The message that survives the fix, `check_empty_table` at BEST_PRACTICE_VIOLATION, is worth a separate conversation and I opened #761 for it. Nothing there changes this PR, the crash fix stands on its own.

PR #731, opened 2026-09-04
Read only the identifier during the inspect_all pre-scan

### LINKED ISSUE #730: Identifier pre-scan in inspect_all leaks file handles and reads every file twice
`inspect_all` runs a pre-scan over every file in the folder to find duplicated identifiers. It does this by calling `read_nwbfile` on each file, which builds the full in-memory NWBFile through PyNWB, and it never closes the IO object it gets back:

```python
for nwbfile_path in nwbfiles:
    try:
        nwbfile = read_nwbfile(nwbfile_path=nwbfile_path)
        identifiers[nwbfile.identifier].append(nwbfile_path)
```

I measured one leaked file descriptor per file. The NWBFile, its IO, and the builder tree hold references to each other, so the handles survive until the cycle collector runs, and in my test a later attempt to write to one of the files failed with "unable to truncate a file which is already open". Beyond the leak, this doubles the read cost of every folder inspection, since `inspect_nwbfile` reads each file again a moment later. For folders of large files the pre-scan alone can take minutes.

The identifier is a single scalar dataset at `/identifier`. For HDF5 files it can be read with h5py in a context manager without touching PyNWB at all. For Zarr stores the existing `read_nwbfile` path can stay as a fallback, but the IO should be closed afterwards. The `_MissingHdmfZarrError` behavior (raising when a `.nwb.zarr` store is found but `hdmf-zarr` is not installed) should be preserved.

bendichter:
Fixes #730.

The duplicate-identifier pre-scan in `inspect_all` built the full NWBFile for every file in the folder through `read_nwbfile` and never closed the IO object it got back. Each file left one handle open, and the whole folder was read twice.

The pre-scan now calls a small helper that reads only the `/identifier` dataset. For HDF5 files (detected with `h5py.is_hdf5`) it opens the file with h5py in a context manager, so nothing is built and nothing is left open. Anything h5py cannot open, which in practice means Zarr stores, falls back to `read_nwbfile` and closes the IO in a `finally` block. Because the fallback still goes through `read_nwbfile`, the `_MissingHdmfZarrError` raised for a `.nwb.zarr` store without `hdmf-zarr` installed propagates exactly as before, and `tests/test_missing_hdmf_zarr.py` is unaffected.

The regression test runs `inspect_all` over the unique-identifier fixture folder and then asks the HDF5 library which files it still has open (`h5f.get_obj_ids` with `OBJ_ALL`), asserting that none of them live in the fixture folder. It fails on `dev` and passes here. The existing pass and fail tests for `check_unique_identifiers` cover the identifier values themselves on both backends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
+* Fixed the identifier pre-scan in `inspect_all` leaking one open file handle per file and reading every file through PyNWB before the inspection read it again. HDF5 files now have only their `/identifier` dataset read with h5py, and the fallback reader used for Zarr stores closes its IO object. [#730](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/730)
 * Fixed unit tests that failed at construction time against recent PyNWB and HDMF. [#707](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/707) [#708](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/708)
 * Skipped Ontobee in the documentation external link check, which frequently timed out and caused spurious CI failures. [#709](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/709)
 * Fixed `run_checks` raising `TypeError` when a `progress_bar_class` was passed without `progress_bar_options`; the keyword arguments are now coalesced to an empty dict before being forwarded to the progress-bar constructor. [#701](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/701)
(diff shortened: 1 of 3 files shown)


h-mayorquin (changes requested):
Can you read `/identifier` directly for Zarr too instead of falling back to `read_nwbfile`. The store keeps it as a root dataset the same way HDF5 does. The whole zarr branch is `zarr.open(store=nwbfile_path, mode="r")["identifier"][0]` and the `[0]` instead of `[()]` is because hdmf-zarr writes the value as a shape `(1,)` object array.

I tried it on a folder of five Zarr stores and the pre-scan goes from 97.7 ms to 0.9 ms. You need to keep the `nwbfile_path.endswith(".nwb.zarr")` guard in front of the zarr read, otherwise the `_MissingHdmfZarrError` message that `tests/test_missing_hdmf_zarr.py` asserts on never gets raised.

The changelog line reads as if the double read is fixed for both backends. That needs an edit too if the fallback stays.

Nit: `_read_nwbfile_identifier` does backend detection and file reading and would fit better in `tools/_read_nwbfile.py`.

bendichter:
Done. The Zarr branch now reads `["identifier"][0]` through `zarr.open` instead of building the file, with the `.nwb.zarr` guard in front of it. I pulled the guard out into `_raise_if_hdmf_zarr_is_missing` so the full reader and the pre-scan raise the same message, and I regenerated the fixture and ran `test_missing_hdmf_zarr.py` in an environment without `hdmf-zarr` to confirm it still passes. The helper lives in `tools/_read_nwbfile.py` now, and the changelog line says what is read for each backend.

h-mayorquin (approved):
LGTM

PR #733, opened 2026-09-04
Make four checks safe for in-memory objects

### LINKED ISSUE #732: Four checks raise on in-memory objects (unsaved NWBFile, list-backed SpatialSeries data)
`inspect_nwbfile_object` is documented as working on an in-memory NWBFile, but four checks assume the object came from disk and raise on ordinary in-memory inputs. The exceptions are caught by `run_checks` and reported as ERROR-level messages, so the report for an unsaved file contains spurious errors.

`check_image_series_external_file_valid` builds `Path(get_nwbfile_path_from_internal_object(image_series))`. For an NWBFile that has not been written, `container_source` is `None`, so this raises:

```
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

`check_spatial_series_dims` reads `spatial_series.data.shape`, and `check_spatial_series_radians_magnitude` and `check_spatial_series_degrees_magnitude` compare `spatial_series.data[:nelems]` directly against a float. PyNWB accepts a plain list for `data`, in which case these raise:

```
AttributeError: 'list' object has no attribute 'shape'
TypeError: '>' not supported between instances of 'list' and 'float'
```

Reproduction:

```python
from pynwb.behavior import SpatialSeries
from pynwb.image import ImageSeries
from nwbinspector.checks import (
    check_image_series_external_file_valid,
    check_spatial_series_dims,
    check_spatial_series_radians_magnitude,
)
from nwbinspector.testing import make_minimal_nwbfile

nwbfile = make_minimal_nwbfile()
nwbfile.add_acquisition(
    ImageSeries(name="im", external_file=["movie.avi"], starting_frame=[0], format="external",
                timestamps=[0.0, 1.0], unit="n.a.")
)
list(check_image_series_external_file_valid(nwbfile.acquisition["im"]))  # TypeError

spatial_series = SpatialSeries(name="ss", data=[[1.0, 2.0], [3.0, 4.0]], reference_frame="x", unit="radians", rate=1.0)
check_spatial_series_dims(spatial_series)  # AttributeError
check_spatial_series_radians_magnitude(spatial_series)  # TypeError
```

The external file check cannot resolve a relative path without knowing where the NWB file lives, so it should return `None` when there is no container source. The SpatialSeries checks should go through `get_data_shape` and `np.asarray`, as the other TimeSeries checks already do.

bendichter:
Fixes #732.

`check_image_series_external_file_valid` called `Path()` on the NWBFile's container source, which is `None` for a file that has not been written yet, so `inspect_nwbfile_object` on an unsaved file reported a `TypeError` as an ERROR-level message for every ImageSeries with an external file. The check only validates relative paths, and those cannot be resolved without a location on disk, so it now returns `None` when there is no container source. Absolute paths are still handled by `check_image_series_external_file_relative`.

`check_spatial_series_dims` read `data.shape` directly and the radians and degrees magnitude checks compared a raw slice of `data` against a float. PyNWB accepts a plain list for `data`, which has neither. The dims check now uses `get_data_shape` and the magnitude checks wrap the slice in `np.asarray`, matching what the other TimeSeries checks already do.

Tests cover the in-memory ImageSeries case and both the passing and triggering cases for each of the three SpatialSeries checks with list-backed data. All seven fail on `dev` and pass here. The existing tests with numpy-backed data are unchanged.


h-mayorquin (approved):
This looks good to me.

One thing I noticed while looking at the `get_data_shape` change. Our copy of it exists because hdmf used to return `maxshape` for resizable datasets, but hdmf checks `shape` first since 3.4.0 and returns the same thing as ours on lists, arrays and datasets:
https://github.com/hdmf-dev/hdmf/pull/747

So we don't need our own anymore. It has 18 call sites so I would not fold it here, I can do a follow up after this is merged that deletes it and imports it from hdmf.

bendichter:
`get_data_shape` is useful when the input is a list, which doesn't support .shape

PR #735, opened 2026-09-04
Handle subject ages with more than one slash

### LINKED ISSUE #734: check_subject_age raises ValueError when the age contains more than one slash
`check_subject_age` and `check_subject_proper_age_range` both split the age string on `/` and unpack the result into exactly two names:

```python
subject_lower_age_bound, subject_upper_age_bound = subject.age.split("/")
```

An age such as `"P1D/P2D/P3D"` is malformed, but instead of the usual "does not follow ISO 8601 duration format" message the user gets a `ValueError: too many values to unpack` wrapped as an ERROR-level message, which reads as an inspector bug rather than a problem with the file.

```python
from pynwb.file import Subject
from nwbinspector.checks import check_subject_age
check_subject_age(Subject(subject_id="s", age="P1D/P2D/P3D"))  # ValueError
```

`check_subject_age` should treat any string with more than one slash as an invalid age and return the normal message. `check_subject_proper_age_range` should return `None` in that case, since the format problem is already reported by `check_subject_age`.

bendichter:
Fixes #734.

Both `check_subject_age` and `check_subject_proper_age_range` split the age on `/` and unpacked the pieces into two names, so an age with more than one slash raised `ValueError: too many values to unpack` and showed up in the report as an ERROR-level message. Both checks now only treat a string with exactly one slash as a range. `check_subject_age` falls through to its usual "does not follow ISO 8601 duration format" message for anything else, and `check_subject_proper_age_range` returns `None` since the format problem is already reported.

Two tests are added, one for each check, using `P1D/P2D/P3D`. Both fail on `dev` and pass here.


h-mayorquin (approved):
Makes sense, it should print an inspector error not fail and abort.

PR #490, opened 2024-08-23
Add DANDI Inspector API; replace ros3 with remfile

CodyCBakerPhD:
## Motivation

To make a smooth deprecation cycle of `nwbinspector` CLI with removal of ros3, it will need to route to the intended way of running things, so this needs to be implemented first

PR #737, opened 2026-09-04
Submit all assets before collecting in get_s3_urls_and_dandi_paths

### LINKED ISSUE #736: get_s3_urls_and_dandi_paths parallel branch collects results inside the submit loop
`get_s3_urls_and_dandi_paths` in `nwbinspector/tools/_dandi.py` has a parallel branch for `n_jobs != 1`, but the loop that collects results is indented inside the loop that submits them:

```python
for asset in dandiset.get_assets():
    if asset.path.split(".")[-1] == "nwb":
        futures.append(executor.submit(_get_content_url_and_path, asset=asset, ...))
    for future in as_completed(futures):
        s3_urls_to_dandi_paths.update(future.result())
```

After every submit, the function waits for every future submitted so far, so at most one request is ever in flight and the pool adds overhead without adding concurrency. The parallel branch should submit all assets first and then collect.

The helper it calls also ignores its own arguments:

```python
def _get_content_url_and_path(asset, follow_redirects=1, strip_query=True):
    return {asset.get_content_url(follow_redirects=1, strip_query=True): asset.path}
```

That is harmless with the current callers, which pass the defaults, but it should forward the arguments it accepts.

bendichter:
Fixes #736.

In the `n_jobs != 1` branch of `get_s3_urls_and_dandi_paths`, the loop over `as_completed(futures)` was indented inside the loop that submitted the futures, so after each submit the function drained every outstanding request before submitting the next one. The pool never had more than one request in flight. The branch now submits all NWB assets first and collects afterwards.

Two smaller things came along with it. `_get_content_url_and_path` accepted `follow_redirects` and `strip_query` but hardcoded both in the call, and now forwards them. The function also takes an optional `client`, matching `get_nwb_assets_from_dandiset`, which is what makes it testable without the DANDI API: when a client is passed the `dandi` import and the context manager are skipped.

The tests use a fake client whose assets record wall-clock start and end times in the URL they return. The serial test checks the mapping and that the arguments reach the asset. The parallel test uses four assets with a half-second delay each and two workers, and asserts that at least one pair of requests overlapped in time. I confirmed it fails when the original indentation is put back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
+* Fixed the parallel branch of `get_s3_urls_and_dandi_paths` collecting results inside the submit loop, which serialized every request. The helper it calls now also forwards its `follow_redirects` and `strip_query` arguments, and the function accepts an optional `client` like `get_nwb_assets_from_dandiset`. [#736](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/736)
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
This looks good to me.

I have a question though, is this still used? Nothing inside the inspector calls `get_s3_urls_and_dandi_paths` since #490 moved the streaming path to `get_nwb_assets_from_dandiset` and I don't see it in the docs either. If nobody is calling it maybe we should deprecate it instead of fixing it?

Maybe we should do a clean up of things that are not used so we don't waste cycles on that type of features in the era of agents.

PR #739, opened 2026-09-04
Flag table region indices equal to the target table length

### LINKED ISSUE #738: check_dynamic_table_region_data_validity misses an index equal to the table length
`check_dynamic_table_region_data_validity` flags indices that are greater than the length of the target table:

```python
if np.any(np.asarray(dynamic_table_region.data[:nelems]) > len(dynamic_table_region.table)):
```

Indices are zero-based, so an index equal to the table length is already out of range, and the check lets it through. A DynamicTableRegion pointing at a 10-row table with a value of 10 is reported by HDMF as a warning on read but produces no inspector message. The comparison should be `>=`.

Both the existing test for the "greater than" case and a new test for the boundary value should pass with the change.

bendichter:
Fixes #738.

`check_dynamic_table_region_data_validity` used `>` against the length of the target table, so an index equal to the length, which is the first out-of-range value for zero-based indices, was not reported. The comparison is now `>=`, and the message reads "greater than or equal to the length of the target table" so that it matches the condition.

Two tests are added: one builds a region whose data contains `len(table)` and expects the message, and one uses `len(table) - 1` and expects no message. The first fails on `dev`. The existing test for a value well past the end is updated for the new wording.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM

PR #741, opened 2026-09-04
Keep integer dtype when building units duration indices

### LINKED ISSUE #740: check_units_table_duration fails on signed spike_times_index arrays
`check_units_table_duration` reads the `spike_times_index` array and builds the first and last spike index of each unit like this:

```python
idxs = np.asarray(units_table["spike_times"].data[:])
first_spike_idxs = np.concatenate([[np.uint64(0)], idxs[idxs != idxs[-1]]])
last_spike_idxs = idxs[idxs != 0] - 1
all_indices = np.unique(np.concatenate([first_spike_idxs, last_spike_idxs]))
boundary_spike_times = spike_times_data[all_indices]
```

NumPy promotes `uint64` combined with a signed integer array to `float64`, so if the index array is signed the fancy index at the end raises:

```
IndexError: arrays used as indices must be of integer (or boolean) type
```

HDMF writes `VectorIndex` data with the smallest unsigned dtype that fits, which is why the current tests pass. Files written by other tools, or a `VectorIndex` constructed in memory with `np.int64` data, hit the promotion, and the check is then reported as an ERROR-level message instead of a result.

The zero should be created with the same dtype as the index array so that the concatenation never leaves integer types.

bendichter:
Fixes #740.

`check_units_table_duration` prepends a zero to the `spike_times_index` array to get the first spike of each unit. That zero was a `np.uint64`, and when the index array is a signed integer type NumPy promotes the concatenation to `float64`, so the fancy index into `spike_times` raised (`TypeError: Indexing arrays must have integer dtypes` from h5py). HDMF writes the index with the smallest unsigned dtype, so PyNWB-written files were unaffected, but files from other writers can carry an `int64` index. The zero is now created with the dtype of the index array.

HDMF also casts the index on construction, so an in-memory VectorIndex cannot reproduce this. The tests write a file with PyNWB, rewrite the index dataset as `int64` with h5py (copying its attributes, including the `target` reference), and read it back. One test expects no message for a short recording and one expects the duration message for a long one. Both raise the `TypeError` on `dev` and pass here.


h-mayorquin (approved):
LGTM, this will avoid the implicit casting.

PR #743, opened 2026-09-04
Sample the ends of the series in check_regular_timestamps

### LINKED ISSUE #742: check_regular_timestamps reads the entire timestamps dataset twice
`check_regular_timestamps` passes the whole `timestamps` dataset to `is_regular_series`, which calls `np.diff` on it, and then calls `np.array(time_series.timestamps)` a second time to size the severity:

```python
if (
    time_series.timestamps is not None
    and len(time_series.timestamps) > 2
    and is_regular_series(series=time_series.timestamps, tolerance_decimals=time_tolerance_decimals)
    and (time_series.timestamps[1] - time_series.timestamps[0]) != 0
):
    timestamps = np.array(time_series.timestamps)
```

For an h5py dataset both of those read the entire array into memory. I measured a peak of 36 MB while checking a 2M-sample series whose timestamps dataset is 16 MB. A multi-hour recording at 30 kHz has on the order of 1e9 timestamps, so the check reads 8 GB, twice, per series. When streaming from DANDI this is the dominant cost of an inspection. The other timestamp checks (`check_timestamps_ascending`, `check_timestamps_without_nans`) already limit themselves to the first `nelems` samples.

Sampling only the head would make the check claim regularity for a series that is regular for 200 samples and irregular afterwards, so I would rather keep the answer exact for the common failure modes while reading O(nelems) values: check that the first and last `nelems` samples are each regular with the same step, and that the total span `timestamps[-1] - timestamps[0]` equals `(n - 1) * step` to within half a step. A gap, a rate change, or a jitter section anywhere in the middle changes the total span, so it is still caught. Series shorter than `2 * nelems` can keep being checked in full.

bendichter:
Fixes #742.

`check_regular_timestamps` handed the whole timestamps dataset to `np.diff` and then cast it to an array a second time, so an h5py-backed series was read into memory twice regardless of its length. For a long recording streamed from DANDI this was the most expensive step of the inspection.

The check now reads the first and last `nelems` timestamps (default 200) and reports a constant rate only if both ends have the same constant step and the total span `timestamps[-1] - timestamps[0]` equals `(n - 1) * step` to within half a step. A gap, a rate change, or a jitter section anywhere in the unread middle changes the span, so those cases are still reported correctly. The one thing this cannot see is a single shifted sample in the middle whose displacement leaves the span unchanged; that trade-off is documented in a test and can be avoided by passing `nelems=None`, which reads the full array as before. Series shorter than `2 * nelems` are still read in full, so behavior on small series is unchanged. The severity threshold now uses the dataset length and dtype rather than a materialized copy.

Tests cover the long regular series, a gap in the middle, a rate change at the midpoint, the documented jitter trade-off under both settings, and a memory regression test that writes 2M timestamps to an HDF5 file and asserts with `tracemalloc` that the check allocates under 2 MB (it allocated 36 MB on `dev`). The expected-report tests in `test_inspector.py` still pass, so the message text and the results on the fixture files are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added `check_subject_age_reference` to validate that `Subject.age__reference`, when present, is one of the supported values (`"birth"` or `"gestational"`). This catches invalid references in files written by tools that do not enforce the schema constraint. [#250](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/250)
 
 ### Improvements
+* `check_regular_timestamps` no longer reads the entire timestamps dataset (twice) into memory. It now reads the first and last `nelems` timestamps and compares the total span of the series against the step, so a gap or rate change anywhere in the series is still detected while the memory used no longer grows with the length of the series. Pass `nelems=None` to restore the full read. [#742](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/742)
 * Raised the minimum required PyNWB to `>=4.0` to track the latest PyNWB release. [#708](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/708)
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
--- a/src/nwbinspector/checks/_time_series.py
+++ b/src/nwbinspector/checks/_time_series.py
@@ -8,35 +8,83 @@ from pynwb.ecephys import SpikeEventSeries
 from pynwb.image import ImageSeries, IndexSeries
 
 from .._registration import Importance, InspectorMessage, Severity, register_check
-from ..utils import get_data_shape, is_ascending_series, is_regular_series
+from ..utils import (
+    cache_data_selection,
+    get_data_shape,
+    is_ascending_series,
+    is_regular_series,
+)
+
+NELEMS = 200
 
 
 @register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=TimeSeries)
 def check_regular_timestamps(
-    time_series: TimeSeries, time_tolerance_decimals: int = 9, gb_severity_threshold: float = 1.0
+    time_series: TimeSeries,
+    time_tolerance_decimals: int = 9,
+    gb_severity_threshold: float = 1.0,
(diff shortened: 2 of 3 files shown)


h-mayorquin (changes requested):
I think it would be better to keep the default exact and make the sampling the opt-in for streaming. Most people run the inspector on a local file they are about to upload and not on a stream, and this PR pays for the streaming case by weakening the answer on the common path.

The cost is bigger than the trade-off in the test. The span check depends only on `tail[-1] - head[0]` and nothing in the middle can move it, so it is not one shifted sample that goes unseen but the whole middle. I ran a few series that are only broken there and all of them are reported as regular:

```python
n = 100_000
base = np.arange(n) / 30000.0
nan = base.copy(); nan[n // 2] = np.nan
jitter = base.copy(); jitter[1000:-1000] += rng.normal(0, 1e-3, n - 2000)
zeros = base.copy(); zeros[1000:-1000] = 0.0
shifted = base.copy(); shifted[1000:-1000] += 0.5
```

All four pass with the default `nelems` and all four fail with `nelems=None`. The body says a jitter section changes the span but jitter is zero mean and it doesn't. The message then tells the user to drop the timestamps for a `rate`, and for the NaN case that is the advice we don't want to give (#689 scans the full array for exactly this reason, padding that was not cleaned before writing).

The memory can be bounded without giving up the full read by changing the criterion instead of the amount of data. Take the period as `(last - first) / (n - 1)` and call the series regular if `first + index * period` displaces no sample by more than a microsecond. I did this in neuroconv last month:
https://github.com/catalystneuro/neuroconv/pull/1925

It chunks with no state between chunks, as each chunk only needs `first`, the period and its own offset. I tried it on your 2M timestamps file under `tracemalloc` (thanks for that test btw, it made this easy to check!): 36 MB on `dev`, 1.6 MB chunked at 100k. That is, the same memory as this PR with a full read of the dataset. What is left is the I/O of reading the dataset once, and that is the part I would keep as the default and let `nelems` turn off.

One more thing that is older than this PR but the criterion above fixes it too. `is_regular_series` rounds the differences to a nanosecond and float64 at epoch magnitude has a spacing of 238 ns. So it never fires on a series whose `t_start` is a Unix time:

```python
t0 = 1688252801.3273287  # t_start of the blackrock PTP file in ephy_testing_data
for rate in (30000.0, 1000.0, 30.0):
    ts = t0 + np.arange(100_000) / rate
    print(rate, is_regular_series(series=ts), is_regular_series(series=ts - t0))
```
```
30000.0 False True
1000.0 False True
30.0 False True
```

Same series measured from zero is accepted. Sampling the ends inherits this as the head and the tail go through the same function.

PR #745, opened 2026-09-04
Bound the amount of spike-time data read by the Units checks

### LINKED ISSUE #744: Units checks read every spike time in the file
Three Units checks read far more spike-time data than they need, which matters for large sorted datasets and for streaming from DANDI.

`check_negative_spike_times` and `check_spike_times_without_nans` both call `units_table["spike_times"].target.data[:]`, which loads every spike time in the file into memory. A dataset with hundreds of millions of spikes is read in full, twice, to answer two yes/no questions. Spike trains are sorted within each unit, so a negative spike time is the first spike of some unit, and NaN padding sits at the end of a unit's train, so both questions can be answered from the first and last spike of each unit. `check_units_table_duration` already builds exactly that set of boundary indices from the small `spike_times_index` array and reads them in one fancy-index call.

`check_ascending_spike_times` loops over every unit and evaluates `units_table["spike_times"][unit_id]`, which for an h5py-backed table reads the unit's full train before the `[:nelems]` slice is applied. The `nelems` limit therefore does not limit I/O. The check can instead read `target.data[start:start + nelems]` for each unit using the index array, so that at most `nelems` values per unit are ever transferred.

I would factor the boundary-index construction out of `check_units_table_duration` into a helper shared by the three checks, and make the ascending check slice through the index.

bendichter:
Fixes #744.

`check_negative_spike_times` and `check_spike_times_without_nans` loaded every spike time in the file with `target.data[:]`, and `check_ascending_spike_times` evaluated `units_table["spike_times"][unit_id]` for every unit, which reads the unit's whole train before the `nelems` slice is applied.

The boundary-index construction that `check_units_table_duration` already had (first and last spike of every non-empty unit, built from the small `spike_times_index` array and read in one selection) is now a helper shared with `check_negative_spike_times`. Spike trains are sorted within a unit, so a negative time is the first spike of some unit; unsorted trains are the job of the ascending check. `check_ascending_spike_times` now walks the index array and slices the flat dataset directly, so at most `nelems` values per unit are transferred.

I had planned to use the boundary spikes for the NaN check as well, on the grounds that padding sits at the end of a train, but the existing test has a NaN in the middle of a train and I think that expectation is right for a CRITICAL check. That check therefore still scans every spike time, but in 500k-element chunks, so its memory no longer grows with the dataset. Its I/O is unchanged.

The new test class writes a 40 MB Units table (10 units, 500k spikes each) and measures each check with `tracemalloc`. On `dev` the peaks were 45 MB, 45 MB, and 8 MB for the negative, NaN, and ascending checks; here the negative, duration, and ascending checks stay under 1 MB and the NaN check under 15 MB. Three smaller tests place the negative time, the NaN padding, and the descending train in a later unit so that they are past the first 200 flat values, and the on-disk ascending test confirms the slice-through-index path finds a problem in the second unit.

This overlaps with #741 on the lines that build the boundary indices in `check_units_table_duration` (the helper here carries the same dtype fix). Whichever merges second will need a trivial rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added `check_subject_age_reference` to validate that `Subject.age__reference`, when present, is one of the supported values (`"birth"` or `"gestational"`). This catches invalid references in files written by tools that do not enforce the schema constraint. [#250](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/250)
 
 ### Improvements
+* `check_negative_spike_times` and `check_units_table_duration` now read only the first and last spike of each unit through a shared helper, `check_ascending_spike_times` reads at most `nelems` spike times per unit directly through the index array instead of materializing each unit's full train, and `check_spike_times_without_nans` scans the spike times in fixed-size chunks. None of these checks loads every spike time in the file into memory any more. [#744](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/744)
 * Raised the minimum required PyNWB to `>=4.0` to track the latest PyNWB release. [#708](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/708)
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
--- a/src/nwbinspector/checks/_ecephys.py
+++ b/src/nwbinspector/checks/_ecephys.py
@@ -1,7 +1,8 @@
 """Check functions specific to extracellular electrophysiology neurodata types."""
 
-from typing import Iterable, Optional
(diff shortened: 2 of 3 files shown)


h-mayorquin (changes requested):
The ascending and the NaN changes look good to me, they give the same answer as before with bounded memory.

I have only one requirement, that the negative check keeps the full read. Reading the first spike of each unit is exact only if the trains are sorted, and `check_ascending_spike_times` only samples the first `nelems` spikes, so a negative value past that is missed by both checks. The chunked scan you wrote for the NaN check gives it the same flat memory with a full read, so I would use that here and leave the boundary read out of this PR.

Also, the methods that we did for this on pynwb are already released: `Units.get_duration()` and `Units.get_starting_time()` are in pynwb 4.0, which we require, so we don't need the helpers.

Note that the ascending check was already sampled before this PR, it stops at `nelems` spikes per unit, so the sortedness that the boundary read relies on was never checked in full. I opened #763 with a way out of that, one chunked pass over the spike times that gives all four checks their answer, but it is bigger than this PR.

I opened #762 to discuss how the streaming case should opt into these heuristics in general, as I don't know that workflow well enough to ask for a mechanism here. Once that settles the boundary read can come back behind it.

PR #747, opened 2026-09-04
Copy the ignore list in configure_checks before adding SKIP entries

### LINKED ISSUE #746: configure_checks appends SKIP entries to the caller's ignore list
When a config contains a `SKIP` section, `configure_checks` appends the skipped check names to the `ignore` list it was given:

```python
if config is not None:
    validate_config(config=config)
    ignore = ignore or []
    for check in checks:
        ...
        if importance_name == "SKIP":
            ignore.append(check.__name__)
```

If the caller passed a list, that list is modified in place. `inspect_nwbfile` forwards `ignore` to `inspect_nwbfile_object`, which calls `configure_checks` on every file, so a user's list grows by the SKIP entries once per inspected file, and any other use the caller had for the list sees the extra names.

```python
from nwbinspector import configure_checks
ignore = ["check_regular_timestamps"]
configure_checks(config=dict(SKIP=["check_data_orientation"]), ignore=ignore)
print(ignore)  # ['check_regular_timestamps', 'check_data_orientation']
```

The function should work on its own copy of the list.

bendichter:
Fixes #746.

`configure_checks` appended the `SKIP` names from a config to the `ignore` list passed in by the caller. Since `inspect_nwbfile_object` calls `configure_checks` for every file, a user's list grew by the SKIP entries once per file. The function now copies the list before adding to it.

The test passes a one-element `ignore` list together with a SKIP config, asserts that the list is unchanged afterwards, and asserts that both the ignored and the skipped check are absent from the result. It fails on `dev`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
Great.

bendichter:

PR #749, opened 2026-09-04
Name the --overwrite flag in file-exists errors

### LINKED ISSUE #748: Overwrite error messages refer to a -o flag that does not exist
Two "file already exists" errors tell the user to pass a `-o` flag:

```
FileExistsError: The file report.json already exists! Specify the '-o' flag to overwrite.
FileExistsError: The file report.txt already exists! Set 'overwrite=True' or pass '-o' flag.
```

The CLI has no `-o` short option. The flag is `--overwrite` (`src/nwbinspector/_nwbinspector_cli.py`), and following the message's advice gives a click usage error. The messages in `_nwbinspector_cli.py` and `_formatting.py` should name `--overwrite`.

bendichter:
Fixes #748.

The "already exists" errors raised by `save_report` and by the CLI for `--json-file-path` told the user to pass `-o`, which is not a flag the CLI defines. Both now say `--overwrite`.

One test calls `save_report` against an existing file and matches on `--overwrite`; the other drives the CLI with click's `CliRunner` against an existing JSON path and checks the exception text. Both fail on `dev`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
+* Fixed the "file already exists" errors from `save_report` and the CLI `--json-file-path` option telling the user to pass a `-o` flag, which does not exist; they now name `--overwrite`. [#748](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/748)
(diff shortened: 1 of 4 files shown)


h-mayorquin (changes requested):
The fix is good, thanks for catching this (the message was sending people to a flag that errors out!).

One thing we could improve here. Both existence checks run after the whole inspection is done. The `json_file_path` one is after `messages = list(messages_iterator)` and the `save_report` one is later still, and neither of them needs the messages, they are only looking at a path. If we move them up to the top with the rest of the argument handling then a user who points `--json-file-path` at a file that already exists finds out right away instead of after scanning a whole dandiset.

That would also make `test_cli_existing_json_file_message_names_overwrite_flag` a lot lighter. Right now it writes an NWB file with `NWBHDF5IO` and runs the whole CLI to check a substring, and with the check at the top it does not need a file at all.

bendichter:
Good call. Both checks now run with the rest of the argument handling, before anything is inspected, and they cover `--report-file-path` as well as `--json-file-path`. The late check on the JSON path is gone since it could no longer fire; the one inside `save_report` stays for callers of the library. The CLI tests now point the command at an empty folder, so no NWB file is written, and there is a third test that runs with `--overwrite` to confirm both files are then rewritten.

h-mayorquin (approved):
LGTM, thanks for addressing the feedback.

PR #751, opened 2026-09-04
Remove arguments deprecated for removal after 3/1/2025

### LINKED ISSUE #750: Remove the stream, version_id, driver, and max_retries arguments deprecated for removal after 3/1/2025
`inspect_all` still accepts `stream` and `version_id`, and `inspect_nwbfile` still accepts `driver` and `max_retries`. All four carry a notice that they "will be removed after 3/1/2025", which is now eighteen months ago. `inspect_all` warns and redirects to `inspect_dandiset`; `inspect_nwbfile` raises a `ValueError` pointing at `inspect_dandi_file_path`. The parallel branch of `inspect_all` also carries a dead `if stream:` remap of `message.file_path` that can never run, because the `stream` branch returns before reaching it.

Since the replacements (`inspect_dandiset`, `inspect_dandi_file_path`, `inspect_url`) have been available for over a year and the CLI's `--stream` option already routes to them, the arguments can be removed. The only in-repo caller is `tests/streaming_tests.py`, which should call `inspect_dandiset` directly.

The `backend` argument of `read_nwbfile` and the `read_nwbfile_and_io` function have later removal dates (12/1/2026 and 2026-11-13) and should stay for now.

bendichter:
Fixes #750.

This removes the four arguments whose deprecation notices said they would be removed after 3/1/2025: `stream` and `version_id` on `inspect_all`, and `driver` and `max_retries` on `inspect_nwbfile`. The `stream` redirect into `inspect_dandiset`, the `ValueError` for `driver` and `max_retries`, and the unreachable `if stream:` remap in the parallel branch go with them. The docstrings now point at `inspect_dandiset`, `inspect_dandi_file_path`, and `inspect_url` for DANDI content. The CLI's `--stream` and `--version-id` options are unaffected, since they already route to those functions.

The only in-repo caller was `test_inspect_all_streaming`, which inspected Dandiset 000126 with the same single check as `test_inspect_dandiset` and asserted the same message, so it is removed rather than rewritten. The `backend` argument of `read_nwbfile` and `read_nwbfile_and_io` have later removal dates and are left alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added `check_subject_age_reference` to validate that `Subject.age__reference`, when present, is one of the supported values (`"birth"` or `"gestational"`). This catches invalid references in files written by tools that do not enforce the schema constraint. [#250](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/250)
 
 ### Improvements
+* Removed the `stream` and `version_id` arguments of `inspect_all` and the `driver` and `max_retries` arguments of `inspect_nwbfile`, all of which had been deprecated for removal after 3/1/2025. Use `inspect_dandiset`, `inspect_dandi_file_path`, or `inspect_url` to inspect files on the DANDI archive. [#750](https://github.com/NeurodataWithoutBorders/nwbinspector/issues/750)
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM, time has come, this will break code for users who were relying on positional arguments but I don't see how to avoid that given that have not kwargs only on the library at the moment.

PR #753, opened 2026-09-04
Drop the unreachable pkg_resources fallback in get_package_version

### LINKED ISSUE #752: get_package_version has an unreachable pkg_resources fallback
`get_package_version` in `nwbinspector/utils/_utils.py` wraps `importlib.metadata.version` in a `try` that catches `ModuleNotFoundError` and falls back to `pkg_resources.get_distribution`:

```python
try:
    from importlib.metadata import version as importlib_version
    package_version = importlib_version(name)
except ModuleNotFoundError:  # Remove the except clause when minimal supported version becomes 3.8
    from pkg_resources import get_distribution
    package_version = get_distribution(name).version
```

The comment refers to Python 3.8, and the project requires 3.10, so `importlib.metadata` is always present. Beyond that, `importlib.metadata.version` raises `PackageNotFoundError` for an unknown package, not `ModuleNotFoundError`, so the fallback never ran even on older Pythons. `pkg_resources` is deprecated in setuptools and is not a declared dependency, so the fallback would fail to import if it were ever reached.

The function should call `importlib.metadata.version` directly and let `PackageNotFoundError` propagate.

bendichter:
Fixes #752.

`get_package_version` caught `ModuleNotFoundError` around `importlib.metadata.version` and fell back to `pkg_resources`. That branch could never run: `importlib.metadata` is always importable on Python 3.10 and up, and an unknown package raises `PackageNotFoundError`, which was not caught. The function now calls `importlib.metadata.version` directly and documents that `PackageNotFoundError` propagates.

Tests cover the installed case (returns a `Version`, and PyNWB reports at least 4.0) and the missing-package case. The remaining mypy errors reported for this module are pre-existing and in `is_ascending_series`; this change removes one of the nine reported on `dev`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added `check_subject_age_reference` to validate that `Subject.age__reference`, when present, is one of the supported values (`"birth"` or `"gestational"`). This catches invalid references in files written by tools that do not enforce the schema constraint. [#250](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/250)
 
 ### Improvements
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
This was unreachable.

bendichter:

PR #757, opened 2026-09-04
Raise ValueError from calculate_number_of_cpu instead of asserting

### LINKED ISSUE #755: calculate_number_of_cpu validates its argument with assert
`calculate_number_of_cpu` in `utils/_utils.py` validates its argument with `assert` statements:

```python
assert requested_cpu <= total_cpu, f"Requested more CPUs ({requested_cpu}) than are available ({total_cpu})!"
assert requested_cpu >= -(total_cpu - 1), f"Requested fewer CPUs ({requested_cpu}) than are available ({total_cpu})!"
```

`assert` is stripped when Python runs with `-O`, so under optimized bytecode an out-of-range `--n-jobs` value is passed through to `ProcessPoolExecutor` unchecked. Argument validation in library code should raise `ValueError`. The second message is also misleading, since it fires when the requested value is too negative rather than when fewer CPUs were requested than exist.

Since callers may currently catch `AssertionError`, this is a small behavior change and the tests that check for it need to be updated.

bendichter:
Fixes #755.

`calculate_number_of_cpu` validated `requested_cpu` with two `assert` statements, which Python removes under `-O`, so in that mode an out-of-range `n_jobs` reached `ProcessPoolExecutor` unchecked. Both checks now raise `ValueError`. The message for a too-negative request used to say "Requested fewer CPUs than are available", which described the wrong condition; it now states the minimum and why it exists.

Callers that caught `AssertionError` for this case will need to catch `ValueError` instead; there are none in the repo. The two existing tests are updated for the exception type and the new message, and a positive-value test is added. The behavior for `0` (use all CPUs) is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Bumped GitHub Actions workflow dependencies (`actions/checkout` v4 -> v6, `actions/setup-python` v5 -> v6, `codecov/codecov-action` v4 -> v5) to migrate off Node.js 20, which GitHub is removing from runners on 2026-09-16. [#702](https://github.com/NeurodataWithoutBorders/nwbinspector/pull/702)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
While python -O really has not other meaning than removing assertions the error specificity here looks like an improvement to me.

bendichter:

PR #758, opened 2026-09-08
Add `check_photon_series_undeclared_depth`

h-mayorquin:
I was doing some measurements for ophys support of volumetric series for the TAB meeting and I found that there are some datasets on dandi that declare depth without content. 
The trailing 1 is a channel axis that was never squeezed out after splitting the data into one series per channel. But NWB defines the fourth axis of a photon series as depth, so a reader is obliged to treat that file as volumetric.

I think this is ambiguous and we should invite the users to either remove the meaningless depth or to document that it is indeed volumetric on any of those entries:

* `grid_spacing` on the `ImagingPlane`
* `origin_coords` on the `ImagingPlane`
* `dimension` on the series

The check fires whenever none of the three is set, whatever the length of the depth axis, as a depth with no spacing and no position cannot be interpreted either way.

This is added in this PR as a `BEST_PRACTICE_VIOLATION`.


bendichter (commented):
Thanks, the check is right and the tests cover the cases that matter. My comments are on the best practice, which I think should lead with what goes wrong rather than with the schema rule, and on two design points: whether ``dimension`` should count as a declaration, and whether a depth greater than one belongs at the same importance level.

docs/best_practices/ophys.rst
```diff
@@ -29,3 +29,24 @@ subjects, we recommend using terms from the :allen-brain-map:`Allen Brain Atlas
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+NWB defines the fourth axis of a ``TwoPhotonSeries`` or ``OnePhotonSeries`` as depth, so any reader is obliged to treat
+a four-dimensional series as volumetric. Nothing else in the file says how those planes are spaced or where they sit
+unless you declare it, so a four-dimensional series with no depth geometry cannot be interpreted as a volume.
+
+If the data is volumetric, declare it: give the ``ImagingPlane`` a three-component ``grid_spacing`` (or
+``origin_coords``), or set a three-component ``dimension`` on the series. If the depth is one, the axis is usually a
+channel or plane axis that was never squeezed out after splitting the data into one series per channel, and the data
+should be stored as ``(time, rows, columns)`` instead.
```
bendichter:
```suggestion
.. _best_practice_photon_series_depth_axis:

Depth Axis
~~~~~~~~~~

The ``data`` of a ``TwoPhotonSeries`` or ``OnePhotonSeries`` is either ``(time, width, height)`` for a single plane or
``(time, width, height, depth)`` for a volume. Readers use the number of axes to tell the two apart, so a trailing axis
of length one turns a planar recording into a one-plane volume: viewers render it as a volume, archive metadata counts
it as volumetric, and readers that expect planar data reject it. This usually happens when a channel or plane axis is
left in place after splitting a multi-channel recording into one series per channel, which leaves each series with a
shape like ``(time, width, height, 1)``. If the recording is a single plane, store it with three axes.

If the recording is volumetric, describe the depth geometry so the volume can be interpreted. Set a three-component
``grid_spacing`` on the ``ImagingPlane``, which gives the spacing between planes, and where known a three-component
``origin_coords``. Without the spacing, the planes have no known distance from each other, and the volume cannot be
measured, rendered to scale, or registered to anything.
```
The current text starts from the schema rule rather than from what goes wrong, so a reader who has not hit this does not see why it matters. This version leads with the consequence of the spurious axis, then covers the real-volume case. Two smaller things folded in: schema 2.10 (shipped with pynwb 4.1) defines the data as ``data[time, width, height]`` with width being columns, so "(time, rows, columns)" is the old convention; and I dropped ``dimension`` from the recommendation, see the comment on the check. The label changes with the heading, so the docstring reference needs the same edit.

src/nwbinspector/checks/_ophys.py
```diff
@@ -134,3 +135,46 @@ def check_imaging_plane_location_allen_ccf(imaging_plane: ImagingPlane) -> Optio
         )
 
     return None
+
+
+def _declares_depth(image_series: ImageSeries, imaging_plane: ImagingPlane) -> bool:
+    """Return True when the file states somewhere that the series has a depth axis."""
+    for declaration in (image_series.dimension, imaging_plane.grid_spacing, imaging_plane.origin_coords):
```
bendichter: I would drop ``dimension`` from the accepted declarations. It is the pixel count per axis, so ``dimension=[512, 512, 1]`` repeats the shape and adds no geometry. As written it silences the check without resolving the ambiguity it exists to catch.

src/nwbinspector/checks/_ophys.py
```diff
@@ -134,3 +135,46 @@ def check_imaging_plane_location_allen_ccf(imaging_plane: ImagingPlane) -> Optio
...
+        if declaration is not None and len(declaration) == 3:
+            return True
+    return False
+
+
+@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
```
bendichter: A depth of one with nothing declared is almost always a mistake, and a violation fits. A depth greater than one with no ``grid_spacing`` is missing metadata, which is the same state as a planar series with no ``grid_spacing``, and nothing flags that today. Treating it as a violation is inconsistent. I would fire the violation only when the depth is one, and either leave the larger case alone for now or make it a separate suggestion about missing plane geometry.

src/nwbinspector/checks/_ophys.py
```diff
@@ -134,3 +135,46 @@ def check_imaging_plane_location_allen_ccf(imaging_plane: ImagingPlane) -> Optio
...
+    if _declares_depth(image_series=image_series, imaging_plane=imaging_plane):
+        return None
+
+    return InspectorMessage(
+        message=(
+            f"The data is four-dimensional with a depth axis of length {data_shape[3]}, but neither "
+            f"the series nor its imaging plane ('{imaging_plane.name}') declares a depth. Set "
+            "'grid_spacing' (or 'origin_coords') on the imaging plane to three components, or "
+            "'dimension' on the series, so the data can be interpreted as a volume. If the axis is a "
+            "leftover from splitting channels or planes, store the data as (time, rows, columns) instead."
```
bendichter: Two wording points on the message. "declares a depth" reads oddly; "describes the depth axis" or "gives the spacing between planes" says what is missing. And "(time, rows, columns)" should be "(time, width, height)" to match the schema's own axis names.

tests/unit_tests/test_ophys.py
```diff
@@ -424,3 +426,109 @@ def test_skip_check_imaging_plane_location_allen_ccf_non_mouse():
...
+        data=np.ones(shape),
+        dimension=dimension,
+        unit="n.a.",
+        rate=30.0,
+    )
+
+
+class TestCheckPhotonSeriesUndeclaredDepth(TestCase):
+    def test_undeclared_singleton_depth_triggers(self):
+        """The dandiset 000491 shape: 4D with depth 1 and no geometry declared anywhere."""
```
bendichter: Please drop the dandiset from the docstring, here and in the PR description. The shape is enough to say what the test covers.

src/nwbinspector/checks/_ophys.py
```diff
@@ -134,3 +135,46 @@ def check_imaging_plane_location_allen_ccf(imaging_plane: ImagingPlane) -> Optio
...
 
     return None
+
+
+def _declares_depth(image_series: ImageSeries, imaging_plane: ImagingPlane) -> bool:
+    """Return True when the file states somewhere that the series has a depth axis."""
+    for declaration in (image_series.dimension, imaging_plane.grid_spacing, imaging_plane.origin_coords):
+        if declaration is not None and len(declaration) == 3:
+            return True
+    return False
```
bendichter:
```suggestion
def _describes_depth(imaging_plane: ImagingPlane) -> bool:
    """Return True when the imaging plane gives the spacing or position of the planes along the depth axis."""
    for geometry in (imaging_plane.grid_spacing, imaging_plane.origin_coords):
        if geometry is not None and len(geometry) == 3:
            return True
    return False
```
`dimension` is the pixel count per axis, so `dimension=[512, 512, 1]` repeats the shape and adds no geometry. Accepting it lets a file silence the check without resolving the ambiguity it exists to catch, so this drops it and keeps the two fields that carry the spacing or position of the planes.

src/nwbinspector/checks/_ophys.py
```diff
@@ -134,3 +135,46 @@ def check_imaging_plane_location_allen_ccf(imaging_plane: ImagingPlane) -> Optio
...
+
+    return InspectorMessage(
+        message=(
+            f"The data is four-dimensional with a depth axis of length {data_shape[3]}, but neither "
+            f"the series nor its imaging plane ('{imaging_plane.name}') declares a depth. Set "
+            "'grid_spacing' (or 'origin_coords') on the imaging plane to three components, or "
+            "'dimension' on the series, so the data can be interpreted as a volume. If the axis is a "
+            "leftover from splitting channels or planes, store the data as (time, rows, columns) instead."
+        )
+    )
```
bendichter, line 179:
```suggestion
@register_check(importance=Importance.BEST_PRACTICE_VIOLATION, neurodata_type=ImageSeries)
def check_photon_series_undeclared_depth(image_series: ImageSeries) -> Optional[InspectorMessage]:
    """
    Check that a photon series with a depth axis of length one is either planar or a described volume.

    Readers tell a plane from a volume by the number of axes, so a trailing axis of length one turns a
    planar recording into a one-plane volume. That axis is usually a channel or plane axis left in place
    after splitting a recording into one series per channel. A single plane should be stored with three
    axes, and a real one-plane volume should give the spacing or position of its planes through a
    three-component ``grid_spacing`` or ``origin_coords`` on the imaging plane.

    Best Practice: :ref:`best_practice_photon_series_depth_axis`
    """
    imaging_plane = getattr(image_series, "imaging_plane", None)
    if imaging_plane is None:  # a plain ImageSeries has no imaging plane and no depth semantics
        return None

    data_shape = get_data_shape(image_series.data, strict_no_data_load=True)
    if data_shape is None or len(data_shape) != 4 or data_shape[3] != 1:
        return None

    if _describes_depth(imaging_plane=imaging_plane):
        return None

    return InspectorMessage(
        message=(
            f"The data is four-dimensional with a depth axis of length 1, but the imaging plane "
            f"('{imaging_plane.name}') does not describe the depth axis. If the recording is a single plane, "
            "store the data as (time, width, height). If it is a one-plane volume, set a three-component "
            "'grid_spacing' or 'origin_coords' on the imaging plane."
        )
    )
```
This narrows the check to a depth of one, which is almost always a mistake and fits a violation. A depth greater than one with no `grid_spacing` is missing metadata, the same state as a planar series with no `grid_spacing`, which nothing flags today, so treating it as a violation here would be inconsistent. That case could become a separate suggestion about missing plane geometry later. The message uses the schema's own axis names, since schema 2.10 defines the data as `data[time, width, height]`, and the label follows the new heading in the docs.

bendichter (commented):
Here are the earlier comments as concrete edits. They are written to be applied together with the docs suggestion above: the helper drops `dimension`, the check narrows to a depth of one and uses the schema's axis names and the new docs label, the tests follow, and the changelog entry describes the narrowed check.

tests/unit_tests/test_ophys.py
```diff
@@ -424,3 +426,109 @@ def test_skip_check_imaging_plane_location_allen_ccf_non_mouse():
...
+            object_name="TwoPhotonSeries",
+            location="/",
+        )
+
+    def test_undeclared_real_depth_triggers(self):
+        """A depth greater than one is still uninterpretable as a volume without a declared geometry."""
+        imaging_plane = _make_imaging_plane()
+        photon_series = _make_photon_series(shape=(20, 10, 10, 4), imaging_plane=imaging_plane)
+
+        assert check_photon_series_undeclared_depth(photon_series) is not None
```
bendichter:
```suggestion
class TestCheckPhotonSeriesUndeclaredDepth(TestCase):
    def test_undeclared_singleton_depth_triggers(self):
        """A 4D series with a depth of one and no geometry on the imaging plane."""
        imaging_plane = _make_imaging_plane()
        photon_series = _make_photon_series(shape=(20, 10, 10, 1), imaging_plane=imaging_plane)

        assert check_photon_series_undeclared_depth(photon_series) == InspectorMessage(
            message=(
                "The data is four-dimensional with a depth axis of length 1, but the imaging plane "
                "('ImagingPlane') does not describe the depth axis. If the recording is a single plane, "
                "store the data as (time, width, height). If it is a one-plane volume, set a three-component "
                "'grid_spacing' or 'origin_coords' on the imaging plane."
            ),
            importance=Importance.BEST_PRACTICE_VIOLATION,
            check_function_name="check_photon_series_undeclared_depth",
            object_type="TwoPhotonSeries",
            object_name="TwoPhotonSeries",
            location="/",
        )

    def test_undescribed_real_depth_passes(self):
        """A depth greater than one is a volume whatever the imaging plane says, so it is out of scope here."""
        imaging_plane = _make_imaging_plane()
        photon_series = _make_photon_series(shape=(20, 10, 10, 4), imaging_plane=imaging_plane)

        assert check_photon_series_undeclared_depth(photon_series) is None
```
Matches the narrowed check and the new message, and drops the dandiset from the docstring. Please drop it from the PR description as well.

tests/unit_tests/test_ophys.py
```diff
@@ -424,3 +426,109 @@ def test_skip_check_imaging_plane_location_allen_ccf_non_mouse():
...
+        imaging_plane = _make_imaging_plane(origin_coords=[1.0, 2.0, 3.0])
+        photon_series = _make_photon_series(shape=(20, 10, 10, 1), imaging_plane=imaging_plane)
+
+        assert check_photon_series_undeclared_depth(photon_series) is None
+
+    def test_singleton_depth_declared_by_dimension_passes(self):
+        imaging_plane = _make_imaging_plane()
+        photon_series = _make_photon_series(shape=(20, 10, 10, 1), imaging_plane=imaging_plane, dimension=[10, 10, 1])
+
+        assert check_photon_series_undeclared_depth(photon_series) is None
```
bendichter:
```suggestion
    def test_singleton_depth_with_dimension_still_triggers(self):
        """A three-component dimension repeats the shape and adds no geometry, so the axis is still undescribed."""
        imaging_plane = _make_imaging_plane()
        photon_series = _make_photon_series(shape=(20, 10, 10, 1), imaging_plane=imaging_plane, dimension=[10, 10, 1])

        assert check_photon_series_undeclared_depth(photon_series) is not None
```

changelog_entries/758.check.md
```diff
@@ -0,0 +1 @@
+Added `check_photon_series_undeclared_depth`, which flags a four-dimensional `TwoPhotonSeries` or `OnePhotonSeries` when neither the series' `dimension` nor the imaging plane's `grid_spacing` or `origin_coords` declares a third dimension. Without one of those the depth axis has no spacing or position, so the data cannot be interpreted as a volume.
```
bendichter:
```suggestion
Added `check_photon_series_undeclared_depth`, which flags a `TwoPhotonSeries` or `OnePhotonSeries` whose data has a depth axis of length one when the imaging plane's `grid_spacing` and `origin_coords` say nothing about a third dimension. Such an axis is usually a channel or plane axis left in place after splitting a recording into one series per channel, and it turns a planar recording into a one-plane volume for every reader.
```

docs/best_practices/ophys.rst
h-mayorquin: Yes, I like this, education and then formalities rather than a dry rule upfront.

src/nwbinspector/checks/_ophys.py
h-mayorquin: Makes sense.

src/nwbinspector/checks/_ophys.py
h-mayorquin: Yes, this is more specific which is better. Thanks.

h-mayorquin:
Thanks for the review, I removed the reference from the PR message.

bendichter (approved):

PR #759, opened 2026-09-08
Add `check_image_series_external_file_format`

### LINKED ISSUE #669: [Discussion] Add best practice check for browser-compatible video codecs in ImageSeries external files
A core goal of DANDI and the NWB ecosystem is making neuroscience data not just archived but genuinely accessible and reusable. For behavioral video stored as `ImageSeries` external files, the choice of video codec has a direct impact on storage efficiency and interoperability. Older or camera-native codecs like MJPEG, MPEG-4 Part 2 (mp4v), and DIVX produce significantly larger files than modern alternatives like H.264 for the same visual content, and are poorly supported across platforms and tooling. Modern codecs like H.264, VP9, and AV1 have near-universal support. In practice, most labs don't intentionally choose these older codecs; they inherit them from camera SDK defaults (FLIR SpinView, Basler pylon) or OpenCV scripts. Cameras favor MJPEG because each frame encodes independently, which is computationally cheap enough for onboard hardware and avoids latency or data loss propagation during capture - but these advantages are irrelevant once the data is archived.

The NWB Inspector already checks that `ImageSeries` external files exist and use relative paths. A natural addition would be a best practice check that probes the codec of each referenced video file and suggests re-encoding to a widely supported modern codec when an inefficient or poorly supported one is detected. Since most of these files are already lossy-compressed, re-encoding to H.264 actually improves compression (smaller files) with negligible additional quality loss at high quality settings (CRF 17-18). The check message could include a ready-to-use ffmpeg command, e.g. `ffmpeg -i input.avi -c:v libx264 -crf 18 -pix_fmt yuv420p output.mp4`. For the rare case of intentionally lossless video (FFV1, raw), users can ignore the suggestion as with any best practice check.

| Detected codec | Suggest | Reason |
|---|---|---|
| MJPEG | H.264 | No inter-frame compression, very large files, poor tooling support |
| MPEG-4 Part 2 (mp4v, DIVX, XVID) | H.264 | Outdated lossy codec, larger files and worse support than H.264 |
| Raw / uncompressed | H.264 | No compression at all, enormous files |
| FFV1 | H.264 | Lossless and not widely supported; note that re-encoding is lossy |
| H.264, VP8, VP9, AV1 | No suggestion | Already widely supported and efficient |

h-mayorquin:
In #669 we discussed which video formats belong in an NWB file and @bendichter asked to move forward with a best practice for it. This PR adds that entry to the ImageSeries best practices and a `BEST_PRACTICE_SUGGESTION` check that opens each `external_file` with PyAV and asks for H.264, VP8, VP9 or AV1 in an MP4 or WebM container for lossy video, and FFV1 for lossless video with no opinion on its container. The message names the four lossy codecs without picking one and only notes that H.264 is covered by patents while VP8, VP9 and AV1 are royalty-free.

This should close #669.

### DIFF
--- a/docs/best_practices/image_series.rst
+++ b/docs/best_practices/image_series.rst
@@ -45,3 +45,39 @@ If there is no external file, there should be no starting frame set. This is a l
 older versions of PyNWB (< 2.2.0).
 
 Check function: :py:meth:`~nwbinspector.checks._image_series.check_image_series_starting_frame_without_external_file`
+
+
+Video Files
+-----------
+
+
+.. _best_practice_external_file_format:
+
+Use a standard video container and codec
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The format of a video has two parts, the codec that the frames are encoded with and the container that holds them,
+and the right choice for each depends on whether the video is lossy or lossless.
+
+For lossy video, use H.264, VP8, VP9 or AV1 in an MP4 or WebM container. All four codecs are efficient and play
+everywhere, and which of them to pick depends on the tooling and hardware available. H.264 is covered by patents
+managed through a patent pool, while VP8, VP9 and AV1 are royalty-free. MP4 and WebM are the two containers that every
+platform supports and the only two a browser can play: Safari supports neither Matroska (``.mkv``) nor the other
+legacy containers, and the
+`MDN container guide <https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Containers>`_ lists MP4 and
+WebM as the only two with universal support. Most scientific video is in ``.avi`` or ``.mov`` with a legacy codec such
+as MJPEG, MPEG-4 Part 2 (``mp4v``, DIVX, XVID), WMV or DV, because that is what the camera SDK or the operating system
+produced by default, not because either was chosen.
+
+When only the container is wrong, no re-encoding is needed. The stream is copied byte for byte and only the container
+headers are rewritten, which takes seconds even for a large file: ``ffmpeg -i input.mkv -c copy output.mp4``. When the
+codec is wrong the re-encoding settles both at once:
+``ffmpeg -i input.avi -c:v libx264 -crf 18 -pix_fmt yuv420p output.mp4``.
+
+For lossless video, use FFV1. It compresses two to three times smaller than uncompressed video with identical pixel
+values, adds per-frame checksums, and handles grayscale and high bit depth without the chroma subsampling that
+consumer codecs impose, which is why the preservation community standardised on it. FFV1 is normally paired with MKV,
+but the container makes no practical difference here: every tool that reads FFV1 handles MKV and AVI alike, and no
+browser decodes lossless video whatever it is held in. Lossless data should never be re-encoded to a lossy codec.
+
+Check function: :py:meth:`~nwbinspector.checks._image_series.check_image_series_external_file_format`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dependencies = [
     "click",
     "tqdm",
     "isodate",
+    "av",
 ]
 
 [project.optional-dependencies]
--- a/src/nwbinspector/checks/_image_series.py
+++ b/src/nwbinspector/checks/_image_series.py
@@ -55,6 +55,69 @@ def check_image_series_external_file_relative(image_series: ImageSeries) -> Opti
     return None
 
 
+RECOMMENDED_LOSSY_CONTAINERS = (".mp4", ".webm")
+RECOMMENDED_LOSSY_CODECS = ("h264", "vp8", "vp9", "av1")
+RECOMMENDED_LOSSLESS_CODECS = ("ffv1",)
+
+
+@register_check(importance=Importance.BEST_PRACTICE_SUGGESTION, neurodata_type=ImageSeries)
+def check_image_series_external_file_format(image_series: ImageSeries) -> Optional[Iterable[InspectorMessage]]:
+    """
+    Check if the external_file of an ImageSeries uses a standard video container and codec.
+
+    Best Practice: :ref:`best_practice_external_file_format`
+    """
+    # False positive case; TwoPhotonSeries are a subclass of ImageSeries, but their external files are imaging
+    # data such as TIFF stacks rather than video
+    if isinstance(image_series, TwoPhotonSeries) or image_series.external_file is None:
+        return None
+    try:
+        import av
+    except ImportError:  # the format cannot be read without PyAV, so nothing can be said about it
+        return None
+
+    nwbfile_path = Path(get_nwbfile_path_from_internal_object(neurodata_object=image_series))
+    for file_path in image_series.external_file:
+        file_path = file_path.decode() if isinstance(file_path, bytes) else file_path
+        resolved_path = Path(file_path) if Path(file_path).is_absolute() else nwbfile_path.parent / file_path
+        if not resolved_path.exists():  # reported by check_image_series_external_file_valid
+            continue
+        try:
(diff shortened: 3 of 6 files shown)


docs/best_practices/image_series.rst
```diff
@@ -45,3 +45,39 @@ If there is no external file, there should be no starting frame set. This is a l
...
+.. _best_practice_external_file_format:
+
+Use a standard video container and codec
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The format of a video has two parts, the codec that the frames are encoded with and the container that holds them,
+and the right choice for each depends on whether the video is lossy or lossless.
+
+For lossy video, use H.264, VP8, VP9 or AV1 in an MP4 or WebM container. All four codecs are efficient and play
+everywhere, and which of them to pick depends on the tooling and hardware available. H.264 is covered by patents
```
bendichter:
```suggestion
For lossy video, use H.264, VP8, VP9 or AV1 in an MP4 or WebM container. All four codecs are efficient and
widely supported, and which of them to pick depends on the tooling and hardware available. H.264 is covered by patents
```

bendichter (commented):
Thanks, this reads well and the check itself looks right to me. My comments are on the wording in the best practice, which makes a few claims that are broader than we can defend, and on the new dependency.

docs/best_practices/image_series.rst
```diff
@@ -45,3 +45,39 @@ If there is no external file, there should be no starting frame set. This is a l
...
+
+For lossy video, use H.264, VP8, VP9 or AV1 in an MP4 or WebM container. All four codecs are efficient and
+widely supported, and which of them to pick depends on the tooling and hardware available. H.264 is covered by patents
+managed through a patent pool, while VP8, VP9 and AV1 are royalty-free. MP4 and WebM are the two containers that every
+platform supports and the only two a browser can play: Safari supports neither Matroska (``.mkv``) nor the other
+legacy containers, and the
+`MDN container guide <https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Containers>`_ lists MP4 and
+WebM as the only two with universal support. Most scientific video is in ``.avi`` or ``.mov`` with a legacy codec such
+as MJPEG, MPEG-4 Part 2 (``mp4v``, DIVX, XVID), WMV or DV, because that is what the camera SDK or the operating system
+produced by default, not because either was chosen.
```
bendichter:
```suggestion
For lossy video, use H.264, VP8, VP9 or AV1 in an MP4 or WebM container. All four codecs are efficient and
widely supported, and which of them to pick depends on the tooling and hardware available. H.264 is covered by patents
managed through a patent pool, while VP8, VP9 and AV1 are royalty-free. MP4 and WebM are the containers with the widest
support, including in web browsers, which lets a video be viewed without downloading it. Scientific video is often in
``.avi`` or ``.mov`` with an older codec such as MJPEG, MPEG-4 Part 2 (``mp4v``, DIVX, XVID), WMV or DV, because that
is what the camera software produced by default.
```
"Every platform supports" and "the only two a browser can play" overreach: browsers also play Ogg, and Safari's WebM support is recent and partial. Matroska is not a legacy container either, WebM is a profile of it. I would drop the Safari detail and the MDN citation and say "widest support", which is what we can defend.

docs/best_practices/image_series.rst
```diff
@@ -45,3 +45,39 @@ If there is no external file, there should be no starting frame set. This is a l
...
+When only the container is wrong, no re-encoding is needed. The stream is copied byte for byte and only the container
+headers are rewritten, which takes seconds even for a large file: ``ffmpeg -i input.mkv -c copy output.mp4``. When the
+codec is wrong the re-encoding settles both at once:
+``ffmpeg -i input.avi -c:v libx264 -crf 18 -pix_fmt yuv420p output.mp4``.
+
+For lossless video, use FFV1. It compresses two to three times smaller than uncompressed video with identical pixel
+values, adds per-frame checksums, and handles grayscale and high bit depth without the chroma subsampling that
+consumer codecs impose, which is why the preservation community standardised on it. FFV1 is normally paired with MKV,
+but the container makes no practical difference here: every tool that reads FFV1 handles MKV and AVI alike, and no
+browser decodes lossless video whatever it is held in. Lossless data should never be re-encoded to a lossy codec.
```
bendichter:
```suggestion
For lossless video, use FFV1. It is much smaller than uncompressed video with identical pixel values, adds per-frame
checksums, and handles grayscale and high bit depth, which is why it is widely used for archival video. FFV1 is usually
paired with MKV, but the container matters less here, since browsers do not play lossless video in any container. If
lossless data is important, be careful not to accidentally re-encode it to a lossy codec.
```
The compression ratio depends on the content, H.264 has 4:4:4 profiles so consumer codecs do not all impose subsampling, and "every tool that reads FFV1" is hard to stand behind. The recommendation does not need any of those claims.

src/nwbinspector/checks/_image_series.py
```diff
@@ -55,6 +55,69 @@ def check_image_series_external_file_relative(image_series: ImageSeries) -> Opti
...
+                    f"The external file '{file_path}' uses the '{codec}' codec, which is not one of the standard "
+                    "video codecs. Please use H.264, VP8, VP9 or AV1 in an MP4 or WebM container, or FFV1 if the "
+                    "video has to stay lossless. Note that H.264 is covered by patents while VP8, VP9 and AV1 are "
+                    "royalty-free."
+                )
+            )
+            continue  # the re-encoding settles the container as well
+
+        if codec in RECOMMENDED_LOSSLESS_CODECS:
+            continue  # every tool that reads FFV1 handles its containers alike, so the container gains nothing
```
bendichter: Same "every tool" claim as in the docs. Something like "the container of a lossless video is not reported, since browsers do not play lossless video in any container" says what the branch does without the sweeping part.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
     "click",
     "tqdm",
     "isodate",
+    "av",
```
bendichter, line 47: PyAV bundles FFmpeg and its wheel is tens of megabytes, and the check already returns quietly when the import fails. I would make it an optional extra, `nwbinspector[video]`, the way `[zarr]` and `[dandi]` work, and add it to the dev dependency group so the tests still run.

src/nwbinspector/checks/_image_series.py
h-mayorquin: Yes, fair point, this is one of those comments that the LLM writes as a result as an internal discussion rather than the intended audience. Thanks.

docs/best_practices/image_series.rst
h-mayorquin: I agree wit this and the other comments, the general pattern here is over justifying

pyproject.toml
h-mayorquin: My reasoning was that, yes, it is heavy, but this is an important check and we should run it all the time. An alternative I entertained was to make it optional here and mandatory on DANDI. The downside of that approach is that people might validate locally, then go to DANDI and become annoyed because there is a new thing to do. The trade-off is then package size vs user ergonomics, and I chose the latter. How do you feel about it?

That said, now that you bring it up, I think there is also packaging fragility, as the bundled build will fail for some new Python versions and architectures. So mabye the extra approach makes more sense after all. If we move towards making it an extra, I think we should make it required in dandi-cli, that is, always install the video extra there (or should we call it media?).

And regardless, we should also raise the level from BEST_PRACTICE_VIOLATION to CRITICAL over there. Note that I just updated suggestion to violation here as I think it feelst better.

src/nwbinspector/checks/_image_series.py
h-mayorquin: Changed this.

PR #764, opened 2026-09-16
Use towncrier for the changelog

bendichter:
## Motivation

With many pull requests open at once, nearly every one of them edits the same lines at the top of `CHANGELOG.md`, so they conflict with each other as they merge. This PR introduces [towncrier](https://towncrier.readthedocs.io/) to handle the changelog the way neuroconv now does (catalystneuro/neuroconv#2042).

Changelog entries live one file per entry under `changelog_entries/`, named `<PR number>.<type>.md`, so two pull requests never touch the same lines. The types are `check`, `improvement` and `fix`, which render under the three headings the changelog already uses. The `[#747](...)` link is generated from the file name, so entries no longer carry it in the text.

The 19 entries sitting in the `v0.7.3 (Upcoming)` section moved into files, and `towncrier build --draft --version 0.7.3` reproduces all of them. The few that linked to an issue rather than the pull request are now named after the pull request, since the file name has to match the PR for the CI check. CI now requires a file named after this pull request's own number instead of a change to `CHANGELOG.md`, and the error message says what to add.

At release, `towncrier build --version 0.7.3 --date "September 16, 2026"` writes the section and deletes the entry files. The developer guide gains a section on writing an entry and its release checklist replaces the "update the header" step with the build command.

I will move the entries of the open pull requests into this format in separate commits on their branches.


h-mayorquin (approved):
LGTM

PR #765, opened 2026-09-16
Add check_single_chunk_dataset

### LINKED ISSUE #722: [Add Check]: check if a dataset is just one big uncompressed chunk

bendichter:
## Motivation

Closes #722. Replaces #724.

Some datasets are chunked, but with one chunk that covers the whole array, and no compression. That layout has the overhead of chunked storage without its benefits: a reader still has to load the whole dataset to touch any part of it, and nothing is compressed. The right fix depends on size. A small dataset of this kind would be better stored contiguously. A large one should be split into several chunks and compressed.

#724 came close, but it only fired above 50 MB. Above that size an uncompressed dataset already trips `check_small_dataset_compression` or `check_large_dataset_compression`, so everything it flagged got two messages, while the medium-sized case from the issue never fired at all, and the message hedged across both situations.

## Changes

`check_single_chunk_dataset` fires on any non-empty HDF5 dataset that is chunked, uncompressed, and whose single chunk spans the full shape. Below `mb_lower_bound` (default 50 MB, matching the compression check) the message suggests contiguous storage. At or above it, the message says to split the dataset into several chunks and enable compression, and leaves the missing compression itself to the existing checks. It is a `BEST_PRACTICE_SUGGESTION` on `NWBContainer`. Since the advice applies to any dataset in the file, the chunking and compression best practices move out of the Time Series page into a new Data Storage page, with their labels unchanged, and the new check is documented there.

Tests cover both messages, a resizable dataset whose chunk is larger than its current shape, and the contiguous, compressed, multi-chunk and empty cases that should pass.


h-mayorquin (approved):
Makes sense to me.

A small comment regarding resizable datasets. I have seen that some users (some tooling of the Clandinin lab specifically) write small NWB files with dynamic tables initiated with `maxshape=None` and save a small seeding that is meant to be appended to later in their pipeline. The check will misfire for such use cases. We could avoid this by skipping the suggestion when the dataset is resizable (`field.maxshape != field.shape`) and below `mb_lower_bound`, as HDF5 requires chunked storage for resizable datasets and the contiguous advice can't be followed there.

But...
* This is an edge case
* This is only a best practice suggestion
* I don't think users in the case above will run the inspector in between the seeding and the appending.

So it should be OK to ignore.

bendichter:
oh that's a good point about appendable dynamic tables. I like your idea of skipping them. I'll open a follow-up PR

bendichter:
Opened #766, which skips a small dataset that can still be resized and keeps the split-and-compress advice for a large one.

PR #766, opened 2026-09-16
Skip small resizable datasets in check_single_chunk_dataset

bendichter:
## Motivation

Follow-up to #765, from @h-mayorquin's review there. Some pipelines write a small NWB file whose dynamic tables are created with `maxshape=None` and seeded with a few rows that are appended to later. Those datasets are chunked, uncompressed, and covered by one chunk, so the check told the user to store them contiguously, which HDF5 does not allow for a resizable dataset.

## Changes

Below the size threshold, a dataset that can still be resized is skipped. At or above it the advice is to split the dataset into several chunks and compress it, which a resizable dataset can follow, so that case still fires. The best practice text gets a sentence on the exception, and the test that used a resizable dataset to show a chunk larger than the current shape now checks both branches.


h-mayorquin (commented):
LGTM

bendichter:
@h-mayorquin can you please approve officially?

h-mayorquin (approved):

